Issue #85, opened 2023-01-25
Make `beluga_amcl` a drop-in replacement of `nav2_amcl`


nahueespinosa:
## Description
This ticket is meant to track the progress in making a drop-in replacement of `nav2_amcl` in terms of interface and parameters.
Specific features and requirements will be best tracked in their own issues.

## Definition of done

Complete the implementation of the following list of parameters, services, publishers and subscribers:
- Parameters
  - ROS topics and frames
    - [x] `base_frame_id`
    - [x] `global_frame_id`
    - [x] `odom_frame_id`
    - [x] `scan_topic`
    - [x] `map_topic`
    - [x] `initial_pose_topic` (extra)
  - Initial pose (see #41)
    - [x] `set_initial_pose`
    - [x] `initial_pose`
    - <del>`save_pose_rate`</del>
    - [x] `always_reset_initial_pose`
    - [x] `first_map_only`
  - Publish map to odom transform (see #40)
    - [x] `tf_broadcast`
    - [x] `transform_tolerance`
  - Particle filter
    - [x] `max_particles`
    - [x] `min_particles`
    - [x] `pf_err`
    - [x] `pf_z`
    - [x] `recovery_alpha_fast`
    - [x] `recovery_alpha_slow`
    - [x] `spatial_resolution` (extra)
    - [x] `resample_interval` (see #57)
    - [x] `selective_resampling` (see #57)
  - Motion model
    - [x] `robot_model_type`
    - [x] `alpha1`
    - [x] `alpha2`
    - [x] `alpha3`
    - [x] `alpha4`
    - [x] `alpha5` (see #56)
    - [x] `update_min_a`
    - [x] `update_min_d`
  - Sensor model
    - [x] `laser_model_type`
    - [x] `laser_max_range`
    - [x] `laser_min_range`
    - [x] `max_beams` (see #84) 
    - [x] `sigma_hit`
    - [x] `z_hit`
    - [x] `z_rand`
  - Likelihood field model exclusive
    - [x] `laser_likelihood_max_dist`
    - [ ] `beam_skip_distance`
    - [ ] `beam_skip_error_threshold`
    - [ ] `beam_skip_threshold`
    - [ ] `do_beamskip` (see #187)
  - Beam model exclusive (see #54)
    - [x] `lambda_short`
    - [x] `z_max`
    - [x] `z_short`
- Services
  - [x] `reinitialize_global_localization`
  - [ ] `request_nomotion_update` (see #184)
[...]

PR #189, opened 2023-05-16
Add support for map updates in laser localization filters

hidmic:
Related to #85.

## Summary

This patch exposes a map update API for `beluga::LaserLocalizationInterface2d` compatible filters, and makes use of it in `beluga_amcl` to support map updates on an existing particle filter. Built on top of #188.

## Checklist
- [x] Read the [contributing guidelines](https://github.com/Ekumen-OS/beluga/blob/main/CONTRIBUTING.md).
- [x] Configured pre-commit and ran colcon test locally.
- [x] Signed all commits for DCO.
- [x] Added tests (regression tests for bugs, coverage of new code for features).
- [x] Updated documentation (as needed).
- [x] Checked that CI is passing.

### DIFF
--- a/beluga/include/beluga/sensor/likelihood_field_model.hpp
+++ b/beluga/include/beluga/sensor/likelihood_field_model.hpp
@@ -97,9 +97,7 @@ class LikelihoodFieldModel : public Mixin {
       : Mixin(std::forward<Args>(rest)...),
         params_{params},
         grid_{std::move(grid)},
-        free_states_{
-            grid_.coordinates_for(grid_.free_cells(), OccupancyGrid::Frame::kGlobal) |
-            ranges::to<std::vector<Eigen::Vector2d>>},
+        free_states_{compute_free_states(grid_)},
         likelihood_field_{make_likelihood_field(params, grid_)} {}
 
   /// Returns the likelihood field, constructed from the provided map.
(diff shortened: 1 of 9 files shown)


nahueespinosa (approved):
LGTM!

beluga_amcl/src/amcl_node.cpp
```diff
@@ -852,13 +838,44 @@ void AmclNode::map_callback(nav_msgs::msg::OccupancyGrid::SharedPtr map)
...
+      reinitialize_with_pose(pose, covariance);
+      should_reset_initial_pose = false;
+    }
+  } else {
+    particle_filter_->update_map(OccupancyGrid{std::move(map)});
+    should_reset_initial_pose = get_parameter("always_reset_initial_pose").as_bool();
+  }
+
+  if (should_reset_initial_pose && get_parameter("set_initial_pose").as_bool()) {
     const auto pose = Sophus::SE2d{
```
olmerg, line 879: when the node is deactivated the particleFilter save the state , which is different of the state which is save in the parameters. The question is when is better to use the particlefilter state, because in nav2_amcl the parameters are saved during deactivate state (in theory).

hidmic: Not sure what you mean @olmerg. We allow all combinations AFAICS.

olmerg: ohh I understand with `last_known_estimate_.has_value()` is covered the case, great :)

hidmic:
Rebased to solve merge conflicts.

beluga/include/beluga/sensor/likelihood_field_model.hpp
```diff
@@ -153,17 +152,30 @@ class LikelihoodFieldModel : public Mixin {
...
     points_ = std::move(points);
   }
 
+  /// \copydoc LaserSensorModelInterface2d::update_map(Map&& map)
+  void update_map(OccupancyGrid&& map) final {
+    const auto lock = std::lock_guard<std::shared_mutex>{mutex_};
+    grid_ = std::move(map);
+    free_states_ = compute_free_states(grid_);
+    likelihood_field_ = make_likelihood_field(params_, grid_);
+  }
```
nahueespinosa, line 165: It seems that adding a test for this method should make the codecov status check green.

https://app.codecov.io/gh/Ekumen-OS/beluga/pull/189/blob/beluga/include/beluga/sensor/likelihood_field_model.hpp#L160

hidmic: Added and rebased to solve conflicts. Is there no way to run coverage checks locally?

nahueespinosa: There isn't. I plan to add it soon. Feel free to create a ticket if you think it needs to be tracked.

nahueespinosa (approved):

PR #195, opened 2023-05-23
Remove shared_mutexes to address concurrency performance

glpuga:
### Proposed changes

Removes the `shared_mutex` locks, that were not currently in use since there's no concurrency in how different steps of the filter are applied, but that was causing performance issues.

#### Type of change

### Checklist

- [ ] All commmits have been signed for [DCO](https://developercertificate.org/)

### Additional comments

Pre (before change) and post (after change) performance of the node.

#### Likelihood Sensor Model (linear scale)

likelihood_linear

#### Likelihood Sensor Model (log scale)

likelihood_log

#### Beam Sensor Model (linear scale)

beam_linear

#### Beam Sensor Model (log scale)

beam_log

### DIFF
--- a/beluga_benchmark/beluga_benchmark/compare_results.py
+++ b/beluga_benchmark/beluga_benchmark/compare_results.py
@@ -118,32 +118,55 @@ def main():
         ' implementation (e.g. nav2_amcl, another beluga version, etc).'
     )
     arg_parser.add_argument(
-        '--beluga-results',
-        '-b',
+        '--series',
+        '-s',
         type=Path,
-        help='Folder with parameterized beluga benchmark results',
+        action='append',
+        help='Folder with parameterized benchmark results',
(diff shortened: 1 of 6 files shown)


beluga_benchmark/beluga_benchmark/compare_results.py
```diff
@@ -118,32 +118,55 @@ def main():
...
-        '-o',
-        type=Path,
-        help='Folder with parameterized benchmark results for the other implementation',
+        '--label',
+        '-l',
+        type=str,
+        action='append',
+        help='Label for a series',
         required=True,
     )
```
glpuga, line 135: I still need to add this change to the documentation of the script.

ivanpauno (approved):
Besides the updates needed in the readme, LGTM!

PR #198, opened 2023-05-25
Replace perf with faster patched built-from-source version

glpuga:
### Proposed changes

Replaces the `perf` installation with a built-from-source version that's patched to be much, much faster. This will make it much easier to analyze and compare the performance of different versions. After this patch, the flamegraph takes about 8 minutes to be generated on an `Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz`.

The patched version is the one described in [here](https://github.com/flamegraph-rs/flamegraph/issues/74#issuecomment-1529140274).

#### Type of change

### Checklist

### Additional comments

While I could have opted for not installing `linux-tools-generic` in the first place, instead of installing and the removing before patching, I decided to execute the process from the same starting point used in the instructions in https://github.com/flamegraph-rs/flamegraph/issues/74#issuecomment-1529140274 , which assume `linux-tools-generic` (and the packages it depends on) to have been previously installed.

#### Reference results

The performance results for both the `likelihood` and `beam` configurations can be seen below.

In all cases, the results were obtained by profiling e72e148 with the step-by-step instructions from `PROFILING.md`.

The computer is an `Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz`.

#### Beam sensor model

flamegraph_beam

#### Likelihood sensor model

flamegraph_likelihood

### DIFF
--- a/beluga/include/beluga/localization.hpp
+++ b/beluga/include/beluga/localization.hpp
@@ -52,12 +52,16 @@ namespace beluga {
  */
 
 /// Pure abstract class representing an interface for laser-based localization algorithms.
+/**
+ * \tparam Map Environment representation type.
+ */
+template <class Map>
 using LaserLocalizationInterface2d = beluga::mixin::compose_interfaces<
     BaseParticleFilterInterface,
     StorageInterface<Sophus::SE2d, beluga::Weight>,
     EstimationInterface2d,
     OdometryMotionModelInterface2d,
-    LaserSensorModelInterface2d>;
+    LaserSensorModelInterface2d<Map>>;
 
 /// Commonly used resampling policies combined in a single mixin.
 using CombinedResamplingPolicy = ciabatta::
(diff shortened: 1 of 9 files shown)


docker/files/srcline.c
```diff
@@ -0,0 +1,920 @@
+// SPDX-License-Identifier: GPL-2.0
```
hidmic, line 1: @glpuga the GPL is pervasive, quoting:

> 2. You may modify your copy or copies of the Program or any portion of it [...] provided that you also meet all of these conditions:
> b) You must cause any work that you distribute or publish, that in whole or **in part contains** or is derived from the Program or any part thereof, to be licensed as a whole at no charge to all third parties under the terms of this License.

There _is_ a notion of separate work in GPL terms and conditions, but I'm no lawyer, and this smells as a liability. How about we relocate this to a different repository? A minimal Github Actions workflow to push to Dockerhub should do to make it available to Beluga.

glpuga: Hmm... good catch. I flew totally past me checking the license implications of adding the file replacement.

This is not essential, so I'll take down this PR while we figure if it's worth adding it back in some other form.

PR #199, opened 2023-05-26
Optimize likelihood field model runtime performance

glpuga:
### Proposed changes

A number of optimizations to functions in the tightest loops of the Likelihood Sensor model.

#### Type of change

### Checklist

### Additional comments

**This is working draft.  The pre-cubing of the likelihood values broke tests, I haven't yet spent time fixing them.**

### Results

All performance evaluations were performed on a `Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz`.

#### Comparison of pre and post optimization performance

optimized_vs_nonoptimized

Pre-optimization flamegraph:

pre-optimization

Post-optimization flamegraph:

post-optimization

#### Optimized code vs. Nav2 AMCL

All three (Beluga multi-threaded, Beluga single-threaded and Nav2 AMCL):

optimized_vs_amcl_all

Closer detail of Beluga single-threaded vs. Nav2 AMCL:

optimized_vs_amcl_seq


beluga/include/beluga/sensor/data/regular_grid.hpp
```diff
@@ -74,8 +74,12 @@ class BaseRegularGrid2 : public ciabatta::ciabatta_top<Derived> {
...
    */
   [[nodiscard]] Eigen::Vector2i cell_near(double x, double y) const {
-    const auto xi = static_cast<int>(std::floor(x / this->self().resolution()));
-    const auto yi = static_cast<int>(std::floor(y / this->self().resolution()));
+    const auto inv_resolution = 1 / this->self().resolution();
+    const auto scaled_x = x * inv_resolution;
+    const auto scaled_y = y * inv_resolution;
+    // don't use floor() due to performance reasons
+    const auto xi = static_cast<int>(scaled_x < 0.0 ? -2.0 : scaled_x);
+    const auto yi = static_cast<int>(scaled_y < 0.0 ? -2.0 : scaled_y);
```
nahueespinosa, line 82: Hmm, I don't follow why there is a `-2.0` in there.

nahueespinosa (commented):
@glpuga Nice!

beluga/include/beluga/sensor/data/regular_grid.hpp
glpuga: No particular reason. Any negative that does not cast into 0 will do, because we don't care about negatives since they are not in the grid. This would probably work as well:
```
const auto xi = static_cast<int>(scaled_x < 0.0 ? scaled_x - 1.0 : scaled_x);
```
but [it's not exactly the same](https://stackoverflow.com/a/825916) as `std::floor` and wastes cycles on negatives.

nahueespinosa: Makes sense, can we document the return value in the doxygen block of this method? And adding a local comment on why we use `-2.0` wouldn't hurt either.

glpuga: Indeed. I'll add it.

glpuga: Thinking it over, 
```
const auto xi = scaled_x < 0.0 ?  -1 : static_cast<int>(scaled_x);
```
looks far less esoteric and should be just as performant. Still requires comments though.

beluga/include/beluga/sensor/data/regular_grid.hpp
```diff
@@ -74,8 +74,12 @@ class BaseRegularGrid2 : public ciabatta::ciabatta_top<Derived> {
    * \return Grid cell coordinates.
    */
   [[nodiscard]] Eigen::Vector2i cell_near(double x, double y) const {
```
glpuga, line 76: @hidmic why is it important that this function is surjective? Is the intended co-domain all of the discrete plain or only the valid grid?

beluga/include/beluga/sensor/data/regular_grid.hpp
```diff
@@ -67,15 +67,25 @@ class BaseRegularGrid2 : public ciabatta::ciabatta_top<Derived> {
...
-    const auto xi = static_cast<int>(std::floor(x / this->self().resolution()));
-    const auto yi = static_cast<int>(std::floor(y / this->self().resolution()));
+    const auto inv_resolution = 1 / this->self().resolution();
+    const auto scaled_x = x * inv_resolution;
+    const auto scaled_y = y * inv_resolution;
+    // Truncate towards zero. Use a poor man's replacement of
+    // std::floor for coodinates in the positive quadrant because
```
glpuga, line 83:
```suggestion
    // std::floor for coordinates in the positive quadrant because
```

beluga/include/beluga/sensor/data/regular_grid.hpp
```diff
@@ -67,15 +67,25 @@ class BaseRegularGrid2 : public ciabatta::ciabatta_top<Derived> {
...
+    // the `std::floor` is awfully slow and this function is
```
glpuga, line 84:
```suggestion
    // `std::floor` is awfully slow and this function is
```

beluga/include/beluga/sensor/data/regular_grid.hpp
```diff
@@ -67,15 +67,25 @@ class BaseRegularGrid2 : public ciabatta::ciabatta_top<Derived> {
  public:
   /// Compute nearest grid cell coordinates given plane coordinates.
   /**
-   * Note this is a surjective function.
+   * For performance reasons this function is optimized for coordiantes in the
```
hidmic, line 70: @glpuga nit: 
```suggestion
   * For performance reasons this function is optimized for coordinates in the
```

beluga/include/beluga/sensor/data/regular_grid.hpp
hidmic: It's not a requirement as much as a warning. Multiple $(x, y) \in \mathbb{R}^2$ tuples will map to the same $(x_i, y_i) \in \mathbb{Z}^2$ tuple.

> Is the intended co-domain all of the discrete plane or only the valid grid?

For a regular grid, the entire plane, as there is no notion of extent.

beluga/include/beluga/sensor/data/regular_grid.hpp
```diff
@@ -67,15 +67,25 @@ class BaseRegularGrid2 : public ciabatta::ciabatta_top<Derived> {
...
+   * quadrants a representative cell in the same quadrant will be returned instead.
    *
    * \param x Plane x-axis coordinate.
    * \param y Plane y-axis coordinate.
    * \return Grid cell coordinates.
    */
   [[nodiscard]] Eigen::Vector2i cell_near(double x, double y) const {
-    const auto xi = static_cast<int>(std::floor(x / this->self().resolution()));
-    const auto yi = static_cast<int>(std::floor(y / this->self().resolution()));
+    const auto inv_resolution = 1 / this->self().resolution();
```
hidmic, line 79: @glpuga nit: avoid implicit promotion
```suggestion
    const auto inv_resolution = 1.0 / this->self().resolution();
```

beluga/include/beluga/sensor/likelihood_field_model.hpp
```diff
@@ -130,20 +130,26 @@ class LikelihoodFieldModel : public Mixin {
...
+    const auto y_offset = transform.translation().y();
+    const auto cos_theta = transform.so2().unit_complex().x();
+    const auto sin_theta = transform.so2().unit_complex().y();
+    const auto unknown_space_occupancy_prob = 1. / params_.max_laser_distance;
     // TODO(glpuga): Investigate why AMCL and QuickMCL both use this formula for the weight.
     // See https://github.com/Ekumen-OS/beluga/issues/153
+    const auto unknown_space_occupancy_likelihood_cubed =
+        unknown_space_occupancy_prob * unknown_space_occupancy_prob * unknown_space_occupancy_prob;
     return std::transform_reduce(
         points_.cbegin(), points_.cend(), 1.0, std::plus{},
```
hidmic, line 143: @glpuga I saw this before and I figured I would ask. Why do we add up from 1.0? Shouldn't it be 0.0?

beluga/include/beluga/sensor/data/regular_grid.hpp
```diff
@@ -67,15 +67,25 @@ class BaseRegularGrid2 : public ciabatta::ciabatta_top<Derived> {
...
+    const auto scaled_x = x * inv_resolution;
+    const auto scaled_y = y * inv_resolution;
+    // Truncate towards zero. Use a poor man's replacement of
+    // std::floor for coodinates in the positive quadrant because
+    // the `std::floor` is awfully slow and this function is
+    // performance critical. We don't care about coordinates in other
+    // quadrants so we just return some equally invalid representative.
+    const auto xi = scaled_x < 0.0 ? -1 : static_cast<int>(scaled_x);
+    const auto yi = scaled_y < 0.0 ? -1 : static_cast<int>(scaled_y);
```
hidmic, line 88: @glpuga wouldn't:
```suggestion
    const auto xi = static_cast<int>(scaled_x) - (scaled_x < 0.0);
    const auto yi = static_cast<int>(scaled_y) - (scaled_y < 0.0);
```
be a correct `floor` implementation at a negligible extra cost (that of two integer substractions)?

beluga/include/beluga/sensor/likelihood_field_model.hpp
```diff
@@ -130,20 +130,26 @@ class LikelihoodFieldModel : public Mixin {
...
           // Transform the end point of the laser to the grid local coordinate system.
           // Not using Eigen/Sophus because they make the routine x10 slower.
           // See `benchmark_likelihood_field_model.cpp` for reference.
           const auto x = point.first * cos_theta - point.second * sin_theta + x_offset;
           const auto y = point.first * sin_theta + point.second * cos_theta + y_offset;
-          const auto pz = likelihood_field_.data_near(x, y).value_or(unknown_space_occupancy_prob);
-          return pz * pz * pz;
+          // for performance, we store the likelihood already elevated to the cube
+          const auto pz_cubed = likelihood_field_.data_near(x, y).value_or(unknown_space_occupancy_likelihood_cubed);
+          return pz_cubed;
```
hidmic, line 152: @glpuga nit:
```suggestion
          return likelihood_field_.data_near(x, y).value_or(unknown_space_occupancy_likelihood_cubed);
```

hidmic (commented):
Nice!

beluga/include/beluga/sensor/likelihood_field_model.hpp
glpuga: I agree that would be the logical thing to do, but [that's not what Nav2 AMCL does](https://github.com/Ekumen-OS/beluga/pull/149#discussion_r1152578137).

beluga/include/beluga/sensor/data/regular_grid.hpp
glpuga: I can't think of a case where it would not work. At least it does not have the problem of https://stackoverflow.com/questions/824118/why-is-floor-so-slow/825916#825916

Code looks good too; it's branch-less and the compiler exploits that to vectorize lines 87 and 88.
https://godbolt.org/z/Gzssvcr76

beluga/include/beluga/sensor/data/regular_grid.hpp
glpuga: Fixed.

beluga/include/beluga/sensor/data/regular_grid.hpp
glpuga: Fixed.

beluga/include/beluga/sensor/data/regular_grid.hpp
glpuga: I'm wrong. It does provide the wrong answer right in the threshold of negative integers. It's the same problem as the first link.

But as far as wrong answers go for out-of-grid coordinates, I like the compiled code of this version better.

beluga/include/beluga/sensor/likelihood_field_model.hpp
glpuga: Done.

beluga/include/beluga/sensor/data/regular_grid.hpp
glpuga: With the new, closer to the original, replacement of floor() I don't need to change this comment.

beluga/include/beluga/sensor/data/regular_grid.hpp
```diff
@@ -74,8 +74,14 @@ class BaseRegularGrid2 : public ciabatta::ciabatta_top<Derived> {
...
-    const auto xi = static_cast<int>(std::floor(x / this->self().resolution()));
-    const auto yi = static_cast<int>(std::floor(y / this->self().resolution()));
+    const auto inv_resolution = 1. / this->self().resolution();
+    const auto scaled_x = x * inv_resolution;
+    const auto scaled_y = y * inv_resolution;
+    // Truncate towards zero. Use a poor man's replacement of
+    // `std::floor` because the later is awfully slow and this
+    // function is performance critical.
+    const auto xi = static_cast<int>(scaled_x) - (scaled_x < 0.0);
+    const auto yi = static_cast<int>(scaled_y) - (scaled_y < 0.0);
```
hidmic, line 84: Hmm, the following should make it right for all negative integers, at the expense of breaking it for all positive numbers closer than an epsilon to the immediately greater integer:
```suggestion
    const auto xi = static_cast<int>(scaled_x + std::numeric_limits<double>::epsilon()) - (scaled_x < 0.0);
    const auto yi = static_cast<int>(scaled_y + std::numeric_limits<double>::epsilon()) - (scaled_y < 0.0);
```

hidmic (approved):

glpuga:
The code after review changes still shows an advantage over Nav2 AMCL.

### All three (Beluga multi-threaded, Beluga single-threaded and Nav2 AMCL):
beluga_vs_amcl_multithreaded


### Closer detail of Beluga single-threaded vs. Nav2 AMCL (log scale):
beluga_vs_amcl


### Closer detail of Beluga single-threaded vs. Nav2 AMCL (linear scale):
beluga_vs_amcl_linear

beluga/include/beluga/sensor/data/regular_grid.hpp
```diff
@@ -74,8 +74,14 @@ class BaseRegularGrid2 : public ciabatta::ciabatta_top<Derived> {
...
    */
   [[nodiscard]] Eigen::Vector2i cell_near(double x, double y) const {
-    const auto xi = static_cast<int>(std::floor(x / this->self().resolution()));
-    const auto yi = static_cast<int>(std::floor(y / this->self().resolution()));
+    const auto inv_resolution = 1. / this->self().resolution();
+    const auto scaled_x = x * inv_resolution;
+    const auto scaled_y = y * inv_resolution;
+    // Truncate towards zero. Use a poor man's replacement of
+    // `std::floor` because the later is awfully slow and this
+    // function is performance critical.
```
ivanpauno, line 82: This is not truncating towards zero, but instead returning the greatest integer not greater than scaled_x/scaled_y (which for positive numbers is the same)

```suggestion
    // Use a poor man's replacement of `std::floor` because
    // the later is awfully slow and this function is performance
    // critical.
```

beluga/include/beluga/sensor/data/regular_grid.hpp
ivanpauno: can actually x and y be negative?
From a quick read to the code, it seems that they shouldn't be.

In that case, maybe we could try `std::trunc`?

glpuga: I can be negative, yes. It'll be discarded later because it falls outside of the grid.

I did not try `trunc` because most likely is as slow as `floor`. The problem with those functions is not with what they do, but in all the additional checks they need to do to be IEEE-754-compliant.

glpuga: `trunc` would also map the (-1.0, 0.0) range to 0, which we'd like to avoid.

ivanpauno: Maybe:

```cpp
static double opt_floor_impl(double x) {
  const auto int_x = static_cast<int>(x);
  bool is_not_integral = static_cast<double>(int_x) != x;
  return x - static_cast<double>(x < 0.0 && is_not_integral);
}
```

that should solve @hidmic comment and it's still maybe faster than the std version ....

glpuga: Given that in this case we only need the `floor`-like behavior to be correct in the ++ quadrant, I propose we leave the current version. 

Still, I think it's worth ticketing looking into other places where we may be using `std::floor` and familiy and determining if replacing those calls for a cheaper `floor`-like function. In this particular case, the function was the main cause of performance loss, and yet the functioin was invisible in the stack trace in perf. We may have further "performance leaks" in other places.

ivanpauno: > Given that in this case we only need the floor-like behavior to be correct in the ++ quadrant, I propose we leave the current version.

I see, makes sense.

glpuga: I created this ticket to track this: https://github.com/Ekumen-OS/beluga/issues/206

Issue #170, opened 2023-05-03
Generalize planar laser scan support


hidmic:
## Description

Currently, `beluga_amcl` supports `sensor_msgs/msg/LaserScan` messages by transforming them to an unorganized pointcloud of sorts, using cartesian coordinates w.r.t. to the robot frame. This is at best unnecessary overhead for sensor models that perform ray tracing, which could use polar coordinates directly, and worse case a source of error for such models, as recomputing ray bearings from the pointcloud w.r.t. to a different frame will skew them.

#168 already attempts a partial generalization of laser scan data storage and representation. I think we should take it to completion.

## Definition of done

- [ ] Extract planar laser scan generalization from #168.
- [ ] Address the [inverted laser scan scenario](https://github.com/ekumenlabs/beluga/pull/168#discussion_r1183061756).

hidmic:
This is done as of #271.

PR #200, opened 2023-05-28
Optimize beam sensor model runtime performance

glpuga:
### Proposed changes

Minor optimizations to the tightest execution loops in the beam sensor model.

While these changes do improve performance a bit, they barely make a difference in the performance disadvantage we have against Nav2 AMCL. We are still missing something much bigger than these optimizations.

#### Type of change

### Checklist

### Additional comments

These changes change the performance profile, as seen through `perf` from this:

baseline_flamegraph

to this:

commit6_flamegraph

**Note: These flamegraphs assume the changes in #199 have already been merged, since there are grid optimizations that are common to both Likelihood and Beam.**

Notice that relative to the unmodified the `beluga::Bresenham2i::Line` block of code (that had no changes done to it, neither in performance per execution nor in total number of executions), the overall time spent in the  `importance_weight` function seems to have reduced significantly.

Notice also the removal of the second stack call tower on the right, which appears to be related to queuing in the `MessageFilter`. 

However, these changes barely changed the cpu usage profile:

beam_vs_beluga_par

beam_vs_beluga_seq

And the change is barely noticeable in the difference against `Nav2 AMCL`.

beam_vs_beluga_vs_nav2_seq

While it's still possible our implementation of Bresenham is less peforming than Nav2's, even if we somehow reduced the tracing runtime cost to 0 with some magic implementation, that would still get us to perform at basically the same level as `Nav2 amcl`. To me that indicates that the problem is somewhere else. 

I suspect we are actually _doing more work_ than Nav2, but I haven't been able to find any proof of that.

Further work is still needed.

### DIFF
--- a/beluga/include/beluga/algorithm/raycasting.hpp
+++ b/beluga/include/beluga/algorithm/raycasting.hpp
@@ -77,10 +77,11 @@ class Ray2d {
    *   That is, regardless of grid cells' state.
    */
   [[nodiscard]] auto trace(const Sophus::SO2d& bearing) const {
-    const auto far_end_pose_in_source_frame = Sophus::SE2d{
-        Sophus::SO2d{0.},
-        Eigen::Vector2d{max_range_ * bearing.unit_complex().x(), max_range_ * bearing.unit_complex().y()}};
-    const auto far_end_pose_in_local_frame = source_pose_in_local_frame_ * far_end_pose_in_source_frame;
+    // Don't do the multiplication in SE2, it's slower
+    const auto& r1 = source_pose_in_local_frame_.so2();
+    const auto& t1 = source_pose_in_local_frame_.translation();
+    const auto t2 = bearing.unit_complex() * max_range_;
+    const auto far_end_pose_in_local_frame = Sophus::SE2d{r1, r1 * t2 + t1};
     const auto far_end_cell = grid_.cell_near(far_end_pose_in_local_frame.translation());
     const auto cell_is_valid = [this](const auto& cell) { return grid_.contains(cell); };
     return algorithm_(source_cell_, far_end_cell) | ranges::views::take_while(cell_is_valid);
--- a/beluga/include/beluga/sensor/beam_model.hpp
+++ b/beluga/include/beluga/sensor/beam_model.hpp
@@ -122,42 +122,47 @@ class BeamSensorModel : public Mixin {
    */
   [[nodiscard]] weight_type importance_weight(const state_type& state) const {
     const auto beam = Ray2d{grid_, state, params_.beam_max_range};
-    return std::transform_reduce(points_.cbegin(), points_.cend(), 0.0, std::plus{}, [this, &beam](const auto& point) {
-      // TODO(Ramiro): We're converting from range + bearing to cartesian points in the ROS node, but we want range +
-      // bearing here. We might want to make that conversion in the likelihood model instead, and let the measurement
-      // type be range, bearing instead of x, y.
+    const double n = 1. / (std::sqrt(2. * M_PI) * params_.sigma_hit);
+    return std::transform_reduce(
+        points_.cbegin(), points_.cend(), 0.0, std::plus{}, [this, &beam, n](const auto& point) {
+          // TODO(Ramiro): We're converting from range + bearing to cartesian points in the ROS node, but we want range
+          // + bearing here. We might want to make that conversion in the likelihood model instead, and let the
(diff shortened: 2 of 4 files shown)


beluga/include/beluga/sensor/beam_model.hpp
```diff
@@ -122,42 +122,47 @@ class BeamSensorModel : public Mixin {
...
+          // + bearing here. We might want to make that conversion in the likelihood model instead, and let the
+          // measurement type be range, bearing instead of x, y.
+
+          // Compute the range according to the measurement.
+          const double z = std::sqrt(point.first * point.first + point.second * point.second);
+
+          // dirty hack to prevent SO2d from calculating the hypot again to normalize the vector.
+          auto beam_bearing = Sophus::SO2d{};
+          beam_bearing.data()[0] = point.first / z;
+          beam_bearing.data()[1] = point.second / z;
```
hidmic, line 138: @glpuga meta: these computations are  something that we know we shouldn't be doing, and will go away once https://github.com/Ekumen-OS/beluga/issues/170 is addressed (and the TODO is resolved).

beluga/include/beluga/sensor/beam_model.hpp
```diff
@@ -122,42 +122,47 @@ class BeamSensorModel : public Mixin {
...
+
+          // Compute range according to the map (raycasting).
+          const double z_mean = beam.cast(beam_bearing).value_or(params_.beam_max_range);
+          // 1: Correct range with local measurement noise.
+          const double eta_hit =
+              2. / (std::erf((params_.beam_max_range - z_mean) / (std::sqrt(2.) * params_.sigma_hit)) -
+                    std::erf(-z_mean / (std::sqrt(2.) * params_.sigma_hit)));
```
hidmic, line 145: @glpuga it's worth pointing out that `nav2_amcl` doesn't normalize like we do. All these $\eta$ values are assumed to be 1.

hidmic (approved):
LGTM

beluga/include/beluga/algorithm/raycasting.hpp
```diff
@@ -77,10 +77,11 @@ class Ray2d {
...
    */
   [[nodiscard]] auto trace(const Sophus::SO2d& bearing) const {
-    const auto far_end_pose_in_source_frame = Sophus::SE2d{
-        Sophus::SO2d{0.},
-        Eigen::Vector2d{max_range_ * bearing.unit_complex().x(), max_range_ * bearing.unit_complex().y()}};
-    const auto far_end_pose_in_local_frame = source_pose_in_local_frame_ * far_end_pose_in_source_frame;
+    // Don't do the multiplication in SE2, it's slower
+    const auto& r1 = source_pose_in_local_frame_.so2();
+    const auto& t1 = source_pose_in_local_frame_.translation();
+    const auto t2 = Eigen::Vector2d{max_range_ * bearing.unit_complex().x(), max_range_ * bearing.unit_complex().y()};
```
nahueespinosa, line 83: nit:
```suggestion
    const auto t2 = bearing.unit_complex() * max_range_;
```

beluga/include/beluga/sensor/data/regular_grid.hpp
```diff
@@ -99,9 +99,8 @@ class BaseRegularGrid2 : public ciabatta::ciabatta_top<Derived> {
    * \return Plane coordinates of the cell centroid.
    */
   [[nodiscard]] Eigen::Vector2d coordinates_at(int xi, int yi) const {
-    return Eigen::Vector2d{
-        (static_cast<double>(xi) + 0.5) * this->self().resolution(),
-        (static_cast<double>(yi) + 0.5) * this->self().resolution()};
+    const auto resolution = this->self().resolution();
+    return Eigen::Vector2d{(static_cast<double>(xi) + 0.5) * resolution, (static_cast<double>(yi) + 0.5) * resolution};
```
nahueespinosa, line 103: nit:
```suggestion
    return Eigen::Vector2d{(static_cast<double>(xi) + 0.5), (static_cast<double>(yi) + 0.5)} * resolution;
```

nahueespinosa (commented):
@glpuga Left two super minor comments. It is also worth noting that `RelWithDebInfo` reduces the optimization level to `-O2` instead of `-O3`, so flamegraph might not be as good at comparing micro-optimizations results like these.

Microbenchmarks using [`googlebenchmark`](https://github.com/google/benchmark) compiled in `Release` mode may be better at detecting true progress.

nahueespinosa (approved):

PR #197, opened 2023-05-25
Re-organize infrastructure scripts

nahueespinosa:
### Proposed changes

This patch groups all infrastructure-related scripts in one place and provides a way to generate code coverage reports locally.

#### Type of change

- [ ] 🐛 Bugfix (change which fixes an issue)
- [x] 🚀 Feature (change which adds functionality)
- [x] 📚 Documentation (change which fixes or extends documentation)

### Checklist

- [x] Lint and unit tests (if any) pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] All commmits have been signed for [DCO](https://developercertificate.org/)

PR #202, opened 2023-05-30
Fix dockerfile and add docker pipeline

nahueespinosa:
### Proposed changes

I was too confident not to rebuild the latest changes to the Dockerfile before merging #197. :sweat_smile: 

#### Type of change

### Checklist


.github/workflows/docker_ci_pipeline.yml
```diff
@@ -0,0 +1,37 @@
...
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Build
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: docker/images/${{ matrix.ros_distro }}/Dockerfile
```
hidmic, line 35: @nahueespinosa shouldn't we specify the [platforms](https://docs.docker.com/build/ci/github-actions/multi-platform/) we want to build for? If not, why install QEMU in the first place? Also, if we are targeting multiple platforms, shouldn't we target multiple platforms for CI? I think we should, but I sense it is out of scope for this PR.

nahueespinosa: I added the specified `linux/amd64` in https://github.com/Ekumen-OS/beluga/pull/202/commits/bca95e21d030e5498325675c396c982d0eae52b2.

> Also, if we are targeting multiple platforms, shouldn't we target multiple platforms for CI?

I agree that it is out of scope. In that case, we may want to run tests on different platforms, not just create a docker image.

hidmic (approved):

ivanpauno (approved):
LGTM!

.github/workflows/docker_ci_pipeline.yml
```diff
@@ -0,0 +1,43 @@
...
+        platform:
+          - linux/amd64
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
```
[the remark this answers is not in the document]
nahueespinosa: We don't, but it doesn't hurt to keep.

PR #204, opened 2023-05-30
Install doxygen-awesome

nahueespinosa:
### Proposed changes

Installs and configures https://jothepro.github.io/doxygen-awesome-css/.
This modern theme updates the visuals of Doxygen without changing its overall layout too much.

#### Type of change

### Checklist

## Additional comments

Here is the output of the documentation build step: [docs.zip](https://github.com/Ekumen-OS/beluga/files/11599991/docs.zip)


nahueespinosa:
## Before:
Screenshot from 2023-05-30 09-37-45

## After:
Screenshot from 2023-05-30 09-37-29

.gitmodules
```diff
@@ -2,3 +2,6 @@
   path = flatland
   url = git@github.com:avidbots/flatland.git
   branch = ros2-humble
+[submodule "beluga/docs/doxygen-awesome-css"]
+  path = beluga/docs/doxygen-awesome-css
+  url = git@github.com:jothepro/doxygen-awesome-css.git
```
ivanpauno, line 7: I'm not a big fan of adding everything as a submodule.
For build dependencies I think it's better to use something like `vcs`.

In this case, we could pull the dependency in the CMakeLists that build the docs (with ExternalProject_Add or FetchContent).

ivanpauno (approved):
LGTM!

.gitmodules
nahueespinosa: Done in https://github.com/Ekumen-OS/beluga/pull/204/commits/6a765db8bb518d6c6c3db6a07241b7778e4b3b3a.

ivanpauno (approved):

PR #207, opened 2023-06-03
Updates and fixes before updating performance report

glpuga:
### Proposed changes

- Changes the tolerance value for the MessageFilter to match the value used by Nav2 (the same value as `transform_tolerance`).
- Fixes a problem in the `compare_results.py` script that causes it to fail to plot a whole series of data if a single bad sample is present (this happens frequently when capturing AMCL performance data in the high-end of particle counts).
- Fix a mismatch in the naming of the `playback_rate` command line argument of the `parameterized_run` script and the launchfile.
- Changes the default playback rate to be 1x to avoid distorting the captured CPU stats.

#### Type of change

### Checklist

### Additional comments

The following plots show that the performance at 1x replay rate both before and after the change of the MessageFilter tolerance is the same.

likelihood_before_vs_after

beam_before_vs_after


nahueespinosa (approved):
LGTM!

nit^n: Use imperative mood in the subject line of the commit message.

Issue #95, opened 2023-02-02
Beluga for ROS 1 Noetic


glpuga:
## Description

Since most robotic systems in production today are still using ROS 1, and it can be expected that they'll continue to do so for some time, creating a ROS 1 version of Beluga may help with adoption and increase the chances of people trying it in current systems.

## Definition of done

A version of beluga_amcl that can be build and run in ROS 1 Noetic exists.

## Additional considerations

Noetic and ROS 1 are only 2.5 years away from EOL, but since ROS 1 is still being widely used, the amount of immediate users probably compensates for the short life cycle when evaluating this effort.

hidmic:
I've started drafting this, using build system conditionals. It'll take me a while though.

PR #210, opened 2023-06-05
Add ROS Noetic support to `beluga_amcl`

hidmic:
### Proposed changes

First partial take on #95. Depends on #209. Both `beluga_amcl` and `beluga_example` packages have been adapted to work under both ROS major versions. Documentation and CI need an update to reflect ROS 1 support.

To test locally:

```sh
ROSDISTRO=noetic ./docker/run.sh --build
colcon build --packages-up-to beluga_example
source install/setup.bash
roslaunch beluga_example example_rosbag.launch
```

#### Type of change

### Checklist

### Additional comments

A few decisions pending:

- Do we want CI for ROS Noetic? I'm inclined to say yes, but it will add to CI job durations. Either way, we probably need to sort out how we compute coverage. I haven't delved into it yet. 
- I have not migrated `beluga_benchmark` nor `beluga_system_tests` yet. I'm on the fence as to whether it makes sense to do so (ie. how much value it brings for the maintenance cost it will add).
- Demos using Flatland have not been migrated, as we currently include a ROS 2 Humble compatible version as a submodule in this repository. Perhaps it's time to consider using [`vcstool`](https://github.com/dirk-thomas/vcstool) (or [others](http://wiki.ros.org/rosinstall_generator)) to handle dependencies.


hidmic:
Also, I haven't tested building this in a `catkin` workspace yet. We probably need to sort out what to do about `flatland` and the not yet (not ever?) migrated `beluga` packages in the repo before that can work.

nahueespinosa:
> Do we want CI for ROS Noetic? I'm inclined to say yes, but it will add to CI job durations.

I agree with you, we do want CI for ROS Noetic. IMO, it won't necessarily add to the total duration as we can run jobs in parallel.

> I have not migrated beluga_benchmark nor beluga_system_tests yet.

I wouldn't migrate those just yet. We're leaning towards running benchmarks using LAMBKIN and I think we should find a more ROS-agnostic way of running `beluga_system_tests`.

> Perhaps it's time to consider using [vcstool](https://github.com/dirk-thomas/vcstool) (or [others](http://wiki.ros.org/rosinstall_generator)) to handle dependencies.

+1. Makes sense to me.

nahueespinosa:
I think this PR could use a split. I'm thinking on creating a PR to add a ROS Noetic Dockerfile and CI infrastructure to build and test the `beluga` package only. This one can be a follow-up to include `beluga_amcl` and `beluga_example`.

I'll try to replace `git submodules` with [`vcstool`](https://github.com/dirk-thomas/vcstool) to handle dependencies that need to be installed from source.

hidmic:
Closing this PR in favor of a split submission.

PR #212, opened 2023-06-10
Fix map to odom transform calculation

nahueespinosa:
### Proposed changes

This bug was not caught before because map to odom in our `perfect_odometry` bag is the identity.

#### Type of change

### Checklist


glpuga (approved):
LGTM! Good catch!

PR #215, opened 2023-06-11
Remove simulated odometry noise

nahueespinosa:
### Proposed changes

Flatland odometry noise causes strange behavior that is not replicable in a real life setup. Lowering the noise parameters does not make this more usable. This patch also adds a missing execution dependency.

Current settings:

```
odom_pose_noise: [0.005, 0.005, 0.01]
```

https://github.com/Ekumen-OS/beluga/assets/33042669/9b5ab4e0-e776-41e2-82db-bc696df097a5

Lowered noise parameters:
```
odom_pose_noise: [0.00005, 0.00005, 0.000001]
```
https://github.com/Ekumen-OS/beluga/assets/33042669/8cf343ab-9ed8-4056-a379-6dd6c51289f2

#### Type of change

### Checklist

### DIFF
--- a/beluga_example/models/turtlebot.model.yaml
+++ b/beluga_example/models/turtlebot.model.yaml
@@ -17,7 +17,6 @@ plugins:
     odom_frame_id: odom
     pub_rate: 10
     enable_odom_pub: true
-    odom_pose_noise: [0.005, 0.005, 0.01]
 
   - type: ModelTfPublisher
     name: tf_publisher
--- a/beluga_example/package.xml
+++ b/beluga_example/package.xml
@@ -17,6 +17,7 @@
 
   <exec_depend>beluga_amcl</exec_depend>
   <exec_depend>flatland_server</exec_depend>
+  <exec_depend>flatland_plugins</exec_depend>
   <exec_depend>launch_ros</exec_depend>
   <exec_depend>nav2_amcl</exec_depend>
   <exec_depend>nav2_lifecycle_manager</exec_depend>


hidmic:
Hmm, `flatland`'s diff drive odometry noise is typical [gaussian noise](https://github.com/avidbots/flatland/blob/4e2f6b08159e4fa05a62a859e8229a1945c124c9/flatland_plugins/src/diff_drive.cpp#LL144C5-L144C15). I wouldn't be so quick to disregard malfunction. Can we reproduce with `nav2_amcl`?

nahueespinosa:
Good point, to be clear, `beluga_amcl` can localize when moving, but with this kind of noise it's very easy for particles to drift in any direction over time.

https://github.com/Ekumen-OS/beluga/assets/33042669/6249e062-10af-4a7c-8245-e66c07e9e21c

I tested this in `rolling` with the latest `navigation2` main, `nav2_amcl` doesn't have this problem because it doesn't perform any motion updates when there are no particles.

**Now that I write this out loud, could that be the reason for increased CPU usage? We're always processing scans, even if the robot is standing still.**

This is the odometry signal from flatland with `odom_pose_noise: [0.00005, 0.00005, 0.000001]`:

image

For comparison this is the odometry signal from the carpinchobot:

image

We might need to review our decision of updating the motion and sensor models when the robot is not moving.
We can also try tuning AMCL parameters to handle this type of noise, I think it's possible.

glpuga:
> Now that I write this out loud, could that be the reason for increased CPU usage? We're always processing scans, even if the robot is standing still.

:eyes: 

That definitely fits the bill. I've kept thinking we were somehow _doing more work_ .

I think the key is the definition of "standing still".  Actually that is implemented as "not moving far enough", so it is indeed decimating actual processing when the robot is moving too (the faster it goes, the less it decimates).

For amcl it decimates both sensor and resampling, for us only resampling. So yeah, I think you may be onto something here.

nahueespinosa:
Alright, closing this one in favor of #230.

PR #190, opened 2023-05-16
Use rclcpp::QoS wherever possible

hidmic:
## Summary

Minor fix, to avoid deprecation warnings (and compilation errors if `-Wall` is in force) when building against ROS 2 Rolling.

## Checklist
- [x] Read the [contributing guidelines](https://github.com/Ekumen-OS/beluga/blob/main/CONTRIBUTING.md).
- [x] Configured pre-commit and ran colcon test locally.
- [x] Signed all commits for DCO.
- [ ] Added tests (regression tests for bugs, coverage of new code for features).
- [ ] Updated documentation (as needed).
- [ ] Checked that CI is passing.

PR #216, opened 2023-06-11
Configure vcstool and use docker multi-stage builds

nahueespinosa:
### Proposed changes

Related to #95 and mentioned #210.

This patch removes `git submodules` from the repository in favor of `vcstool` so we can better handle dependencies for different ROS versions. Additionally, it makes use of Docker's multistage builds to cache package manifest files.

#### Type of change

### Checklist

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,0 @@
-[submodule "flatland"]
-  path = flatland
-  url = git@github.com:avidbots/flatland.git
-  branch = ros2-humble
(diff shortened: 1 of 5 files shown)


glpuga (approved):
LGTM

hidmic:
This PR omitted importing .repos files in CI, which didn't break because in Humble, `flatland` is the only dependency and an exec dependency of all types. That is not true for Rolling (which we are not running against in CI).

nahueespinosa:
We were only building Humble code on CI, so I'd say it was out of scope for this PR. I agree that we should add that to a follow-up PR that adds Rolling builds to CI which, by the way, should also fix #190.

PR #217, opened 2023-06-12
Add ROS Noetic CI infrastructure

hidmic:
### Proposed changes

Follow-up PR after #209. This patch adds CI infrastructure to build `beluga` (and only `beluga` for now) against ROS Noetic in Ubuntu Focal.

#### Type of change

### Checklist

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -14,9 +14,9 @@ jobs:
   lint:
     runs-on: ubuntu-22.04
     container:
-      image: ros:rolling-ros-base
+      image: ros:humble-ros-base
     env:
-      ROS_DISTRO: rolling
+      ROS_DISTRO: humble
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
@@ -47,7 +47,6 @@ jobs:
         ros_distro:
           - noetic
           - humble
-          - rolling
     container:
       image: ros:${{ matrix.ros_distro }}-ros-base
     steps:
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -11,12 +11,37 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build-test-lint:
+  lint:
     runs-on: ubuntu-22.04
-    env:
-      ROS_DISTRO: humble
     container:
-      image: ros:humble-ros-base-jammy
+      image: ros:rolling-ros-base
+    env:
+      ROS_DISTRO: rolling
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          path: ${{ github.workspace }}/src/beluga
+
+      - name: Lint
+        working-directory: ${{ github.workspace }}/src/beluga/
+        shell: bash
+        run: |
+          source /opt/ros/${{ env.ROS_DISTRO }}/setup.sh
+          pre-commit run --all-files --verbose --show-diff-on-failure
+
+  build-test:
+    runs-on: ubuntu-22.04
+    needs: lint
+    strategy:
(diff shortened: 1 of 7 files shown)


tools/run-clang-tidy.sh
```diff
@@ -20,7 +20,11 @@
 
 set -o errexit -o xtrace
 
-ROS_PACKAGES="beluga beluga_system_tests"
+if [ "${ROS_DISTRO}" != "noetic" ]; then
+    ROS_PACKAGES="beluga beluga_system_tests"
+else
+    ROS_PACKAGES="beluga"
```
nahueespinosa, line 26: Regarding the CI static analysis failure, we can probably add `-clang-diagnostic-gnu-zero-variadic-macro-arguments` to the checks in `.clang-tidy`.

The include order warning is weird, but adding a newline in between in `test_strongly_typed_numeric.cpp` may fix it:
```cpp
#include <gtest/gtest.h>

#include <beluga/type_traits/strongly_typed_numeric.hpp>
```

nahueespinosa (approved):
LGTM!

.github/workflows/ci_pipeline.yml
```diff
@@ -65,20 +97,13 @@ jobs:
           fail_ci_if_error: true
           verbose: true
```
nahueespinosa, line 98: Should we add a condition to upload documentation artifacts and code coverage reports in only one of the jobs?

Something like:
```yaml
- if: ${{ matrix.ros_distro == 'humble' }}
  name: Upload documentation artifacts
```

tools/run-clang-tidy.sh
nahueespinosa: Hmm, there is also an unused parameter in a googletest macro... I wonder if it is related to the googletest version or the differences between `clang-tidy-10` and `clang-tidy-14`.

.github/workflows/ci_pipeline.yml
nahueespinosa: Or better yet, do something like what is proposed in https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-expanding-configurations.

```yaml
strategy:
      fail-fast: false
      matrix:
        ros_distro:
          - noetic
          - humble
      include:
          - ros_distro: humble
            upload_artifacts: true
```

```yaml
- if: ${{ matrix.upload_artifacts }}
  name: Upload documentation artifacts
  
- if: ${{ matrix.upload_artifacts }}
  name: Upload code coverage report
```

tools/run-clang-tidy.sh
hidmic: I believe it's https://github.com/google/googletest/issues/2650, which has been fixed for googletest 1.11 but not backported to 1.10. I tried and failed to handle googletest versions in CMake -- Kitware's GTest module doesn't detect any versions. Dropping the warning seems like the only path forward.

hidmic: >  adding a newline in-between in test_strongly_typed_numeric.cpp may fix it:

Indeed.

.github/workflows/ci_pipeline.yml
hidmic: OK with uploading documentation artifacts once, but it does make sense to have coverage reports in Noetic as well. Not sure how to configure `codecov` for that though.

tools/run-clang-tidy.sh
hidmic: > we can probably add -clang-diagnostic-gnu-zero-variadic-macro-arguments to the checks in .clang-tidy.

This doesn't seem to work. Passing `-Wno-gnu-zero-variadic-macro-arguments` to the compiler does (and thus to the compiler_commands.json file) does.

nahueespinosa (approved):

PR #231, opened 2023-06-18
Speed up docker builds in CI

nahueespinosa:
### Proposed changes

Build `timem` (and only `timem`) in a separate stage in our Dockerfiles.
Use [GitHub Actions cache backend](https://docs.docker.com/build/cache/backends/) in the Docker CI pipeline.

#### Type of change

### Checklist

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -90,9 +90,9 @@ jobs:
         with:
           api_token: ${{ secrets.GITHUB_TOKEN }}
           update_packager_index: false
-          override_cache_key: ccache-${{ matrix.ros_distro }}-${{ github.head_ref || github.ref }}
+          override_cache_key: ccache-${{ matrix.ros_distro }}-${{ github.head_ref || github.ref_name }}
           override_cache_key_fallback: |
-            ccache-${{ matrix.ros_distro }}-${{ github.base_ref || github.ref }}
+            ccache-${{ matrix.ros_distro }}-${{ github.base_ref || github.ref_name }}
             ccache-${{ matrix.ros_distro }}-
 
       - name: Build and test
--- a/docker/images/humble/Dockerfile
+++ b/docker/images/humble/Dockerfile
@@ -12,6 +12,31 @@ RUN mkdir -p /tmp/ws/src \
   && find ./ -name "COLCON_IGNORE" | xargs cp --parents -t /tmp/ws/src \
   || true
 
+FROM ros:humble-ros-base-jammy AS timem-builder
+
+RUN apt-get update \
+  && apt-get install --no-install-recommends -y \
+    git \
+    python3-pip \
+    python3-venv \
+  && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /opt
+
+RUN git clone -b develop https://github.com/NERSC/timemory.git \
+  && cd timemory \
+  && git checkout 415650ee26f358218908983c87212b620c3a0328 \
+  && python3 -m venv env \
+  && . env/bin/activate \
+  && pip install -r requirements.txt \
+  && cmake -B ./build \
+    -DTIMEMORY_INSTALL_HEADERS=OFF \
+    -DTIMEMORY_INSTALL_CONFIG=OFF \
+    -DTIMEMORY_INSTALL_ALL=OFF \
+    -DTIMEMORY_BUILD_TIMEM=ON \
+    . \
+  && cmake --build ./build --target timem
+
 FROM ros:humble-ros-base-jammy AS builder
 
 ENV DEBIAN_FRONTEND noninteractive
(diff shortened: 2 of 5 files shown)


docker/images/rolling/Dockerfile
```diff
@@ -12,6 +12,26 @@ RUN mkdir -p /tmp/ws/src \
   && find ./ -name "COLCON_IGNORE" | xargs cp --parents -t /tmp/ws/src \
   || true
 
+FROM ros:rolling-ros-base-jammy AS timem-builder
+
+RUN apt-get update \
+  && apt-get install --no-install-recommends -y \
+    git \
```
glpuga, line 19: nit: This line-break can be avoided.

docker/images/noetic/Dockerfile
```diff
@@ -22,6 +22,26 @@ RUN mkdir -p /tmp/ws/src \
   && find ./ -name "COLCON_IGNORE" | xargs cp --parents -t /tmp/ws/src \
   || true
 
+FROM ros:noetic-ros-base-focal AS timem-builder
+
+RUN apt-get update \
+  && apt-get install --no-install-recommends -y \
+    git \
```
glpuga, line 29: nit: same.

docker/images/humble/Dockerfile
```diff
@@ -12,6 +12,26 @@ RUN mkdir -p /tmp/ws/src \
   && find ./ -name "COLCON_IGNORE" | xargs cp --parents -t /tmp/ws/src \
   || true
 
+FROM ros:humble-ros-base-jammy AS timem-builder
+
+RUN apt-get update \
+  && apt-get install --no-install-recommends -y \
+    git \
```
glpuga, line 19: nit: same.

.github/workflows/ci_pipeline.yml
```diff
@@ -90,9 +90,9 @@ jobs:
         with:
           api_token: ${{ secrets.GITHUB_TOKEN }}
           update_packager_index: false
-          override_cache_key: ccache-${{ matrix.ros_distro }}-${{ github.head_ref || github.ref }}
+          override_cache_key: ccache-${{ matrix.ros_distro }}-${{ github.head_ref || github.ref_name }}
           override_cache_key_fallback: |
-            ccache-${{ matrix.ros_distro }}-${{ github.base_ref || github.ref }}
+            ccache-${{ matrix.ros_distro }}-${{ github.base_ref || github.ref_name }}
```
glpuga, line 95: nit: I've been thinking that we don't really benefit from splitting the caches from different branches, since it prevents main from benefiting from the final run from each PR, because the PR cache is lost. Splitting probably makes sense in repositories with lots of traffic, with multiple PRs running CI at the same time which would overwrite each other's cache upon merging.

ccache on its own can keep track of what is common and what is not. The one thing to keep under control is that in reusing the cache, it may never be purged and it will continue to grow, but we can cap the size to something sensible so that it begins evicting items if it grows to large. A full build is about 20-30Mb, so 300 Mb is probably a reasonable cap.

glpuga (approved):
Minor comments, but LGTM.

docker/images/humble/Dockerfile
```diff
@@ -12,6 +12,26 @@ RUN mkdir -p /tmp/ws/src \
...
+  && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /opt
+
+RUN git clone -b develop https://github.com/NERSC/timemory.git \
```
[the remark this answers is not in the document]
nahueespinosa: I prefer to avoid that and at this point the image only has one root user.

docker/images/rolling/Dockerfile
nahueespinosa: I'll keep it that way because there's a lot of copying and pasting from Dockerfiles and this format makes it very clear which packages are being installed.

docker/images/noetic/Dockerfile
nahueespinosa: I'll keep it that way because there's a lot of copying and pasting from Dockerfiles and this format makes it very clear which packages are being installed.

docker/images/humble/Dockerfile
nahueespinosa: I'll keep it that way because there's a lot of copying and pasting from Dockerfiles and this format makes it very clear which packages are being installed.

.github/workflows/ci_pipeline.yml
nahueespinosa: Sounds like we could do that. The current behavior is not that bad though:
- A new PR can use main's cache, then it creates its own cache file and can re-use that after a new push.
- When the PR is merged, the first option is to use main's cache, which sounds sane IMO.

> The one thing to keep under control is that in reusing the cache, it may never be purged and it will continue to grow, but we can cap the size to something sensible so that it begins evicting items if it grows to large.

I'd have to see how this is implemented, but it seems like extra steps for only a small gain.

I think the main benefit of caching compiler output is to speed up builds before merging. The CI run after a push to main can usually wait without affecting any users.

Keeping cache sizes small is also good for upload/download times, although the difference between 30MB and 300MB can be marginal.

nahueespinosa: Resolving the conversation to merge this PR, but we can follow-up on this and create a ticket if necessary.

PR #236, opened 2023-06-25
Remove uncrustify and cpplint from pre-commit hooks

nahueespinosa:
### Proposed changes

Connected to #132.

This patch removes uncrustify, cppcheck, and cpplint from the beluga_amcl package and replaces them with clang-format and clang-tidy. It also formats the code accordingly.

#### Type of change

- [ ] 🐛 Bugfix (change which fixes an issue)
- [x] 🚀 Feature (change which adds functionality)
- [ ] 📚 Documentation (change which fixes or extends documentation)

### Checklist

- [x] Lint and unit tests (if any) pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] All commits have been signed for [DCO](https://developercertificate.org/)

### Additional comments

The plan is to rebase and merge this after #233 lands.

Issue #132, opened 2023-03-07
Modify pre-commit config to not require ROS 2


ivanpauno:
## Description

So you can easily run linters without the need of being inside the docker container a ROS 2 environment.

## Definition of done

- [x] Replace `ament_clang_format` with [clang format](https://github.com/ssciwr/clang-format-wheel#use-from-pre-commit) official hook.
- [x] Replace `ament_clang_tidy` with [clang tidy](https://github.com/pocc/pre-commit-hooks#example-usage).
- [x] Replace `ament_uncrustify` with `clang-format`, format `beluga_amcl` code as needed.
- [x] Remove `ament_cpplint`.
- [x] Replace `ament_flake8` with [flake8](https://github.com/PyCQA/flake8).
  We will need a copy of https://github.com/ament/ament_lint/blob/rolling/ament_flake8/ament_flake8/configuration/ament_flake8.ini if we want to the ROS 2 style exactly (do we want?).
- [x] Replace `pep257` with `flake8-docstrings` (plugin of above hook, also enforces `pep257`)
- [ ] Replace `ament_copyright` with something like https://github.com/leoll2/copyright_notice_precommit.
- [x] Make `clang-tidy` only warn if not in the container (probably the best thing to check is if it's possible to generate the compile commands database), see https://github.com/ekumenlabs/beluga/pull/135#issuecomment-1470806419.

## Additional considerations

Low priority, but I think it would be nice to not need a ros 2 environment to run the pre-commit hooks.

PR #237, opened 2023-06-25
Remove ament_flake8 from pre-commit hooks

nahueespinosa:
### Proposed changes

Connected to #132 and builds on top of #236.

#### Type of change

### Checklist

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,9 @@
 [flake8]
-extend-ignore = B902,C816,D100,D101,D102,D103,D104,D105,D106,D107,D203,D212,D404,I202
+# D100: Missing docstring in public module.
+# D101: Missing docstring in public class.
+# D103: Missing docstring in public function.
+# D104: Missing docstring in public package.
+extend-ignore = D100,D101,D103,D104
 import-order-style = google
 max-line-length = 99
 show-source = true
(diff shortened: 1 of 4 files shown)


.flake8
```diff
@@ -0,0 +1,6 @@
+[flake8]
+extend-ignore = B902,C816,D100,D101,D102,D103,D104,D105,D106,D107,D203,D212,D404,I202
```
hidmic, line 2: @nahueespinosa I'm curious about B902. It does seem to indicate a bug.

hidmic (approved):

.flake8
nahueespinosa: This was the configuration used by `ament_flake8`.
https://github.com/ament/ament_lint/blob/rolling/ament_flake8/ament_flake8/configuration/ament_flake8.ini

I kept only the ones that currently fail in https://github.com/Ekumen-OS/beluga/pull/237/commits/01552b9b253d9b1c4b6aa3025f46b4da105b8f55.

ivanpauno: It's the [ament_flake8](https://github.com/ament/ament_lint/blob/f10eecd28cdc05746c1f8c0f45984db6052cfd40/ament_flake8/ament_flake8/configuration/ament_flake8.ini#L2) style.

IMO, we could use standard flake8 without any ignore.
(some may be required to be compatible with black)

PR #238, opened 2023-07-01
Fix `perfect_odometry` rosbag

### LINKED ISSUE #232: Timeline mixup when running bagfile
### Bug description

A few transforms and messages are getting stamped with unmatched timestamps, probably as a result of a mixup when the bagfile was generated.

As a result of this, amcl is publishing the map->odom tf in a totally different timescale than the odom->base and base->laser_front , causing RViz to fail to display the later.

See timestamps on the logs and on the two transforms on the right:

Screenshot from 2023-06-19 13-59-14

See timestamps on the /clock and /scan topics:

Screenshot from 2023-06-19 14-01-11

On RViz,

Screenshot from 2023-06-19 13-52-31

**Platform (please complete the following information):**
 - OS: Humble development container included in Beluga
 - `Beluga` version: ed7dfe38bcab566a900b636378e7dbf05981120a

### How to reproduce
_List steps to reproduce the issue:_
1. On a console: `cd /ws && colcon build && source install/setup.bash && ros2 launch beluga_example perfect_odometry.launch.xml`
2. On another console: `ros2 topic echo /tf`, `ros2 topic echo /clock`, `ros2 topic echo /scan`
3. On RViz, the TF visor will also show a warning for both base and laser_front.

_Code snippets or minimal examples are always helpful, if not necessary._

**Expected behavior**

All message timestamps to be consistent when running out of a demo and test bagfile.

**Actual behavior**

Timestamps are not consistent with a single timeline, and some information fails to be displayed as a result.

nahueespinosa:
### Proposed changes

Fixes #232.

This patch fixes message timestamps and starting time in the rosbag to match the header stamps recorded with sim time.

* The contents of the messages were not altered.
* <del>The storage backend was changed to mcap.</del>
* The last messages where there was a collision against one of the walls were deleted.

#### Type of change

### Checklist

### DIFF
--- a/beluga_example/bags/perfect_odometry/metadata.yaml
+++ b/beluga_example/bags/perfect_odometry/metadata.yaml
@@ -1,39 +1,49 @@
 rosbag2_bagfile_information:
-  version: 5
-  storage_identifier: "sqlite3"
+  version: 7
+  storage_identifier: mcap
   duration:
-    nanoseconds: 129982460028
+    nanoseconds: 135344191329
   starting_time:
-    nanoseconds_since_epoch: 1676668727259050014
-  message_count: 4595
+    nanoseconds_since_epoch: 36399992720
+  message_count: 4790
   topics_with_message_count:
-    - topic_metadata:
-        name: /odometry/filtered
-        type: nav_msgs/msg/Odometry
-        serialization_format: cdr
-        offered_qos_profiles: "- history: 3\n  depth: 0\n  reliability: 1\n  durability: 2\n  deadline:\n    sec: 9223372036\n    nsec: 854775807\n  lifespan:\n    sec: 9223372036\n    nsec: 854775807\n  liveliness: 1\n  liveliness_lease_duration:\n    sec: 9223372036\n    nsec: 854775807\n  avoid_ros_namespace_conventions: false"
-      message_count: 919
-    - topic_metadata:
-        name: /odometry/ground_truth
-        type: nav_msgs/msg/Odometry
-        serialization_format: cdr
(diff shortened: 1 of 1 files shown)


glpuga (changes requested):
The topics LGTM, but when I tried to run the perfect odometry example rosbag does not start because the MCAP plugin is not installed.

After installing manually ros-humble-rosbag2-storage-mcap it worked just fine. We are probably missing a dependency declaration. 

I had started the docker with `./run --build` .

nahueespinosa:
@glpuga Good catch! I originally tested on Rolling.

I reverted to using sqlite3 instead of mcap so no need to add extra dependencies.

glpuga:
I looked into the results of the `beluga_benchmark` that run on top of this. The data gathered looks good and can be plotted just fine for `beluga_amcl`, but `nav2_amcl` seems to be crashing right after starting. 

This can be duplicated by running:

```
colcon build 
source install/setup.bash
cd src/beluga/beluga_benchmark/docs/reports/2023-06-03
mkdir test
cd test
ros2 run beluga_benchmark parameterized_run 250 300 400 500 750 1000 2000 5000 10000 20000 50000 100000 200000 --params-file ../beam_params.yaml --package nav2_amcl --executable amcl
cd -
``` 

Nothing is left on the logs to indicate why `nav2_amcl` is giving up.

nahueespinosa:
@glpuga I can reproduce the error in Humble, but everything works fine in Rolling. More investigation is needed.

nahueespinosa:
@glpuga I think the failure is most likely related to what @ivanpauno found here: https://github.com/Ekumen-OS/beluga/pull/117#issuecomment-1442370024

I don't know how we ended up re-enabling `recovery_alpha_fast` and `recovery_alpha_slow` for benchmarks, and how that worked in the past, but we have two alternatives:
- Benchmark using Rolling where the `nav2_amcl` bug was fixed.
- Benchmark with adaptive recovery disabled.

Any of those options is unrelated to this PR.

glpuga (approved):

PR #240, opened 2023-07-05
Replace ament_copyright with local script

### LINKED ISSUE #132: Modify pre-commit config to not require ROS 2

nahueespinosa:
### Proposed changes

Closes #132.

I couldn't find a pre-commit hook that I liked, so I wrote one... It isn't half bad IMO.

#### Type of change

### Checklist

### Additional comments

The options I considered were:
- https://github.com/leoll2/copyright_notice_precommit
- https://github.com/sbrunner/hooks

### DIFF
--- a/tools/check_copyright.py
+++ b/tools/check_copyright.py
@@ -1,0 +1,190 @@
+#!/usr/bin/env python3
+
+# Copyright 2023 Ekumen, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Check that files have a copyright notice.
+
+This tool checks that a copyright notice matching a specific template exists in the
+provided files. It can also fix the file by adding a copyright notice or replacing
+an existing one.
+"""
+
+import argparse
+import datetime
+import re
+import textwrap
+
+from pathlib import Path
+
+DEFAULT_COPYRIGHT_HOLDER = 'Ekumen, Inc.'
+
+# Template for the copyright notice.
+# This should not contain any regex patterns since special characters will be escaped.
+DEFAULT_COPYRIGHT_TEMPLATE = '''\
+Copyright {year} {name}
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


tools/check_copyright.py
```diff
@@ -0,0 +1,190 @@
...
+limitations under the License.\
+'''
+
+
+class FileContentChecker:
+    """Helper class to check the contents of a file and apply fixes."""
+
+    # Regex pattern that matches a year (YYYY) and a year range (YYYY-ZZZZ).
+    # It has capturing groups for both years (the second one is optional).
+    _YEAR_PATTERN = r'([0-9]{4})(?:(?:-)([0-9]{4}))?'
```
ivanpauno, line 57: why is the `-` optional?, shouldn't it be:

```suggestion
    _YEAR_PATTERN = r'([0-9]{4})(?:-([0-9]{4}))?'
```

I may have got confused ...

tools/check_copyright.py
```diff
@@ -0,0 +1,190 @@
...
+        """Verify that the file has a valid copyright notice.
+
+        :param template: Template copyright notice.
+        :return: True if a copyright notice was found. False otherwise.
+        """
+        pattern = (
+            # Comment lines and escape special characters in the template...
+            re.escape(self._comment_lines(template.format(year='YYYY', name='NAME')))
+            # Then insert year and name patterns.
+            .replace('YYYY', self._YEAR_PATTERN).replace('NAME', self._NAME_PATTERN)
```
ivanpauno, line 100: why combining format and replace and not directly formatting with the patterns?

ivanpauno (commented):
I have a couple questions, lgtm

tools/check_copyright.py
nahueespinosa: Good catch! It isn't optional actually, but it is an unnecessary non-capturing group.

Fixed in https://github.com/Ekumen-OS/beluga/pull/240/commits/2ba7d7af002c3a3c61bcabf9a227cee1d31529ac.

tools/check_copyright.py
ivanpauno: I see, thanks for clarifying. I got lost before in where each function call ended.
https://github.com/Ekumen-OS/beluga/commit/ab72795b019ed973811a2086edf16b8001e956a0 LGTM!

ivanpauno (approved):
LGTM

Issue #230, opened 2023-06-17
Particles drift over time when odometry has Gaussian noise in simulation


nahueespinosa:
### Bug description
When we configure the diff drive flatland plugin to add non-zero gaussian noise to the published odometry, and the robot stands still, we observed that particles drift over time severely affecting the estimate.

**Platform (please complete the following information):**
 - OS: Ubuntu Jammy
 - `Beluga` version: https://github.com/Ekumen-OS/beluga/commit/ed7dfe38bcab566a900b636378e7dbf05981120a

### How to reproduce
_List steps to reproduce the issue:_
1. Launch Beluga example with a simulation.
2. Move around and stop the robot.

**Expected behavior**
The estimate doesn't diverge when the robot is not moving.

**Actual behavior**
Noise parameters:
```
odom_pose_noise: [0.00005, 0.00005, 0.000001]
```

https://github.com/Ekumen-OS/beluga/assets/33042669/92643909-8802-488e-85ce-0fa56e0d9b33

### Additional context

See #215 for more context.

nahueespinosa:
Fixed by #233. The noise parameters in the simulation are still too high for both `beluga_amcl` and `nav2_amcl` to localize effectively.

PR #241, opened 2023-07-05
Reduce simulated odometry noise

nahueespinosa:
### Proposed changes

#230 is fixed, but the simulated odometry noise is still too high for the simulation to be usable.

#### Type of change

### Checklist

### DIFF
--- a/beluga_example/models/turtlebot.model.yaml
+++ b/beluga_example/models/turtlebot.model.yaml
@@ -17,7 +17,7 @@ plugins:
     odom_frame_id: odom
     pub_rate: 10
     enable_odom_pub: true
-    odom_pose_noise: [0.005, 0.005, 0.01]
+    odom_pose_noise: [0.0002, 0.0002, 0.0001]
 
   - type: ModelTfPublisher
     name: tf_publisher


glpuga (approved):

beluga_example/models/turtlebot.model.yaml
```diff
@@ -17,7 +17,7 @@ plugins:
     odom_frame_id: odom
     pub_rate: 10
     enable_odom_pub: true
-    odom_pose_noise: [0.005, 0.005, 0.01]
+    odom_pose_noise: [0.0002, 0.0002, 0.0001]
```
hidmic, line 20: @nahueespinosa meta: I'm a bit concerned about the order of magnitude. It's one thing for this to be an artifact of the simulation (ie. unrealistic noise PSD), it's a different thing if Beluga needs very precise odometry to function. Not to be addressed here though.

hidmic (approved):

beluga_example/models/turtlebot.model.yaml
nahueespinosa: > It's a different thing if Beluga needs very precise odometry to function.

I don't think that's the case. FWIW nav2_amcl has the same problem with this set of parameters.

PR #251, opened 2023-07-29
Refactor grid to measure cache-awareness impact on grid tracing

glpuga:
### Proposed changes

- Refactors the grid to avoid low-level access using indexes that assume a memory layout.
- Extracts the grid storage into it's own thing, the LinearGridStorage class.
- Creates multiple cache-friendly storage alternatives.
- Created a number of benchmarks to evaluate them.
- Minor fixes and changes.
- Also adds an alias to the docker to run the pre-commits from any folder.

Note: For review of the changes it's probably better to do a first walk-around commit by commit before looking at the changes as a whole.

#### Type of change

### Checklist

### Additional comments

The summary after the usefulness of cache friendly when raytracing is that it does a lot of difference, but only for long rays (over 100m). For shorter rays, the ratio between the linear storage and the best cache-aware storage inverts.

These are the measurements taken by:
- tracing from a central point
- 182 different bearings from that central point equally dividing the 360 degrees, sequentially visited in clock-wise sequence (similar to actual processing of scan data).
- Obstacles at 20, 50, 100 or 200 meters.

Screenshot from 2023-07-29 14-29-05


hidmic:
How do you suggest we go about this @glpuga? Are cache friendly storage layouts pluggable like other features in the library?

glpuga:
> How do you suggest we go about this @glpuga?

I think the changes are worth merging, if only because the relation between the likelihood sensor model and the grid is a bit more clearcut and low-level accesses to the grid were removed.

Performance-wise, there's no change. Benchmarks seem to indicate that you'd only see meaningful differences with maps much larger than the one we currently use to measure performance.

> Are cache friendly storage layouts pluggable like other features in the library?

They are. To switch between them it's possible to change the storage type parameter of the occupany grid storage mixin.

https://github.com/Ekumen-OS/beluga/pull/251/files#diff-913a8969931817e56d8350462bd5ebba435ef2fde37ec83ee653ba483c05dc4bR51

beluga/include/beluga/sensor/data/cache_friendly_grid_storage_4.hpp
```diff
@@ -0,0 +1,164 @@
...
+    return mapping_function(static_cast<std::size_t>(x), static_cast<std::size_t>(y));
+  }
+
+  static constexpr std::size_t kL1TileCols{2};
+  static constexpr std::size_t kL1TileRows{2};
+
+  static constexpr std::size_t kL2TileSize{kL1TileRows * kL1TileCols * calculate_tile_area(LineLenght / sizeof(T))};
+  static constexpr std::size_t kL2TileSide{kL1TileCols * calculate_tile_side(LineLenght / sizeof(T))};
+  static constexpr std::size_t kL1TileSize{calculate_tile_area(LineLenght / sizeof(T))};
+  static constexpr std::size_t kL1TileSide{calculate_tile_side(LineLenght / sizeof(T))};
```
hidmic, line 152: @glpuga nit: would this be equivalent to:
```suggestion
  static constexpr std::size_t kL1TileCols{2};
  static constexpr std::size_t kL1TileRows{2};
  static constexpr std::size_t kL1TileSize{calculate_tile_area(LineLenght / sizeof(T))};
  static constexpr std::size_t kL1TileSide{calculate_tile_side(LineLenght / sizeof(T))};
  
  static constexpr std::size_t kL2TileSize{kL1TileRows * kL1TileCols * kL1TileSize};
  static constexpr std::size_t kL2TileSide{kL1TileCols * kL1TileSide};
```
?

beluga/include/beluga/sensor/data/cache_friendly_grid_storage_4.hpp
```diff
@@ -0,0 +1,164 @@
...
+
+#include <cstdint>
+#include <iostream>
+#include <memory>
+
+namespace beluga {
+
+/// @brief A cache friendly grid storage that efficiently packs grid patches in cache lines. This algorithm maps cache
+/// lines to square tiles, and also groups tiles in 2x2 groups. The tiles are transposed in a checkerboard fashion to
+/// try to exploit cache prefecth algorithms in the cpu.
```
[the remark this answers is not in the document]
hidmic: Also, an explanation of why this layout optimizes cache prefetching would be useful.

beluga/include/beluga/sensor/data/dense_grid2_mixin.hpp
```diff
@@ -71,16 +57,16 @@ namespace beluga {
...
- * \tparam Derived Concrete dense grid type. It must define
- * `Derived::width()`, `Derived::height()`, `Derived::resolution()`,
- * `Derived::data_at(index)`, and `Derived::index_at(int, int)`
- * as described in \ref DenseGrid2Page.
+ * \tparam Mixin Mixin type
  */
-template <typename Derived>
-class BaseDenseGrid2 : public BaseRegularGrid2<Derived> {
+template <typename Mixin>
+class DenseGrid2Mixin : public Mixin {
```
hidmic, line 63: @glpuga nit^2: do we need to (over)specify it's a mixin by tucking `mixin` at the end of the class name? Same elsewhere.

beluga/include/beluga/sensor/data/occupancy_storage_mixin.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+/// Occupancy grid storage interface layer.
+/**
+ * \tparam Mixin Mixin type
+ * \tparam ValueType Occupancy grid cell value type
+ * \tparam ValueTraitsType Occupancy grid cell value traits type
+ * \tparam MapStorageType Occupancy grid storage type
+ */
+template <class Mixin, typename ValueType, typename ValueTraitsType, typename MapStorageType>
+class OccupancyStorageMixin : public Mixin {
```
hidmic, line 38: @glpuga meta: what makes it an occupancy grid? I think we are lacking type contracts in documentation for value related types.

beluga/include/beluga/sensor/data/value_grid2_mixin.hpp
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * \param data Grid data.
+   * \param width Grid width. Must evenly divide `data` size.
+   * \param resolution Grid resolution.
+   * \param args Arguments to be forwarded to the mixin components.
+   */
+  template <typename... Args>
+  explicit ValueGrid2Mixin(std::vector<T> data, std::size_t width, double resolution, Args&&... args)
+      : Mixin(std::forward<Args>(args)...),
+        data_(std::move(data)),
```
hidmic, line 55: @glpuga meta: shouldn't this grid take a storage as well?

beluga/test/beluga/algorithm/test_distance_map.cpp
```diff
@@ -12,62 +12,62 @@
...
 
+#include <Eigen/Core>
+
 namespace {
 
-auto array_distance(std::size_t first, std::size_t second) {
-  return std::abs(static_cast<intmax_t>(first) - static_cast<intmax_t>(second));
+using CellType = Eigen::Vector2i;
+
+auto array_distance(const CellType& first, const CellType& second) {
```
hidmic, line 27: @glpuga nit: this is no longer an "array" distance but a regular distance in euclidean space.

beluga/include/beluga/sensor/data/occupancy_grid2_mixin.hpp
```diff
@@ -153,31 +130,59 @@ class BaseOccupancyGrid2 : public BaseLinearGrid2<Derived> {
...
+  /// delivers a range of tuples with cell and cell value
+  [[nodiscard]] auto row_major_data() const {
+    const auto width = this->self().width();
+    const auto height = this->self().height();
+    const auto cell_count = width * height;
+    const auto to_raster_data = [this, width](std::size_t raster_index) {
+      const auto xi = static_cast<int>(raster_index % width);
+      const auto yi = static_cast<int>(raster_index / width);
+      const auto cell = Eigen::Vector2i(xi, yi);
+      return std::make_tuple(cell, this->self().data_at(cell).value());
```
hidmic, line 183: @glpuga meta: this assumes the underlying grid is at least dense (by rasterizing w/o checking for data availability). FWIW one of the reasons why at the time I refrained from using mixins for the grid hierarchy are these dependencies. They aren't really independent (although I like that we split dense and regular, that was unnecessary).

beluga/test/beluga/sensor/data/test_cache_friendly_grid_storage.cpp
```diff
@@ -0,0 +1,127 @@
...
+      6,  6,  6,  6,  7,  7,  7,  7,  8,  8,  8,  8,  9,  9,  9,  9,  10, 10, 10, 10, 11, 11,  //
+      6,  6,  6,  6,  7,  7,  7,  7,  8,  8,  8,  8,  9,  9,  9,  9,  10, 10, 10, 10, 11, 11,  //
+      12, 12, 12, 12, 13, 13, 13, 13, 14, 14, 14, 14, 15, 15, 15, 15, 16, 16, 16, 16, 17, 17,  //
+      12, 12, 12, 12, 13, 13, 13, 13, 14, 14, 14, 14, 15, 15, 15, 15, 16, 16, 16, 16, 17, 17,  //
+      12, 12, 12, 12, 13, 13, 13, 13, 14, 14, 14, 14, 15, 15, 15, 15, 16, 16, 16, 16, 17, 17,  //
+      12, 12, 12, 12, 13, 13, 13, 13, 14, 14, 14, 14, 15, 15, 15, 15, 16, 16, 16, 16, 17, 17,  //
+      18, 18, 18, 18, 19, 19, 19, 19, 20, 20, 20, 20, 21, 21, 21, 21, 22, 22, 22, 22, 23, 23,  //
+      18, 18, 18, 18, 19, 19, 19, 19, 20, 20, 20, 20, 21, 21, 21, 21, 22, 22, 22, 22, 23, 23,  //
+  });
+}
```
hidmic, line 80: @glpuga hmm I think we left the assertions out :smiley:

beluga_amcl/src/amcl_node_utils.cpp
```diff
@@ -50,4 +56,23 @@ std::vector<std::pair<double, double>> make_points_from_laser_scan(
...
+    return Sophus::SE2d{rotation, translation};
+  };
+
+  ROSOccupancyGrid::MapStorage grid_storage(ros_msg_ptr->info.width, ros_msg_ptr->info.height);
+  for (std::size_t x = 0; x < ros_msg_ptr->info.width; ++x) {
+    for (std::size_t y = 0; y < ros_msg_ptr->info.height; ++y) {
+      const auto index = x + y * ros_msg_ptr->info.width;
+      grid_storage.cell(static_cast<int>(x), static_cast<int>(y)) = ros_msg_ptr->data[index];
+    }
+  }
```
hidmic, line 72: @glpuga meta: one thing I don't love about this is that now we unconditionally copy the entire grid. That's OK if grids never update and performance improves. It's not that cool for large grids that change frequently or if there is no cache to leverage and memory is scarce (thinking of microcontrollers). Having a storage that maps an existing memory layout to our grids APIs would be nice. FWIW that's another reason why the grid hierarchy is mostly behavior with no state.

hidmic (commented):
Damn, that was a long PR :sweat_smile: Awesome work @glpuga!

One thought: we are focusing on the effects of data fetching on CPU caches. I wonder what is going on with code fetching. I would expect near zero vtable lookups helping us there.

glpuga:
Thanks for the review @hidmic I'll try to address your comments soon.

beluga/include/beluga/sensor/data/cache_friendly_grid_storage.hpp
```diff
@@ -0,0 +1,167 @@
...
+#include <cstdint>
+#include <iostream>
+#include <memory>
+
+namespace beluga {
+
+/// @brief A cache friendly grid storage that efficiently packs grid patches in cache lines. This algorithm maps cache
+/// lines to square tiles on the grid.
+/// @tparam T Type of the data to be stored.
+/// @tparam LineLenght Length of the minimum cache line to optimize for, in bytes.
```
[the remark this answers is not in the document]
glpuga: Fixed.

beluga/include/beluga/sensor/data/cache_friendly_grid_storage.hpp
```diff
@@ -0,0 +1,167 @@
...
+      return elements_per_line;
+    }
+    // the elements in a line can't be mapped to a square tile (2, 8, 32, etc), so
+    // we use two consecutive lines for each tile
+    return elements_per_line * 2;
+  }
+
+  [[nodiscard]] static constexpr std::size_t calculate_tile_side(std::size_t elements_per_line) {
+    auto tile_area = calculate_tile_area(elements_per_line);
+    // we calculate the side length by calculating the square root base four,
```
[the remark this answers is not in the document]
glpuga: Indeed. Fixed.

beluga/include/beluga/sensor/data/cache_friendly_grid_storage.hpp
```diff
@@ -0,0 +1,167 @@
...
+  [[nodiscard]] std::size_t mapping_function(std::size_t x, std::size_t y) const {
+    if constexpr (!is_a_square_power_of_two(LineLenght / sizeof(T))) {
+      // full formula for rectangular tiles
+      const auto tile_x = x / kTileSide;
+      const auto tile_y = y / kTileSide;
+      auto cell_x = x % kTileSide;
+      auto cell_y = y % kTileSide;
+
+      // if we are using two lines per tile, we need to swap the coordinates
+      // in alternating tiles, to avoid large scale anisotropy
```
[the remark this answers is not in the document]
glpuga: When the combination of cache size and cell type is such that there's no way to arrange a single cache line's worth of cells into a square, we group two lines create a tile. If you could see the arrangement of lines in the buffer you'd see this.

```
== == == ==
== == == ==
== == == ==
```
For each "tile" one line holds the upper half, and the other one the lower half. This means that there's a difference in speed when traversing the tile horizontally as opposed to vertically (double the amount of lines are crossed vertically), so there's still a preferred direction.

By transposing alternating cells, the arrangement changes to this:
```
== || == ||
|| == || ==
== || == ||
```

I updated the comment to try to make that more clear.

beluga/include/beluga/sensor/data/cache_friendly_grid_storage.hpp
```diff
@@ -0,0 +1,167 @@
...
+        std::swap(cell_x, cell_y);
+      }
+
+      const auto tile_index = (tile_y * tile_cols_ + tile_x) * kTileSize;
+      const auto cell_index = cell_y * kTileSide + cell_x;
+      return tile_index + cell_index;
+
+    } else {
+      // fast-track formula for square tiles (the performance difference with the one above is
+      // about 10% when used for square tiles)
```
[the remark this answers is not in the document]
glpuga: I tried to improve the comment a bit in this regard.

beluga/include/beluga/sensor/data/cache_friendly_grid_storage.hpp
```diff
@@ -0,0 +1,167 @@
...
+      tile_side *= 2;
+    }
+    return tile_side;
+  }
+
+  [[nodiscard]] static constexpr std::size_t ceil_div(std::size_t numerator, std::size_t denominator) {
+    return ((numerator + denominator - 1) / denominator);
+  }
+
+  [[nodiscard]] std::size_t mapping_function(std::size_t x, std::size_t y) const {
```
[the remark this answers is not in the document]
glpuga: Got it.

beluga/include/beluga/sensor/data/cache_friendly_grid_storage_2.hpp
```diff
@@ -0,0 +1,123 @@
...
+#define BELUGA_SENSOR_DATA_CACHE_FRIENDLY_GRID_STORAGE_2_HPP
+
+#include <cstdint>
+#include <iostream>
+#include <memory>
+
+namespace beluga {
+
+/// @brief A cache friendly grid storage that efficiently packs grid patches in cache lines. This algorithm maps divides
+/// the grid into linesizexlinesize tiles tiles and alternates the orientation of half of the tiles in a checkerboard
```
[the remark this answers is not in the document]
glpuga: Fixed.

beluga/include/beluga/sensor/data/cache_friendly_grid_storage_3.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+  std::vector<WrappedT> storage_;
+
+  const std::size_t grid_width_;
+  const std::size_t grid_height_;
+
+  [[nodiscard]] static constexpr std::size_t next_even(std::size_t n) { return (n % 2 == 0) ? n : n + 1; }
+
+  [[nodiscard]] auto& cell_impl(std::size_t x, std::size_t y) {
+    const auto transpose = (x + y) & 0x01;
```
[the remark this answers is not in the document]
glpuga: Once the compiler gets to it, it's the same. I'll add a comment.

beluga/include/beluga/sensor/data/cache_friendly_grid_storage_3.hpp
```diff
@@ -0,0 +1,110 @@
...
+  /// @brief Returns the width of the map (number of cells).
+  [[nodiscard]] auto width() const { return grid_width_; }
+
+  /// @brief Returns the height of the map (number of cells).
+  [[nodiscard]] auto height() const { return grid_height_; }
+
+ private:
+  struct WrappedT {
+    T value{};
+  };
```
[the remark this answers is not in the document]
glpuga: Fixed.

beluga/include/beluga/sensor/data/cache_friendly_grid_storage_3.hpp
```diff
@@ -0,0 +1,110 @@
...
+  using cell_type = T;
+
+  /// @brief Constructs a map with the given initial values.
+  /// @param width Width of the grid.
+  /// @param height Height of the grid.
+  /// @param init_values Initial contents of the map, in row-major order.
+  CacheFriendlyGridStorage3(std::size_t width, std::size_t height, std::initializer_list<T> init_values = {})
+      : storage_width_(next_even(std::max(width, height))),
+        storage_height_(std::max(width, height)),
+        storage_(next_even(std::max(width, height)) * std::max(width, height), WrappedT{}),
```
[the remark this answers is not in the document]
glpuga: > @glpuga nit: you may use the members initialized above:

Updated.

> Also, no cache alignment on allocation here?

That's be cause this particular method does not depend on the buffer origin to have any particular alignment to work. The mistery here for me is, why did I force the buffer width to be even-numbered?

beluga/include/beluga/sensor/data/cache_friendly_grid_storage_4.hpp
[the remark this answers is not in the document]
glpuga: This came from an informal empirical observation, and some apparent confirmation by some cache prefetching algorithms enumerations around the net, that the CPU seemed to be prefetching consecutive lines (the one after or the one before) even before we stepped on them, which only improves performance for mainly horizontal directions.

This algorithm thus scales the tiling one level more, by mapping _tiles_ in a 2x2 macro-tile arrangement, and transposing half of them to cheat the prefetching algorithm into prefetching horizontally or vertically 50% of the time each.This would tend to equalize times for horizontal and vertical directions.

beluga/include/beluga/sensor/data/cache_friendly_grid_storage_4.hpp
```diff
@@ -0,0 +1,164 @@
...
+ public:
+  /// @brief Type of the data stored in the grid.
+  using cell_type = T;
+
+  /// @brief Constructs a map with the given initial values.
+  /// @param width Width of the grid.
+  /// @param height Height of the grid.
+  /// @param init_values Initial contents of the map, in row-major order.
+  CacheFriendlyGridStorage4(std::size_t width, std::size_t height, std::initializer_list<T> init_values = {})
+      : kL2Tilecols_(ceil_div(width, kL2TileSide)),
```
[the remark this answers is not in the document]
glpuga: There was a bunch of renaming in this file due to linting issues, I think these were caught in it. Fixed.

beluga/include/beluga/sensor/data/cache_friendly_grid_storage_4.hpp
glpuga: Yes, that's ok.

beluga/test/beluga/algorithm/test_distance_map.cpp
glpuga: Fixed.

beluga/test/beluga/sensor/data/test_cache_friendly_grid_storage.cpp
glpuga: Fixed, this an some other initialization weirdness that appeared because default value initialization of the grids got lost along he way when leaving std::vector to use aligned std::unique_ptr to hold buffers.

beluga_amcl/src/amcl_node_utils.cpp
glpuga: I see the point, and nothing prevents us from creating a storage that just maps accesses to an existing buffer like we did before if needed, except lack of generality in the code above to leave initialization out of the grid storage scope to isolate it from ROS types.

I'm not too concerned about that kind of systems constrains (low memory + lack of cache) in the `beluga_amcl` code above too much, though; any system that would finds itself under those constraints is unlikely to run ROS in the first place.

hidmic: > any system that would finds itself under those constraints is unlikely to run ROS in the first place.

Definitely. As long as we don't propagate the constraint to the core library, I'm good :)

hidmic (commented):
Looking great! CI seems somewhat unhappy though.

hidmic:
@glpuga why close this? It was a good addition.

glpuga:
> @glpuga why close this? It was a good addition.

Maybe, but it was not adding new functional features, and it's been stalled for months. Almost every file in the MR was in conflict with the current main due to files moving to new locations and other changes, so the amount of refactor needed just to bring this back is considerable.

I don't have the time to revive this right now, and having an 8 months old PR pinned in the repo is not really good optics so I'd rather close it.

Issue #253, opened 2023-09-02
The nav2_amcl version present in the humble container crashes replaying the perfect odometry bagfile


glpuga:
### Bug description

Running the perfect odometry bagfile against nav2 amcl in the Humble development container causes amcl to crash early in the bagfile.

This is an issue for us because this prevents performance measurements from being captured for this node, and therefore lacking any significant comparison baseline for `beluga`.

Building `nav2_amcl` from source with the source taken from a the current HEAD of navigation2's repository runs just fine, so it's probably an issue within amcl itself, but a relevant change has not been identified as the fix.

**Platform (please complete the following information):**
 - OS: Humble development container.
 - `Beluga` version: c0e2de1

### How to reproduce

1. Go to the beluga working copy
2. `./docker/run --build`
3. Once in the container: `colcon build && . install/setup.bash && ros2 launch beluga_example perfect_odometry.launch.xml localization_package:=nav2_amcl localization_node:=amcl`

**Expected behavior**

AMCL should be able to produce localization data from the bagfile replay.

**Actual behavior**

AMCL crashes. When capturing performance data, no data is captured at all.

### Additional context

This issue has been present since early June or so, but it was initially thought to be caused by #232 and other problems in the bagfiles.

The stack dump of the crash can be obtained with:

```
sudo apt update && sudo apt install xterm -y
colcon build
. install/setup.bash
ros2 launch beluga_example perfect_odometry.launch.xml localization_package:=nav2_amcl localization_node:=amcl localization_prefix:="xterm -e gdb --args "
```
Once in `xterm` start the amcl process with the `run` command and wait until it crashes. Then use the `back` command.

![Screenshot from 2023-09-02 18-36-45](https://github.com/Ekumen-OS/beluga/assets/17802342/52a6dc02-ec4f-475b-9820-392e37c34613)

glpuga:
This is somehow related to the launchfiles being broken and loading the wrong param file in #254 . When the right param file is loaded, nav2_amcl works fine.

The failure may then be related to some value in the default param yaml file.

nahueespinosa:
This might be related to https://github.com/Ekumen-OS/beluga/pull/238#issuecomment-1629892401. I think we agreed not to use Humble to benchmark.

Most likely, adaptive recovery is enabled in the default param yaml file, which causes the crash. We do need to fix the launch files.

glpuga:
[...]

PR #254, opened 2023-09-03
Add updated performance report

glpuga:
### Proposed changes

This PR adds a new performance report, since the latest one predates important performance enhancing changes in the code.

This PR also identified #253 and had to circumvent it to make the actual capture of metrics.

**This results need to be independently replicated before merging.** The results are out of line with the latest results for a similar code version, and I've don't have a full explanation for the differences (see https://github.com/Ekumen-OS/beluga/pull/233 )

beam_beluga_vs_beluga_vs_amcl

likelihood_beluga_vs_beluga_vs_amcl

#### Type of change

### Checklist


glpuga:
@hidmic @nahueespinosa since these results are so much different from the latest ones we got, and I can't really explain all the differences, this needs to be thoroughly reviewed and replicated before merging.

glpuga:
Forget about it.

The launch files are broken, all the measurements are using the same configuration.

PR #260, opened 2023-09-24
Make particle cloud message weights represent pdf

glpuga:
### Proposed changes

This change makes the weights in the particle cloud message be representative of the estimated probability distribution.

Screenshot from 2023-09-23 21-22-15

#### Type of change

### Checklist


beluga_amcl/src/amcl_node.cpp
```diff
@@ -816,19 +818,71 @@ void AmclNode::timer_callback() {
...
+       ranges::views::zip(particle_filter_->states_view(), particle_filter_->weights_view())) {
+    auto it = representatives_map.find(state);
+    if (it != representatives_map.end()) {
+      auto& representative = it->second;
+      representative.weight += weight;
+      if (representative.weight > max_weight) {
+        max_weight = representative.weight;
+      }
+    } else {
+      representatives_map.emplace(state, RepresentativeData{state, weight});
```
glpuga, line 871: `max_weight` should also be updated here.

glpuga: Fixed.

beluga_amcl/src/amcl_node.cpp
```diff
@@ -816,19 +818,67 @@ void AmclNode::timer_callback() {
     return;
   }
 
+  // Particle weights from the filter may or may not be representative of the
+  // true distribution. If we resampled, they are not, and there will be multiple copies
+  // of the most likely candidates, all with unit weight. In this case the number of copies
```
hidmic, line 823: @glpuga meta: it makes me wonder, are we copying more than we should? Resampling over particle indices (whatever those indices may be) to build up weights instead of duplicating particles wouldn't be particularly hard. Not sure if sequential copying trumps random access because of caching though.

beluga_amcl/src/amcl_node.cpp
```diff
@@ -816,19 +818,67 @@ void AmclNode::timer_callback() {
...
+
+  for (const auto& [state, weight] :
+       ranges::views::zip(particle_filter_->states_view(), particle_filter_->weights_view())) {
+    auto& representative = representatives_map[state];  // if the element does not exist, create it
+    representative.state = state;
+    representative.weight += weight;
+    if (representative.weight > max_weight) {
+      max_weight = representative.weight;
+    }
+  }
```
hidmic, line 869: @glpuga meta: the ability to retrieve a normalized probability distribution from the particle filter sounds useful. Should this functionality be part of the filter itself? Or at least a core utility.

hidmic (commented):
LGTM, @nahueespinosa PTAL!

beluga_amcl/src/amcl_node.cpp
glpuga: I thought about it, but from what I saw, unless the estimation algorithm reconstructs the pdf to make an estimation, there little use for this for the particle filter itself. It's also kind of an expensive luxury, https://en.wikipedia.org/wiki/Kernel_density_estimation .

RViz2, on the other hand, would probably benefit from this as a better and more clear representation that those currently available (arrow, other arrows, and triads).

beluga_amcl/src/amcl_node.cpp
glpuga: The benefits would also not last long, since the next sampling would already force all the particles to become different and make indexing an overhead.

Also we sample far more frequently than we resample; having multiple identical copies is the exemption more than the rule. It just happens that we basically only publish this message when we resample (assuming the usual resample_interval == 1 case), so dealing with that situation is important for this function, but not so much from a performance perspective.

hidmic (approved):
Works as expected, LGTM!

PR #266, opened 2023-11-05
Use double buffering in particle storage

hidmic:
### Proposed changes

Pretty much what the title says. This patch fully initializes a particle cloud in the background before replacing the one in foreground.

#### Type of change

### Checklist

### Additional comments

The rationale for double buffering stems from the use of ranges for particle cloud initialization during resampling. Since range evaluation is lazy, the resampling algorithm would end up reading from the same particle set it is mutating, skewing the results. 

[Nav2 AMCL is careful about this](https://github.com/ros-planning/navigation/blob/9ad644198e132d0e950579a3bc72c29da46e60b0/amcl/src/amcl/pf/pf.c#L377-L378). We were not (and that's how I realized).

### DIFF
--- a/beluga/include/beluga/storage.hpp
+++ b/beluga/include/beluga/storage.hpp
@@ -177,7 +177,7 @@ class StoragePolicy : public Mixin {
    */
   template <class Range>
   void initialize_particles(Range&& input) {
-    static_assert(std::is_same_v<particle_type, ranges::range_value_t<Range>>, "Invalid value type");
+    static_assert(std::is_convertible_v<ranges::range_value_t<Range>, particle_type>, "Invalid value type");
     const std::size_t size = this->self().max_samples();
     particles_[1].resize(size);
     const auto first = std::begin(views::all(particles_[1]));
@@ -216,7 +216,7 @@ class StoragePolicy : public Mixin {
   [[nodiscard]] auto weights() const { return views::weights(particles_[0]) | ranges::views::const_; }
 
  private:
-  Container particles_[2];
+  std::array<Container, 2> particles_;
 };
 
 /// A storage policy that implements a structure of arrays layout.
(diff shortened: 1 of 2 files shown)


beluga/include/beluga/storage.hpp
```diff
@@ -179,10 +179,11 @@ class StoragePolicy : public Mixin {
   void initialize_particles(Range&& input) {
     static_assert(std::is_same_v<particle_type, ranges::range_value_t<Range>>, "Invalid value type");
     const std::size_t size = this->self().max_samples();
-    particles_.resize(size);
-    const auto first = std::begin(views::all(particles_));
+    particles_[1].resize(size);
```
hidmic, line 182: Mutating while iterating issues never triggered SEGFAULTs because `particles_` is always expanded to fit `max_samples` and [`std::vector::resize`](https://en.cppreference.com/w/cpp/container/vector/resize) does not change vector capacity.

beluga/include/beluga/storage.hpp
```diff
@@ -179,10 +179,11 @@ class StoragePolicy : public Mixin {
...
+    const auto first = std::begin(views::all(particles_[1]));
     const auto last = ranges::copy(input | ranges::views::take(size), first).out;
```
hidmic, line 184: This is one of the knots of the problematic execution path. The other two are [here](https://github.com/Ekumen-OS/beluga/blob/31445270321482c72dc8a7608385a0682880151b/beluga/include/beluga/algorithm/sampling.hpp#L381) and [here](https://github.com/Ekumen-OS/beluga/blob/31445270321482c72dc8a7608385a0682880151b/beluga/include/beluga/algorithm/sampling.hpp#L162). Sampling an existing state means reading from the same `particles_` we are writing to.

glpuga (approved):
Great catch! LGTM

nahueespinosa (approved):
LGTM!

PR #267, opened 2023-11-05
Fix differential drive motion model

hidmic:
### Proposed changes

Going over the math a few times, it doesn't seem to me that the rotation variance computation distributes over rotation composition. Not with those nonlinear max, min, abs operations in between.

#### Type of change

### Checklist

### Additional comments

Not sure how to write a test for this. Would a formal proof (that `rotation_variance(a) * rotation_variance(b) != rotation_variance(a * b)`) do?

### DIFF
--- a/beluga/include/beluga/motion/differential_drive_model.hpp
+++ b/beluga/include/beluga/motion/differential_drive_model.hpp
@@ -155,8 +155,8 @@ class DifferentialDriveModel : public Mixin {
 
   static double rotation_variance(const Sophus::SO2d& rotation) {
     // Treat backward and forward motion symmetrically for the noise models.
-    static const auto flipping_rotation = Sophus::SO2d{Sophus::Constants<double>::pi()};
-    const auto flipped_rotation = rotation * flipping_rotation;
+    static const auto kFlippingRotation = Sophus::SO2d{Sophus::Constants<double>::pi()};
+    const auto flipped_rotation = rotation * kFlippingRotation;
     const auto delta = std::min(std::abs(rotation.log()), std::abs(flipped_rotation.log()));
     return delta * delta;
   }
--- a/beluga/include/beluga/motion/differential_drive_model.hpp
+++ b/beluga/include/beluga/motion/differential_drive_model.hpp
@@ -123,7 +123,6 @@ class DifferentialDriveModel : public Mixin {
               ? Sophus::SO2d{std::atan2(translation.y(), translation.x())} * previous_orientation.inverse()
               : Sophus::SO2d{0.0};
       const auto second_rotation = current_orientation * previous_orientation.inverse() * first_rotation.inverse();
-      const auto combined_rotation = first_rotation * second_rotation;
 
       {
         first_rotation_params_ = DistributionParam{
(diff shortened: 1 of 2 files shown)


glpuga:
> Going over the math a few times, it doesn't seem to me that the rotation variance computation distributes over rotation composition. Not with those nonlinear max, min, abs operations in between.

I totally agree that that operation most definitely is not distributive. 

However, I have broader concerns regarding the `rotation_variance()` function itself. From what I gather it's meant to bring the 2nd/3rd quadrant rotation values that result from `first_rotation` when moving backwards because $\Delta x$ is negative (and square the value at the same time).

But that seems to be done at the _end_ of all the math. By then, we already calculated the second rotation using the pre-adjustment first rotation, and it's not obvious to me that that can be fixed by moving second rotation back into 1st/4th quadrant too afterwards.

Not having gone through t all the math in Prob. Robotics, the definitions of $\delta_{rot1}$ and $\delta_{rot2}$ seem very straightforward (the aperture angles between the end/star pose and the direction between the end/start pose, range $-\pi$ to $\pi$). Maybe we should just do the quadrant adjustment once when calculating first $\delta_{rot1}$ the first time; that should take care of $\delta_{rot2}$ right away.

> ### Additional comments
> Not sure how to write a test for this. Would a formal proof (that `rotation_variance(a) * rotation_variance(b) != rotation_variance(a * b)`) do?

From this code we don't really have good observability to the results of this function, which would be the best way to test this. Given that the parameters are private you can only test by generating samples of the distribution and comparing it, which is not ideal.

I think it's worth either

1. providing observability to the distribution parameter values through getters for testing's sake (not a fan),
2. better yet, architecting this differently so that the actual code that does this critical calculation is in an external function/class that can be tested independently, feeding in `pose` and `last_pose`, and getting the distribution parameters at the other end, like we do with `calculate_covariance()` in the estimation mixin.

In this last scenario `update_motion()` would just call this function to calculate the distribution parameters, and manage the mixin `last_pose_` state.

beluga/include/beluga/motion/differential_drive_model.hpp
```diff
@@ -123,7 +123,6 @@ class DifferentialDriveModel : public Mixin {
               ? Sophus::SO2d{std::atan2(translation.y(), translation.x())} * previous_orientation.inverse()
               : Sophus::SO2d{0.0};
```
glpuga, line 124: nit: from my experience with perf, `SO2d{0.0}` does a bunch of caculations to determine real and imaginary. `SO2d{}` is much faster because it just sets the right values. Not that it makes a lot of difference here.

glpuga:
> FWIW, we have tests that generate samples and assert distribution parameter values, no need to provide observability to the internal state.
> 
> https://github.com/Ekumen-OS/beluga/blob/9de0417ce9c5d0a22af021b16bc9a44ed665da2d/beluga/test/beluga/motion/test_differential_drive_model.cpp#L151

Good point. Bears asking why that test in insensitive to this fix then. 

In that test, the translation distance is zero, so `first_rotation` would be zero and the result both after and before the change would be the same, right?.

nahueespinosa:
> In that test, the translation distance is zero, so first_rotation would be zero and the result both after and before the change would be the same, right?.

Right, I would imagine a new test where all parameters but `translation_noise_from_rotation` are zero, and we set a move that requires two rotations. I volunteer to give it a shot tonight if it helps.

nahueespinosa (approved):

beluga/include/beluga/motion/differential_drive_model.hpp
hidmic: Addressed in https://github.com/Ekumen-OS/beluga/pull/267/commits/2ed0708d68dda5bffd9d0274cc7c74e8cc440e4b.

PR #270, opened 2023-12-10
Add `beluga_ros` package

nahueespinosa:
### Proposed changes

This patch creates a `beluga_ros` package in order to decouple common functionality to interface with ROS from `beluga_amcl`. The goal is to reduce the boilerplate needed to create a new ROS localization package using `beluga`.

#### Type of change

### Checklist

### Additional comments

More utilities and refactoring coming up in following PRs.


beluga_amcl/cmake/ament.cmake
```diff
@@ -66,17 +54,15 @@ rclcpp_components_register_node(
...
   rclcpp_components
   rclcpp_lifecycle
-  sensor_msgs
-  tf2
-  tf2_eigen
-  tf2_geometry_msgs)
+  std_srvs)
 ament_export_include_directories("include/${PROJECT_NAME}")
-ament_export_definitions(BELUGA_AMCL_ROS_VERSION=2)
+ament_export_definitions(BELUGA_ROS_VERSION=2)
```
hidmic, line 65: @nahueespinosa meta: should `beluga_amcl` re-export definitions already exported by `beluga_ros`? If we need to condition compilation in `beluga_amcl` based on the ROS distribution version, in addition to that in `beluga_ros` already (it doesn't look like it), shouldn't the definition remain `BELUGA_AMCL_ROS_VERSION`? Same for ROS 1.

beluga_amcl/include/beluga_amcl/amcl_node_utils.hpp
```diff
@@ -46,7 +46,7 @@ inline Eigen::IOFormat make_eigen_comma_format() {
  * \return A vector of pairs of coordinates in the robot's reference frame.
  */
 std::vector<std::pair<double, double>> make_points_from_laser_scan(
-    const beluga_amcl::messages::LaserScan& laser_scan,
+    const beluga_ros::msg::LaserScan& laser_scan,
```
hidmic, line 49: @nahueespinosa meta: doesn't this function belong to `beluga_ros`?

beluga_ros/test/test_occupancy_grid.cpp
```diff
@@ -0,0 +1,71 @@
...
+  constexpr std::size_t kWidth = 100;
+  constexpr std::size_t kHeight = 200;
+
+#if BELUGA_ROS_VERSION == 2
+  auto message = std::make_shared<beluga_ros::msg::OccupancyGrid>();
+#elif BELUGA_ROS_VERSION == 1
+  auto message = boost::make_shared<beluga_ros::msg::OccupancyGrid>();
+#else
+#error BELUGA_ROS_VERSION is not defined or invalid
+#endif
```
hidmic, line 34: @nahueespinosa meta: should we have a `beluga_ros::utility` namespace to hide this?

hidmic (commented):
Looks great @nahueespinosa! Left some comments here and there.

beluga_amcl/cmake/ament.cmake
nahueespinosa: Ah, thanks for catching that! Fixed in https://github.com/Ekumen-OS/beluga/pull/270/commits/f209efc8852159635c04fd0b6269ff4e75f4627e.

beluga_amcl/include/beluga_amcl/amcl_node_utils.hpp
nahueespinosa: It does, I will address this in #170. Coming shortly.

beluga_ros/test/test_occupancy_grid.cpp
nahueespinosa: For now, this is the only place I've had to do this, just for a test, so I'd rather wait and see if we need to implement our own `utility` namespace yet.

hidmic: Sounds reasonable to me.

hidmic (approved):
LGTM

PR #273, opened 2023-12-25
Improve spatial_hash function to avoid spatial periodicity in the hash function

glpuga:
### Proposed changes

Adds two commits:

- The first one adds periodicity checks that the existing hash function does not pass.
- The second one updates the hash function to pass old and new tests.

Based on:

- https://en.wikipedia.org/wiki/Hash_function#Fibonacci_hashing
- https://probablydance.com/2018/06/16/fibonacci-hashing-the-optimization-that-the-world-forgot-or-a-better-alternative-to-integer-modulo/

#### Type of change

### Checklist

### Additional comments

### DIFF
--- a/beluga/include/beluga/algorithm/spatial_hash.hpp
+++ b/beluga/include/beluga/algorithm/spatial_hash.hpp
@@ -34,22 +34,31 @@ namespace beluga {
 
 namespace detail {
 
-/// Returns the floor of a value shifted.
+/// Returns the hashed and rotated floor of a value.
 /**
  * \tparam N Number of bits to be used from the integer result, the least significant will be used.
  * \tparam I Result will be shifted by I*N.
- * \param value Input value to be floored and shifted.
+ * \param value Input value to be hashed.
  * \return The calculated result.
  */
 template <std::size_t N, std::size_t I>
-constexpr std::size_t floor_and_shift(double value) {
-  // Compute the largest integer value not greater than value.
-  auto signed_value = static_cast<std::intmax_t>(std::floor(value));
-  // Compute the smallest unsigned integer equal to the source value modulo 2^n
-  // (where n is the number of bits used to represent the destination type).
-  auto unsigned_value = static_cast<std::uintmax_t>(signed_value);
-  // Create a fixed-size sequence of N bits and perform a binary shift left I * N positions.
-  return std::bitset<N>{unsigned_value}.to_ullong() << (I * N);
+constexpr std::size_t floor_and_fibo_hash(double value) {
+  static_assert(std::is_same_v<std::size_t, std::uint64_t>);
+  constexpr auto kFib64 = 11400714819323198485LLU;  // golden ratio for 64 bits
+  // floor the value and convert to integer
+  const auto signed_value = static_cast<std::int64_t>(std::floor(value));
+  // work with unsigned from now on
+  const auto unsigned_value = static_cast<std::uint64_t>(signed_value);
+  // spread number information all through the 64 bits using the fibonacci hash
+  const auto div_hashed_value = kFib64 * unsigned_value;
+  // rotate bits to avoid aliasing between different values of I
+  if constexpr (N * I != 0) {
+    const auto left_hash = (div_hashed_value << N * I);
+    const auto right_hash = (div_hashed_value >> (64 - N * I));
+    return left_hash | right_hash;
+  } else {
+    return div_hashed_value;
+  }
 }
 
 /// Hashes a tuple or array of scalar types, using a resolution for each element and using the same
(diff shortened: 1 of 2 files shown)


beluga/include/beluga/algorithm/spatial_hash.hpp
```diff
@@ -34,22 +34,29 @@ namespace beluga {
...
-constexpr std::size_t floor_and_shift(double value) {
-  // Compute the largest integer value not greater than value.
-  auto signed_value = static_cast<std::intmax_t>(std::floor(value));
-  // Compute the smallest unsigned integer equal to the source value modulo 2^n
-  // (where n is the number of bits used to represent the destination type).
-  auto unsigned_value = static_cast<std::uintmax_t>(signed_value);
-  // Create a fixed-size sequence of N bits and perform a binary shift left I * N positions.
-  return std::bitset<N>{unsigned_value}.to_ullong() << (I * N);
+constexpr std::size_t floor_and_fibo_hash(double value) {
+  constexpr auto kFib64 = 11400714819323198485LLU;  // golden ratio for 64 bits
```
nahueespinosa, line 46: Can we assert that `std::size_t` is the same as `std::unit64_t`?
```cpp
static_assert(std::is_same_v<std::size_t, std::uint64_t>);
```

beluga/include/beluga/algorithm/spatial_hash.hpp
```diff
@@ -34,22 +34,29 @@ namespace beluga {
...
+  const auto unsigned_value = static_cast<std::uint64_t>(signed_value);
+  // spread number information all through the 64 bits using the fibonacci hash
+  const auto div_hashed_value = kFib64 * unsigned_value;
+  // rotate bits to avoid aliasing between different values of I
+  const auto left_hash = (div_hashed_value << N * I);
+  if constexpr (N * I != 0) {
+    return left_hash | (div_hashed_value >> (64 - N * I));
+  } else {
+    return left_hash;
+  }
```
nahueespinosa, line 59: nit^n:

```suggestion
  if constexpr (N * I != 0) {
    const auto left_hash = (div_hashed_value << N * I);
    const auto right_hash = (div_hashed_value >> (64 - N * I));
    return left_hash | right_hash;
  } else {
    return div_hashed_value;
  }
```

beluga/include/beluga/algorithm/spatial_hash.hpp
```diff
@@ -34,22 +34,29 @@ namespace beluga {
...
-  auto unsigned_value = static_cast<std::uintmax_t>(signed_value);
-  // Create a fixed-size sequence of N bits and perform a binary shift left I * N positions.
-  return std::bitset<N>{unsigned_value}.to_ullong() << (I * N);
+constexpr std::size_t floor_and_fibo_hash(double value) {
+  constexpr auto kFib64 = 11400714819323198485LLU;  // golden ratio for 64 bits
+  // floor the value and convert to integer
+  const auto signed_value = static_cast<std::int64_t>(std::floor(value));
+  // work with unsigned from now on
+  const auto unsigned_value = static_cast<std::uint64_t>(signed_value);
+  // spread number information all through the 64 bits using the fibonacci hash
```
nahueespinosa, line 51: meta: The previous implementation trivially ensured that no collisions would happen within a defined range of 2^N, given N = 64 / amount_of_elements.  For 2D (x, y, theta) this seemed good enough, and the periodicity effect was expected for values outside the 2^N range.

Did you take that into account when implementing the new version?

nahueespinosa (commented):
@glpuga This is super interesting! What's the motivation behind this? Is this an improvement to better support 3D (6-DOF) or is it something else?

beluga/include/beluga/algorithm/spatial_hash.hpp
glpuga: Done!

beluga/include/beluga/algorithm/spatial_hash.hpp
glpuga: The new version does not guarantee collisions for any specific range, but it should ensure that any hash-space collision is not systematic.

Collisions can happen due to the superposition of multiple dimensions in the hash (x,  y, theta, etc.), since for each single dimension the multiplicative hash will make sure the hashes [end up as spread as possible](https://www.youtube.com/watch?v=lOIP_Z_-0Hs).

Regardless, this implementation satisfies all preexisting tests, including the one checking for duplicates in a volume for the 3-tuple case.

Notice that there's one more layer to the collisions problem. `unordered_map` and `unordered_set` will funnel the hash space created by this function into a much narrower $2^N$ space given by the size of the underlying hash table, using a second compiler-defined hash function for that.

Depending on what hash function is used for this, non-uniform distribution of the information in the key can cause poor performance. For example, some compiler implementations just chop the higher bits with a mask for the sake of speed, so in our case that would mean that for a hash table smaller than a few thousand buckets, the hash bucket would be exclusively determined by the X value, regardless of Y and Theta. GCC seems to use a residue function, which is better than this, but still, custom-made hash functions that are later fed into maps/sets should ideally spread information more or less uniformly in the hash to avoid this kind of problem.

nahueespinosa: Thanks for the explanation! Makes sense to me.

nahueespinosa (approved):
LGTM!

Issue #258, opened 2023-09-06
Clustering for pose estimation


hidmic:
### Feature description

Connected to https://github.com/Ekumen-OS/beluga/discussions/172. Good ol' AMCL computes sampled mean and covariance from particle clusters, whereas Beluga AMCL computes sampled mean and covariance for the entire posterior particle cloud. Regardless of any arguments in favor or against either approach, Beluga AMCL should offer the choice. We need another estimation mixin.

glpuga:
Do we know any paper/source for the implementation of this by AMCL?

hidmic:
Not that I know of. That said, the idea of clustering has been floating around for a long while e.g. https://arxiv.org/pdf/cs/0204044.pdf.

hidmic:
Addressed as of #275.

PR #275, opened 2023-12-27
Add cluster based estimator

glpuga:
### Proposed changes

Addresses #258 and #172.

Add new cluster based estimator. 

- Splits the space in a grid and determines the accumulated weight in each cell.
- Starting from the weight peaks, it groups cells into clusters of ever decreasing weight. It creates as many clusters as needed to cover all the particle occupied space.
- Estimates mean and covariance for each.
- Returns the mean and covariance of the cluster with the highest cumulative weight.

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/algorithm.hpp
+++ b/beluga/include/beluga/algorithm.hpp
@@ -20,6 +20,7 @@
  * \brief Includes all beluga algorithms.
  */
 
+#include <beluga/algorithm/cluster_based_estimation.hpp>
 #include <beluga/algorithm/distance_map.hpp>
 #include <beluga/algorithm/effective_sample_size.hpp>
 #include <beluga/algorithm/estimation.hpp>
--- a/beluga/include/beluga/algorithm/cluster_based_estimation.hpp
+++ b/beluga/include/beluga/algorithm/cluster_based_estimation.hpp
@@ -64,7 +64,7 @@ template <class Map, class Proj>
 [[nodiscard]] auto make_priority_queue(const Map& map, Proj&& proj) {
   struct KeyWithPriority {
     double priority;  // priority value used to order the queue (higher value first).
-    std::size_t key;  // hash of the cell in the grid cell data map.
+    typename Map::key_type key;
 
     bool operator<(const KeyWithPriority& other) const {
       return priority < other.priority;  // sort in decreasing priority order
@@ -94,154 +94,155 @@ template <class Map, class Proj>
 template <class Range>
 [[nodiscard]] auto calculate_percentile_threshold(Range&& range, double percentile) {
   auto values = range | ranges::to<std::vector>;
-  const auto n = static_cast<std::size_t>(static_cast<double>(values.size()) * percentile);
+  const auto n = static_cast<std::ptrdiff_t>(static_cast<double>(values.size()) * percentile);
   std::nth_element(values.begin(), values.begin() + n, values.end());
-  return values[n];
+  return values[static_cast<std::size_t>(n)];
 }
 
-/// Implementation utilities for a particle clusterizer algorithm.
-struct ParticleClusterizerImpl {
-  /// A struct that holds the data of a single cell for the clusterization algorithm.
-  template <class State>
-  struct Cell {
-    State representative_state;             ///< state of a particle that is within the cell
-    double weight{0.0};                     ///< average weight of the cell
-    std::size_t num_particles{0};           ///< number of particles in the cell
-    std::optional<std::size_t> cluster_id;  ///< cluster id of the cell
-  };
+namespace clusterizer_detail {
 
-  /// A map that holds the sparse data about the particles grouped in cells.
-  template <class State>
-  using Map = std::unordered_map<std::size_t, Cell<State>>;
-
-  /// Create a cluster map from a range of particles and their corresponding spatial hashes.
-  /**
-   * This method will populate all the relevant fields in the map except for the cluster ID,
-   * which has to be computed with a separate call to `assign_clusters`.
-   *
-   * \tparam States The range type for particle states.
-   * \tparam Weights The range type for particle weights.
-   * \tparam Hashes The range type for particle spatial hashes.
-   * \param states A range of particle states.
-   * \param weights A range of particle weights.
-   * \param hashes A range of particle spatial hashes.
-   * \return A map where each unique hash corresponds to a cell with accumulated weight, particle count, and
-   * representative state.
-   */
-  template <class States, class Weights, class Hashes>
-  [[nodiscard]] static auto make_cluster_map(States&& states, Weights&& weights, Hashes&& hashes) {
-    using State = ranges::range_value_t<States>;
-    Map<State> map;
-
-    // Preallocate memory with a very rough estimation of the number of cells we might end up with.
-    map.reserve(states.size() / 5);
-
-    // Calculate the accumulated cell weight and save a single representative state for each cell.
-    for (const auto& [state, weight, hash] : ranges::views::zip(states, weights, hashes)) {
-      auto [it, inserted] = map.try_emplace(hash, Cell<State>{});
-      auto& [_, entry] = *it;
-      entry.weight += weight;
-      entry.num_particles++;
-      if (inserted) {
-        entry.representative_state = state;
-      }
-    }
-
-    return map;
-  }
-
-  /// Normalize weights and cap them to a given percentile.
-  /**
-   * Given a valid cluster map, normalize the accumulated weight by the number of particles
-   * in each cell to avoid biasing the clustering algorithm towards cells that randomly end up
-   * with more particles than others.
-   *
-   * Then cap the values to a given percentile to flatten the top of the approximated
-   * density function and make the clustering algorithm more robust to estimation noise,
-   * by fusing together any adjacent peaks whose weight is above the threshold.
-   *
-   * \tparam State The state type of the cells in the map.
-   * \param map A reference to the map where cells are stored.
-   * \param percentile The percentile threshold for capping the weights.
-   */
-  template <class State>
-  static void normalize_and_cap_weights(Map<State>& map, double percentile) {
-    auto entries = ranges::views::values(map);
-
-    for (auto& entry : entries) {
-      assert(entry.num_particles > 0);
-      entry.weight /= static_cast<double>(entry.num_particles);
-    }
-
-    const double max_weight =
-        calculate_percentile_threshold(ranges::views::transform(entries, &Cell<State>::weight), percentile);
-
-    for (auto& entry : entries) {
-      entry.weight = std::min(entry.weight, max_weight);
-    }
-  }
-
-  /// Assign cluster ids to an existing cluster map.
-  /**
-   * This function implements a clustering algorithm that assigns cluster IDs to cells in a map
-   * based on their spatial relationships.
-   *
-   * Notice that with this algorithm each cell will go through the priority queue at most twice.
-   *
-   * \tparam State The state type of the cells in the map.
-   * \tparam NeighborsFunction A callable object that, given a state, returns a range of neighboring cell hashes.
-   * \param map A reference to the map where cells are stored.
-   * \param neighbors_function A function that returns neighboring cell hashes for a given state.
-   */
-  template <class State, class NeighborsFunction>
-  static void assign_clusters(Map<State>& map, NeighborsFunction&& neighbors_function) {
-    auto queue = make_priority_queue(map, &Cell<State>::weight);
-    const auto max_priority = queue.top().priority;
-
-    std::size_t next_cluster_id = 0;
-
-    // Process cells in order of descending weight.
-    while (!queue.empty()) {
-      const auto hash = queue.top().key;
-      queue.pop();
-
-      // The hash is guaranteed to exist in the map.
-      auto& cell = map[hash];
-
-      // If there's no cluster id assigned to the cell, assign a new one.
-      if (!cell.cluster_id.has_value()) {
-        cell.cluster_id = next_cluster_id++;
-      }
-
-      // Process the neighbors of the cell; if they don't have a cluster id already assigned
-      // then assign them one and add them to the queue with an inflated priority
-      // to ensure they get processed ASAP before moving on to other local peaks.
-      // Notice that with this algorithm each cell will go through the priority queue at most twice.
-
-      // Define a lambda to check if a neighboring cell is valid for clustering.
-      const auto is_valid_neighbor = [&](const auto& neighbor_hash) {
-        auto it = map.find(neighbor_hash);
-        return (
-            (it != map.end()) &&                     // is within the map
-            (!it->second.cluster_id.has_value()) &&  // AND not yet mapped to a cluster
-            (it->second.weight <= cell.weight));     // AND has lower weight than the current cell
-      };
-
-      // Process the neighbors of the current cell.
-      for (const std::size_t neighbor_hash : neighbors_function(cell.representative_state) |  //
-                                                 ranges::views::cache1 |                      //
-                                                 ranges::views::filter(is_valid_neighbor)) {
-        auto& neighbor = map[neighbor_hash];
-        neighbor.cluster_id = cell.cluster_id;
-        queue.push({max_priority + neighbor.weight, neighbor_hash});  // reintroduce with inflated priority
-      }
-    }
-  }
+/// A struct that holds the data of a single cell for the clusterization algorithm.
+template <class State>
+struct Cell {
+  State representative_state;             ///< state of a particle that is within the cell
+  double weight{0.0};                     ///< average weight of the cell
+  std::size_t num_particles{0};           ///< number of particles in the cell
+  std::optional<std::size_t> cluster_id;  ///< cluster id of the cell
 };
+
+/// A map that holds the sparse data about the particles grouped in cells.
+template <class State>
+using Map = std::unordered_map<std::size_t, Cell<State>>;
+
+/// Create a cluster map from a range of particles and their corresponding spatial hashes.
+/**
+ * This method will populate all the relevant fields in the map except for the cluster ID,
+ * which has to be computed with a separate call to `assign_clusters`.
+ *
+ * \tparam States The range type for particle states.
+ * \tparam Weights The range type for particle weights.
+ * \tparam Hashes The range type for particle spatial hashes.
+ * \param states A range of particle states.
+ * \param weights A range of particle weights.
+ * \param hashes A range of particle spatial hashes.
+ * \return A map where each unique hash corresponds to a cell with accumulated weight, particle count, and
+ * representative state.
+ */
+template <class States, class Weights, class Hashes>
+[[nodiscard]] static auto make_cluster_map(States&& states, Weights&& weights, Hashes&& hashes) {
+  using State = ranges::range_value_t<States>;
+  Map<State> map;
+
+  // Preallocate memory with a very rough estimation of the number of cells we might end up with.
+  map.reserve(states.size() / 5);
+
+  // Calculate the accumulated cell weight and save a single representative state for each cell.
+  for (const auto& [state, weight, hash] : ranges::views::zip(states, weights, hashes)) {
+    auto [it, inserted] = map.try_emplace(hash, Cell<State>{});
+    auto& [_, entry] = *it;
+    entry.weight += weight;
+    entry.num_particles++;
+    if (inserted) {
+      entry.representative_state = state;
+    }
+  }
+
+  return map;
+}
+
+/// Normalize weights and cap them to a given percentile.
+/**
+ * Given a valid cluster map, normalize the accumulated weight by the number of particles
+ * in each cell to avoid biasing the clustering algorithm towards cells that randomly end up
+ * with more particles than others.
+ *
+ * Then cap the values to a given percentile to flatten the top of the approximated
+ * density function and make the clustering algorithm more robust to estimation noise,
+ * by fusing together any adjacent peaks whose weight is above the threshold.
+ *
+ * \tparam State The state type of the cells in the map.
+ * \param map A reference to the map where cells are stored.
+ * \param percentile The percentile threshold for capping the weights.
+ */
+template <class State>
+static void normalize_and_cap_weights(Map<State>& map, double percentile) {
+  auto entries = ranges::views::values(map);
+
+  for (auto& entry : entries) {
+    assert(entry.num_particles > 0);
+    entry.weight /= static_cast<double>(entry.num_particles);
+  }
+
+  const double max_weight =
+      calculate_percentile_threshold(ranges::views::transform(entries, &Cell<State>::weight), percentile);
+
+  for (auto& entry : entries) {
+    entry.weight = std::min(entry.weight, max_weight);
+  }
+}
+
+/// Assign cluster ids to an existing cluster map.
+/**
+ * This function implements a clustering algorithm that assigns cluster IDs to cells in a map
+ * based on their spatial relationships.
+ *
+ * Notice that with this algorithm each cell will go through the priority queue at most twice.
+ *
+ * \tparam State The state type of the cells in the map.
+ * \tparam NeighborsFunction A callable object that, given a state, returns a range of neighboring cell hashes.
+ * \param map A reference to the map where cells are stored.
+ * \param neighbors_function A function that returns neighboring cell hashes for a given state.
+ */
+template <class State, class NeighborsFunction>
+static void assign_clusters(Map<State>& map, NeighborsFunction&& neighbors_function) {
+  auto queue = make_priority_queue(map, &Cell<State>::weight);
+  const auto max_priority = queue.top().priority;
+
+  std::size_t next_cluster_id = 0;
+
+  // Process cells in order of descending weight.
+  while (!queue.empty()) {
+    const auto hash = queue.top().key;
+    queue.pop();
+
+    // The hash is guaranteed to exist in the map.
+    auto& cell = map[hash];
+
+    // If there's no cluster id assigned to the cell, assign a new one.
+    if (!cell.cluster_id.has_value()) {
+      cell.cluster_id = next_cluster_id++;
+    }
+
+    // Process the neighbors of the cell; if they don't have a cluster id already assigned
+    // then assign them one and add them to the queue with an inflated priority
+    // to ensure they get processed ASAP before moving on to other local peaks.
+    // Notice that with this algorithm each cell will go through the priority queue at most twice.
+
+    // Define a lambda to check if a neighboring cell is valid for clustering.
+    const auto is_valid_neighbor = [&](const auto& neighbor_hash) {
+      auto it = map.find(neighbor_hash);
+      return (
+          (it != map.end()) &&                     // is within the map
+          (!it->second.cluster_id.has_value()) &&  // AND not yet mapped to a cluster
+          (it->second.weight <= cell.weight));     // AND has lower weight than the current cell
+    };
+
+    // Process the neighbors of the current cell.
+    for (const std::size_t neighbor_hash : neighbors_function(cell.representative_state) |  //
+                                               ranges::views::cache1 |                      //
+                                               ranges::views::filter(is_valid_neighbor)) {
+      auto& neighbor = map[neighbor_hash];
+      neighbor.cluster_id = cell.cluster_id;
+      queue.push({max_priority + neighbor.weight, neighbor_hash});  // reintroduce with inflated priority
+    }
+  }
+}
+
+}  // namespace clusterizer_detail
 
 /// Parameters used to construct a ParticleClusterizer instance.
 struct ParticleClusterizerParam {
-  double spatial_hash_resolution = 0.20;   ///< clustering algorithm spatial resolution
+  double linear_hash_resolution = 0.20;    ///< clustering algorithm linear resolution
   double angular_hash_resolution = 0.524;  ///< clustering algorithm angular resolution
 
   /// Cluster weight cap threshold.
@@ -286,9 +287,9 @@ class ParticleClusterizer {
     auto weights = beluga::views::weights(particles);
     auto hashes = states | ranges::views::transform(spatial_hash_function_) | ranges::to<std::vector>;
 
-    auto map = ParticleClusterizerImpl::make_cluster_map(states, weights, hashes);
-    ParticleClusterizerImpl::normalize_and_cap_weights(map, parameters_.weight_cap_percentile);
-    ParticleClusterizerImpl::assign_clusters(map, [this](const auto& state) { return neighbors(state); });
+    auto map = clusterizer_detail::make_cluster_map(states, weights, hashes);
+    clusterizer_detail::normalize_and_cap_weights(map, parameters_.weight_cap_percentile);
+    clusterizer_detail::assign_clusters(map, [this](const auto& state) { return neighbors(state); });
 
     return hashes |  //
            ranges::views::transform([&map](std::size_t hash) { return map[hash].cluster_id.value(); }) |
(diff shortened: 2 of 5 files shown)


beluga/test/beluga/estimation/test_cluster_based_estimation.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+  [[nodiscard]] static auto
+  make_particle_multicluster_dataset(double xmin, double xmax, double ymin, double ymax, double step) {
+    std::vector<SE2d> states;
+    std::vector<double> weights;
+
+    const auto xwidth = xmax - xmin;
+    const auto ywidth = ymax - ymin;
+
+    // simulate particles in a grid with (phases x phases) clusters with
```
glpuga, line 70: "Phases x phases"; this does not exist anymore.

beluga/include/beluga/estimation/cluster_based_estimator.hpp
```diff
@@ -0,0 +1,372 @@
...
+   */
+  template <class... Args>
+  explicit ClusterBasedStateEstimator(Args&&... args) : Mixin(std::forward<Args>(args)...) {}
+
+  /// \copydoc EstimationInterface2d::estimate()
+  [[nodiscard]] std::pair<StateType, CovarianceType> estimate() const final;
+
+ private:
+  static constexpr double kSpatialHashResolution = 0.20;   ///< clustering algorithm spatial resolution
+  static constexpr double kAngularHashResolution = 0.524;  ///< clustering algorithm angular resolution
```
glpuga, line 322: These should be parameters.

beluga/include/beluga/estimation/cluster_based_estimator.hpp
```diff
@@ -0,0 +1,372 @@
...
+    std::size_t particle_count{0};
+  };
+  std::unordered_map<std::size_t, ClusterInfo> cluster_info;
+
+  // find out the weight of each of the clusters, along with how many clusters there are
+  for (const auto& [weight, hash] : ranges::views::zip(weights, hashes)) {
+    const auto& grid_cell = grid_cell_data[hash];
+    if (!grid_cell.cluster_id.has_value()) {
+      continue;
+    }
```
nahueespinosa, line 255: Is it possible that we have a hash here that has not been mapped to a cluster?

beluga/include/beluga/estimation/cluster_based_estimator.hpp
```diff
@@ -0,0 +1,372 @@
...
+/// @tparam Weights Range type of the weights.
+/// @tparam Hashes Range type of the hashes.
+/// @param grid_cell_data Grid cell data map that has already been processed by the clustering algorithm.
+/// @param states Range containing the states of the particles.
+/// @param weights Range containing the weights of the particles.
+/// @param hashes Range containing the hashes of the particles.
+/// @return A vector of tuples, containing the weight, mean and covariance of each cluster, in no particular order.
+template <class StateType, class CovarianceType, class GridCellDataMapType, class Range, class Weights, class Hashes>
+[[nodiscard]] auto
+estimate_clusters(GridCellDataMapType&& grid_cell_data, Range&& states, Weights&& weights, Hashes&& hashes) {
```
nahueespinosa, line 243: Can we change the signature to something like this?

```cpp
auto estimate_clusters(Range&& states, Weights&& weights, Clusters&& clusters)
```

I think this is a good candidate for `range_statistics.hpp` if we remove the `GridCellDataMapType` dependency.

I played a bit with an alternative implementation that doesn't remove the dependency but shows how it's a trivial step forward: https://github.com/Ekumen-OS/beluga/commit/a87ce57caf62c7ca9d713847d1a464f7e3c8be79 WDYT?

beluga/include/beluga/estimation/cluster_based_estimator.hpp
```diff
@@ -0,0 +1,372 @@
...
+    if (!grid_cell.cluster_id.has_value()) {
+      continue;
+    }
+    auto [it, inserted] = cluster_info.try_emplace(*grid_cell.cluster_id, ClusterInfo{});
+    it->second.total_weight += weight;
+    ++it->second.particle_count;
+  }
+
+  // for each cluster found, estimate the mean and covariance of the states that belong to it
+  std::vector<std::tuple<double, StateType, CovarianceType>> cluster_estimates;
```
nahueespinosa, line 262: These types can be deduced instead of being template parameters, which would make this function easier to use outside the mixin context.

```suggestion
  using StateType = std::decay_t<decltype(std::get<0>(beluga::estimate(states, weights)))>;
  using CovarianceType = std::decay_t<decltype(std::get<1>(beluga::estimate(states, weights)))>;
  std::vector<std::tuple<double, StateType, CovarianceType>> cluster_estimates;
```

beluga/include/beluga/estimation/cluster_based_estimator.hpp
```diff
@@ -0,0 +1,372 @@
...
+/**
+ * \file
+ * \brief Implementation of a cluster-base estimator mixin.
+ */
+
+namespace beluga {
+
+namespace cse_detail {
+
+/// @brief A struct that holds the data of a single cell in the grid.
```
nahueespinosa, line 52: nit: Use `\` instead of `@`.

beluga/include/beluga/estimation/cluster_based_estimator.hpp
```diff
@@ -0,0 +1,372 @@
...
+      return (
+          (it != grid_cell_data.end()) &&            // is within the map
+          (!it->second.cluster_id.has_value()) &&    // AND not yet mapped to a cluster
+          (it->second.weight <= grid_cell_weight));  // AND has lower weight than the current cell
+    };
+
+    auto valid_neighbor_hashes_view =                     //
+        neighbors |                                       //
+        ranges::views::transform(get_neighbor_hash) |     //
+        ranges::views::filter(filter_invalid_neighbors);  //
```
nahueespinosa, line 216: Between a transform+filter pipeline, it's recommended to add a `ranges::views::cache1`, so the transform is not computed twice for each filter iteration. See https://stackoverflow.com/a/64199729.

nahueespinosa (commented):
@glpuga This is amazing! I didn't get to review the contents of `cse_detail` in much detail yet (no pun intended), but overall everything looks good.

beluga/include/beluga/estimation/cluster_based_estimator.hpp
```diff
@@ -0,0 +1,372 @@
...
+/// @brief Function that creates an vector containing the hashes of each of the states in the input range.
+/// @tparam Range Type of the states range.
+/// @tparam Hasher Hash function type to convert states into hashes.
+/// @param states The range of states.
+/// @param spatial_hash_function_ The hash object instance.
+/// @return A vector containing the hashes of each of the states in the input range.
+template <class Range, class Hasher>
+[[nodiscard]] auto precalculate_particle_hashes(Range&& states, const Hasher& spatial_hash_function_) {
+  const auto state_to_range = [&](const auto& state) { return spatial_hash_function_(state); };
+  return states | ranges::views::transform(state_to_range) | ranges::to<std::vector<std::size_t>>();
```
glpuga, line 72: state_to_hash

nahueespinosa, line 72:
```suggestion
  return states | ranges::views::transform(spatial_hash_function_) | ranges::to<std::vector<std::size_t>>();
```

meta: If this is only used once, it might be better to write it in place, it isn't super hard to understand and there will be less documentation to mantain.

beluga/include/beluga/estimation/cluster_based_estimator.hpp
```diff
@@ -0,0 +1,372 @@
...
+      Sophus::SE2d{Sophus::SO2d{-kAngularHashResolution}, Sophus::Vector2d{0.0, 0.0}},
+  };
+};
+
+template <class Mixin, class StateType, class CovarianceType>
+std::pair<StateType, CovarianceType> ClusterBasedStateEstimator<Mixin, StateType, CovarianceType>::estimate() const {
+  auto hashes = cse_detail::precalculate_particle_hashes(this->self().states(), spatial_hash_function_);
+  auto grid_cell_data = cse_detail::populate_grid_cell_data_from_particles<cse_detail::GridCellDataMap2D>(
+      this->self().states(), this->self().weights(), hashes);
+  const auto weight_cap = cse_detail::calculate_percentile_weight_threshold(grid_cell_data, 0.9);
```
nahueespinosa, line 347: `0.9` should probably be a documented parameter too.

beluga/include/beluga/estimation/cluster_based_estimator.hpp
```diff
@@ -0,0 +1,372 @@
...
+  auto weights = grid_cell_data | ranges::views::transform(extract_weight_f) | ranges::to<std::vector<double>>() |
+                 ranges::actions::sort;
+  return weights[static_cast<std::size_t>(static_cast<double>(weights.size()) * threshold)];
+}
+
+/// @brief Cap the weight of each cell in the grid cell data map to a given value.
+/// @tparam GridCellDataMapType Type of the grid cell data map.
+/// @param grid_cell_data The grid cell data map.
+/// @param weight_cap The maximum weight value to be assigned to each cell.
+template <class GridCellDataMapType>
```
nahueespinosa, line 128: meta: Maybe we can have a `GridCellData` class with member methods and the underlying map container could be a template parameter, instead of having free function templates that are really coupled with the map value type.

beluga/include/beluga/estimation/cluster_based_estimator.hpp
```diff
@@ -0,0 +1,372 @@
...
+        ranges::views::transform(get_neighbor_hash) |     //
+        ranges::views::filter(filter_invalid_neighbors);  //
+
+    for (const auto& neighbor_hash : valid_neighbor_hashes_view) {
+      auto& neighbor = grid_cell_data[neighbor_hash];
+      neighbor.cluster_id = grid_cell.cluster_id;
+      const auto inflated_priority =
+          weight_cap + neighbor.weight;  // since weights are capped at weight_cap, this gives us a value that is
+                                         // guaranteed to be higher than any other weight from a local maximum.
+      grid_cell_queue.push({inflated_priority, neighbor_hash});  // reintroduce with inflated priority
```
hidmic, line 224: @glpuga meta: to flood assign cluster ids from a given local maximum I presume. Is there a textbook / article reference for this algorithm? How does it fare against other clustering algorithms?

beluga/include/beluga/estimation/cluster_based_estimator.hpp
```diff
@@ -0,0 +1,372 @@
...
+  /// @brief Adjacent grid cells used by the clustering algorithm.
+  const std::vector<Sophus::SE2d> adjacent_grid_cells_ = {
+      Sophus::SE2d{Sophus::SO2d{0.0}, Sophus::Vector2d{+kSpatialHashResolution, 0.0}},
+      Sophus::SE2d{Sophus::SO2d{0.0}, Sophus::Vector2d{-kSpatialHashResolution, 0.0}},
+      Sophus::SE2d{Sophus::SO2d{0.0}, Sophus::Vector2d{0.0, +kSpatialHashResolution}},
+      Sophus::SE2d{Sophus::SO2d{0.0}, Sophus::Vector2d{0.0, -kSpatialHashResolution}},
+      Sophus::SE2d{Sophus::SO2d{+kAngularHashResolution}, Sophus::Vector2d{0.0, 0.0}},
+      Sophus::SE2d{Sophus::SO2d{-kAngularHashResolution}, Sophus::Vector2d{0.0, 0.0}},
+  };
+};
```
hidmic, line 340: @glpuga meta: considering that both spatial hashing / indexing and space discretization must have matching resolutions, it would be nice for local neighbor cell poses to be derived from the given spatial hashing / indexing mechanism (or otherwise deduplicated somehow).

beluga/test/beluga/estimation/test_cluster_based_estimation.cpp
```diff
@@ -0,0 +1,373 @@
...
+  // only cells beyond the 10% weight percentile to avoid the messy border
+  // between clusters beneath that threshold
+  const auto ten_percent_threshold = calculate_percentile_weight_threshold(map, 0.15);
+  map_cells_to_clusters(map, spatial_hash_function, adjacent_grid_cells, weight_cap);
+
+  auto cells_above_minimum_threshold_view =
+      coordinates | ranges::views::filter([&](const auto& c) { return std::get<2>(c) >= ten_percent_threshold; });
+
+  auto quadrant_1_view = cells_above_minimum_threshold_view | ranges::views::filter([&](const auto& c) {
+                           return std::get<0>(c) < 18.0 && std::get<1>(c) < 18.0;
```
hidmic, line 231: @glpuga nit: using structured bindings may help readability here. Also, named constants for grid width and height.

beluga/test/beluga/estimation/test_cluster_based_estimation.cpp
```diff
@@ -0,0 +1,373 @@
...
+                           return std::get<0>(c) >= 18.0 && std::get<1>(c) >= 18.0;
+                         });
+
+  const auto coord_to_hash = [&](const auto& coords) {
+    const auto& [x, y, w] = coords;
+    const auto state = SE2d{SO2d{0.}, Vector2d{x, y}};
+    return spatial_hash_function(state);
+  };
+
+  const auto hash_to_id = [&](const auto& hash) { return map[hash].cluster_id ? map[hash].cluster_id.value() : 9999; };
```
hidmic, line 249: @glpuga hmm, shouldn't all particles be assigned a cluster? If so, why not simply invoke `map[hash].cluster_id.value()` unconditionally and let the exception propagate if broken?

hidmic (commented):
Awesome work @glpuga ! I'd love to see if this effectively fixes our last pending issues.

beluga/include/beluga/estimation/cluster_based_estimator.hpp
glpuga: No, we can remove this check.

beluga/include/beluga/estimation/cluster_based_estimator.hpp
glpuga: I wrote in place, and declared for the three uses in the unit test.

beluga/include/beluga/estimation/cluster_based_estimator.hpp
glpuga: Agreed.

beluga/include/beluga/estimation.hpp
```diff
@@ -0,0 +1,49 @@
...
+ * \page StateEstimatorPage Beluga named requirements: StateEstimator
+ * The requirements that a state estimator must satisfy.
+ *
+ * \section StateEstimationRequirements Requirements
+ * `T` is a `StateEstimator` if given a (possibly const) instance `p` of `T`, the following is satisfied:
+ * - `p.estimate()` is a valid expression.
+ * - `std::get<0>(p.estimate())` is valid.
+ *   `decltype(std::get<0>(p.estimate()))` represents the type of the estimated state.
+ * - `std::get<1>(p.estimate())` is valid.
+ *   `decltype(std::get<1>(p.estimate()))` represents the type of the covariance of the estimation.
```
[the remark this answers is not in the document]
glpuga: Fixed.

beluga/include/beluga/estimation/cluster_based_estimator.hpp
```diff
@@ -0,0 +1,372 @@
...
+
+/// @brief Function that creates an vector containing the hashes of each of the states in the input range.
+/// @tparam Range Type of the states range.
+/// @tparam Hasher Hash function type to convert states into hashes.
+/// @param states The range of states.
+/// @param spatial_hash_function_ The hash object instance.
+/// @return A vector containing the hashes of each of the states in the input range.
+template <class Range, class Hasher>
+[[nodiscard]] auto precalculate_particle_hashes(Range&& states, const Hasher& spatial_hash_function_) {
+  const auto state_to_range = [&](const auto& state) { return spatial_hash_function_(state); };
```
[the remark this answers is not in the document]
glpuga: Indeed. It was only to give it a name, but it's not needed.

beluga/include/beluga/estimation/cluster_based_estimator.hpp
glpuga: > to flood assign cluster ids from a given local maximum I presume.
Indeed, with additional provisions to avoid cluster fragmentation due to pdf estimation noise.

> Is there a textbook / article reference for this algorithm?
Nope.

> How does it fare against other clustering algorithms?
God knows. Doing such a comparison would be a paper on its own we may someday write.

beluga/include/beluga/estimation/cluster_based_estimator.hpp
glpuga: I'm not sure I understand this comment.

beluga/include/beluga/estimation/cluster_based_estimator.hpp
glpuga: Good idea. Done.

beluga/include/beluga/estimation/cluster_based_estimator.hpp
glpuga: I won't be addressing this.

beluga/include/beluga/estimation/simple_state_estimator.hpp
nahueespinosa, line 1: I hesitate to merge new mixin features that aren't really going to be used. We should go in the other direction and define the cluster-based estimation algorithm as a function that takes a range of particles, which I think involves reverting some changes in this PR.

At the very least, we should remove directory location and file name changes so that this PR touches fewer files if we want to land this as is.

nahueespinosa (changes requested):

glpuga:
@nahueespinosa PTAL to see if this is closer to what you have in mind.

beluga/include/beluga/estimation.hpp
```diff
@@ -0,0 +1,52 @@
...
+#ifndef BELUGA_ESTIMATION_HPP
+#define BELUGA_ESTIMATION_HPP
+
+/**
+ * \file
+ * \brief Implementation of algorithms that allow calculating the true state estimator of a particle set.
+ */
+
+/**
+ * \page StateEstimatorPage Beluga named requirements: StateEstimator
```
nahueespinosa, line 24: I would prefer the different algorithms to be separate function overloads rather than classes:

```cpp
beluga::estimate(states);  // simple
beluga::weighted_estimate(states, weights);
beluga::weighted_estimate(particles);
beluga::cluster_based_estimate(states, weights);
beluga::cluster_based_estimate(particles);
```

beluga/include/beluga/algorithm.hpp
```diff
@@ -22,7 +22,6 @@
 
 #include <beluga/algorithm/distance_map.hpp>
 #include <beluga/algorithm/effective_sample_size.hpp>
-#include <beluga/algorithm/estimation.hpp>
```
nahueespinosa, line 25: I think the `algorithm` directory is the right place for these.

nahueespinosa (commented):
@glpuga Thanks for rebasing and updating this!

hidmic:
@glpuga can we resurrect this and merge?

hidmic (commented):
Another pass! Nice work.

beluga/include/beluga/algorithm/cluster_based_estimation.hpp
```diff
@@ -0,0 +1,422 @@
...
+ * \param map The map containing the data to be inserted into the priority queue.
+ * \param proj The projection function used to compute the priority of each element.
+ * \return A priority queue where elements are ordered by the priority computed
+ *         from the map's values using the projection function.
+ */
+template <class Map, class Proj>
+[[nodiscard]] auto make_priority_queue(const Map& map, Proj&& proj) {
+  struct KeyWithPriority {
+    double priority;  // priority value used to order the queue (higher value first).
+    std::size_t key;  // hash of the cell in the grid cell data map.
```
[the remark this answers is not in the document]
nahueespinosa: Fixed in https://github.com/Ekumen-OS/beluga/pull/275/commits/b3dd2c482fbafe348be6ae58c3cdf6f0ca792b3a.

beluga/include/beluga/algorithm/cluster_based_estimation.hpp
```diff
@@ -0,0 +1,422 @@
...
+        neighbor.cluster_id = cell.cluster_id;
+        queue.push({max_priority + neighbor.weight, neighbor_hash});  // reintroduce with inflated priority
+      }
+    }
+  }
+};
+
+/// Parameters used to construct a ParticleClusterizer instance.
+struct ParticleClusterizerParam {
+  double spatial_hash_resolution = 0.20;   ///< clustering algorithm spatial resolution
```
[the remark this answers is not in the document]
nahueespinosa: Fixed in https://github.com/Ekumen-OS/beluga/pull/275/commits/b3dd2c482fbafe348be6ae58c3cdf6f0ca792b3a.

beluga/include/beluga/algorithm/cluster_based_estimation.hpp
```diff
@@ -0,0 +1,422 @@
...
+   * \param particles A range of particles to be clustered.
+   * \return A vector of cluster IDs corresponding to the input particles.
+   */
+  template <class Particles>
+  [[nodiscard]] auto operator()(Particles&& particles) {
+    auto states = beluga::views::states(particles);
+    auto weights = beluga::views::weights(particles);
+    auto hashes = states | ranges::views::transform(spatial_hash_function_) | ranges::to<std::vector>;
+
+    auto map = ParticleClusterizerImpl::make_cluster_map(states, weights, hashes);
```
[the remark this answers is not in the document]
nahueespinosa: Replaced with a namespace in https://github.com/Ekumen-OS/beluga/pull/275/commits/b3dd2c482fbafe348be6ae58c3cdf6f0ca792b3a.

beluga/include/beluga/algorithm/cluster_based_estimation.hpp
```diff
@@ -0,0 +1,422 @@
...
+  struct Estimate {
+    Weight weight;
+    EstimateState mean;
+    EstimateCovariance covariance;
+
+    /// Constructor used with `emplace_back`.
+    constexpr Estimate(Weight w, EstimateState m, EstimateCovariance c)
+        : weight{w}, mean{std::move(m)}, covariance{std::move(c)} {}
+
+    constexpr bool operator<(const Estimate& other) const { return weight < other.weight; }
```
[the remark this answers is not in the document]
nahueespinosa: Makes sense, fixed in https://github.com/Ekumen-OS/beluga/pull/275/commits/b3dd2c482fbafe348be6ae58c3cdf6f0ca792b3a.

beluga/test/beluga/algorithm/test_cluster_based_estimation.cpp
```diff
@@ -0,0 +1,416 @@
...
+
+  [[nodiscard]] static auto
+  make_particle_multicluster_dataset(double xmin, double xmax, double ymin, double ymax, double step) {
+    std::vector<std::pair<SE2d, beluga::Weight>> particles;
+
+    const auto xwidth = xmax - xmin;
+    const auto ywidth = ymax - ymin;
+
+    // simulate particles in a grid with 4 (2x2) clusters with
+    // different peak heights. The highest on is the one located on the
```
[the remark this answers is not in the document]
nahueespinosa: Fixed in https://github.com/Ekumen-OS/beluga/pull/275/commits/b3dd2c482fbafe348be6ae58c3cdf6f0ca792b3a.

beluga/test/beluga/algorithm/test_cluster_based_estimation.cpp
```diff
@@ -0,0 +1,416 @@
...
+    // higher-right.
```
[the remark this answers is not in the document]
nahueespinosa: Fixed in https://github.com/Ekumen-OS/beluga/pull/275/commits/b3dd2c482fbafe348be6ae58c3cdf6f0ca792b3a.

beluga/include/beluga/algorithm/cluster_based_estimation.hpp
```diff
@@ -0,0 +1,422 @@
...
+    constexpr bool operator!=(const Estimate& other) const { return weight != other.weight; }
+  };
+
+  // For each cluster found, estimate the mean and covariance of the states that belong to it.
+  const auto unique_clusters = clusters | ranges::to<std::unordered_set>;
+  auto estimates = std::vector<Estimate>{};
+  estimates.reserve(unique_clusters.size());
+
+  for (const auto cluster : unique_clusters) {
+    auto particles = ranges::views::zip_with(&Particle::construct, states, weights, clusters) |          //
```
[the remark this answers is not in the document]
nahueespinosa: Yeah, this is not great. Just out of curiosity, do you think that sorting the indices and rearranging would be faster than simply sorting the container of particles after copying?

`chunk_by` was added in [`range-v3 0.12.0`](https://github.com/ericniebler/range-v3/releases/tag/0.12.0) and doesn't exist in `range-v3 0.10.0` which is distributed with Focal. There was a `group_by` view before but it had different semantics.

For reference: https://togithub.com/ericniebler/range-v3/issues/1637#issuecomment-901514674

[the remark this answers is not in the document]
nahueespinosa: Attempted to address this here: https://github.com/Ekumen-OS/beluga/pull/275/commits/054dd023f753ccf2cd1f770c2b402cbe297c0116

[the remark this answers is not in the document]
hidmic: > Just out of curiosity, do you think that sorting the indices and rearranging would be faster than simply sorting the container of particles after copying?

It depends on how the sorting algorithm works, but probably the same. It would get better if we could take a set of indices from a range without copying, but that's a stretch.

> chunk_by was added in [range-v3 0.12.0](https://github.com/ericniebler/range-v3/releases/tag/0.12.0) only available in [Noble](https://packages.ubuntu.com/source/noble/range-v3). There was a group_by view before but it had different semantics.

Oh, bummer :(

I guess we could achieve the same with an `std::unordered_map<Cluster, std::vector<std::pair<State, Weight>>>` or similar.

hidmic (approved):
Niceee

beluga/include/beluga/algorithm/cluster_based_estimation.hpp
```diff
@@ -0,0 +1,434 @@
...
+ * \param percentile The percentile (between 0 and 1) to calculate the threshold for.
+ * \return The value at the specified percentile in the sorted range.
+ */
+template <class Range>
+[[nodiscard]] auto calculate_percentile_threshold(Range&& range, double percentile) {
+  auto values = range | ranges::to<std::vector>;
+  const auto n = static_cast<std::ptrdiff_t>(static_cast<double>(values.size()) * percentile);
+  std::nth_element(values.begin(), values.begin() + n, values.end());
+  return values[static_cast<std::size_t>(n)];
+}
```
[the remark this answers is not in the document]
nahueespinosa: Done in https://github.com/Ekumen-OS/beluga/pull/275/commits/5124aa3f0e9899b1236eee619c055815e9fad835.

beluga/include/beluga/algorithm/cluster_based_estimation.hpp
```diff
@@ -0,0 +1,434 @@
...
+ * of the cluster with the highest total weight is returned. If no clusters are found,
+ * the overall mean and covariance of the particles are calculated and returned.
+ *
+ * \tparam Particles The type of the particles range.
+ * \param particles A range of particles to be clustered and estimated.
+ * \param parameters Parameters for the particle clusterizer (optional).
+ * \return A pair consisting of the state mean and covariance of the cluster with the highest total weight.
+ */
+template <class Particles>
+[[nodiscard]] auto cluster_based_estimate(Particles&& particles, ParticleClusterizerParam parameters = {}) {
```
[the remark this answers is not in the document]
nahueespinosa: Changed to take separate ranges here: https://github.com/Ekumen-OS/beluga/pull/275/commits/5124aa3f0e9899b1236eee619c055815e9fad835.

beluga/include/beluga/algorithm/cluster_based_estimation.hpp
```diff
@@ -0,0 +1,434 @@
...
+struct Cell {
+  State representative_state;             ///< state of a particle that is within the cell
+  double weight{0.0};                     ///< average weight of the cell
+  std::size_t num_particles{0};           ///< number of particles in the cell
+  std::optional<std::size_t> cluster_id;  ///< cluster id of the cell
+};
+
+/// A map that holds the sparse data about the particles grouped in cells.
+template <class State>
+using Map = std::unordered_map<std::size_t, Cell<State>>;
```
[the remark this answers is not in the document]
nahueespinosa: Done in https://github.com/Ekumen-OS/beluga/pull/275/commits/5124aa3f0e9899b1236eee619c055815e9fad835.

beluga/include/beluga/algorithm/cluster_based_estimation.hpp
```diff
@@ -0,0 +1,434 @@
...
+ * \tparam States Range type of the states.
+ * \tparam Weights Range type of the weights.
+ * \tparam Hashes Range type of the hashes.
+ * \param states Range containing the states of the particles.
+ * \param weights Range containing the weights of the particles.
+ * \param clusters Cluster ids of the particles.
+ * \return A vector of elements, containing the weight, mean and covariance of each cluster, in no particular order.
+ */
+template <class States, class Weights, class Clusters>
+[[nodiscard]] auto estimate_clusters(States&& states, Weights&& weights, Clusters&& clusters) {
```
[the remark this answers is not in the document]
nahueespinosa: So, IMO this is better than a tuple, but I agree that returning a local class is not great. I like the STL approach of incorporating return types for constrained algorithms: https://en.cppreference.com/w/cpp/algorithm/ranges#Return_types

Maybe we can do that as a follow-up?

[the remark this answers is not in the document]
hidmic: Sounds good to me.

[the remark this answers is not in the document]
hidmic: Let's add a ticket somewhere so we don't forget.

hidmic (approved):
Go go go!

beluga/include/beluga/algorithm/cluster_based_estimation.hpp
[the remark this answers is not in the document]
nahueespinosa: https://github.com/Ekumen-OS/beluga/issues/412

nahueespinosa (approved):

beluga/include/beluga/algorithm/cluster_based_estimation.hpp
```diff
@@ -0,0 +1,422 @@
...
+
+  /// Cluster weight cap threshold.
+  /**
+   * This parameters is the upper percentile threshold used to cap the accumulated
+   * weight of the cells in the grid to flatten the top of the approximated
+   * density function and make the clustering algorithm more robust to estimation noise,
+   * by fusing together any adjacent peaks whose weight is above the threshold.
+   * The range of this parameter is 0.0 to 1.0. The 0.9 default places the
+   * threshold at approximately 0.5 standard deviation from the mean, which is
+   * halfway to the inflection of the normal distribution.
```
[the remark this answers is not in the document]
nahueespinosa: Resolving this conversation to unblock merging.
@glpuga Let me know if you have thoughts about this.

PR #278, opened 2024-01-01
Add random intersperse view

nahueespinosa:
### Proposed changes

This patch adds a new random intersperse view to help with adaptive sampling and replace the `make_random_selector` implementation. This is a range adaptor object that will insert values from a generator function between contiguous elements from the source range based on a given probability.

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/views/random_intersperse.hpp
+++ b/beluga/include/beluga/views/random_intersperse.hpp
@@ -45,12 +45,10 @@ struct random_intersperse_view
     : public ranges::view_adaptor<
           random_intersperse_view<Range, Fn, URNG>,
           Range,
-          // The cardinality value can be infinite, unknown, finite or a specific number if known.
-          // If the adapted range cardinality is finite (or a specific number) then we know the
-          // resulting view is finite. Else, we propagate the infinite or unknown specification.
-          // Care must be taken when the adapted range is finite but the intersperse probability
-          // is 1.0, leading to an infinite range in practice.
-          ranges::is_finite<Range>::value ? ranges::finite : ranges::range_cardinality<Range>::value> {
+          // The cardinality value is unknown at compile time.
+          // If the adapted range cardinality is finite then we know the resulting view is finite.
+          // But the intersperse probability could be 1.0, leading to an infinite range in practice.
+          ranges::unknown> {
  public:
   /// Default constructor.
   random_intersperse_view() = default;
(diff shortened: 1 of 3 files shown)


beluga/include/beluga/views/random_intersperse.hpp
```diff
@@ -0,0 +1,185 @@
...
+template <class Range, class Fn, class URNG = typename ranges::detail::default_random_engine>
+struct random_intersperse_view
+    : public ranges::view_adaptor<
+          random_intersperse_view<Range, Fn, URNG>,
+          Range,
+          // The cardinality value can be infinite, unknown, finite or a specific number if known.
+          // If the adapted range cardinality is finite (or a specific number) then we know the
+          // resulting view is finite. Else, we propagate the infinite or unknown specification.
+          // Care must be taken when the adapted range is finite but the intersperse probability
+          // is 1.0, leading to an infinite range in practice.
```
hidmic, line 52: @nahueespinosa is it accounted for? It doesn't seem like it. Shouldn't the view constructor `throw` or `assert` in that case?

beluga/test/beluga/views/test_random_intersperse.cpp
```diff
@@ -0,0 +1,142 @@
...
+  static_assert(!ranges::random_access_range<decltype(input)>);
+  static_assert(!ranges::random_access_range<decltype(output)>);
+
+  static_assert(!ranges::contiguous_range<decltype(input)>);
+  static_assert(!ranges::contiguous_range<decltype(output)>);
+}
+
+TEST(RandomIntersperseView, GuaranteedIntersperseFirstElement) {
+  auto input = std::array{10, 20, 30};
+  auto output = input | beluga::views::random_intersperse([i = 0]() mutable { return i++; }, 1.0);
```
hidmic, line 81: @nahueespinosa nit: consider using a named constant for the intersperse probability, here and elsewhere.

hidmic (commented):
Neat.

beluga/include/beluga/views/random_intersperse.hpp
nahueespinosa: I didn't want to limit the probability values, I use 1.0 in tests too. So I simply specified an `unkown` cardinality here https://github.com/Ekumen-OS/beluga/pull/278/commits/63dffa9881a85f94786717f73509a35b5c15a302.

beluga/test/beluga/views/test_random_intersperse.cpp
nahueespinosa: Done in https://github.com/Ekumen-OS/beluga/pull/278/commits/ec93c23f1f4c3c0ee429bfe383df45508cf744ea.

hidmic (approved):

Issue #279, opened 2024-01-02
Improve the library user experience


nahueespinosa:
### Feature description

The current user experience for the beluga library is not ideal. Mixins are difficult to understand, use and customize. Complex template instantiation and magic functions (`make_mixin`) are needed to make runtime selection possible. I'd argue that we haven't spent time on user-facing documentation since it's really difficult to come up with a simple example.

I strongly believe this library would be much better if our first class citizens were ranges and range algorithms instead of mixins. They can be composed and customized by design, and we wouldn't need to document basic concepts from scratch, since ranges are part of STL since C++20. This approach would allow us to constraint our algorithms better and even optimize them for particular cases.

As a potential library user, this is the code I would want to be shown first:

```cpp
auto particles = beluga::TupleContainer<int, beluga::Weight>{};
auto motion_model = beluga::DifferentialDrive{...};
auto sensor_model = beluga::LikelihoodField{...};

motion_model.update(...);
sensor_model.update(...);

motion_model.propagate(particles);
sensor_model.reweight(particles);

auto temp_particles = particles | 
                      beluga::views::sample |
                      beluga::views::take_exactly(1000) |
                      ranges::to<TupleContainer>;

std::swap(particles, temp_particles);

auto [mean, covariance] = beluga::estimate(particles);
```

Want to insert random particles while sampling?

```cpp
auto particles = beluga::TupleContainer<int, beluga::Weight>{};
auto motion_model = beluga::DifferentialDrive{...};
auto sensor_model = beluga::LikelihoodField{...};

motion_model.update(...);
sensor_model.update(...);

motion_model.propagate(particles);
sensor_model.reweight(particles);

const auto make_random_state = [&]() { return sensor_model.make_random_state(); };
const auto random_state_probability = ...;

auto temp_particles = particles | 
                      beluga::views::sample |
                      beluga::views::random_intersperse(make_random_state, random_state_probability) |
                      beluga::views::take_exactly(1000) |
                      ranges::to<TupleContainer>;

std::swap(particles, temp_particles);

auto [mean, covariance] = beluga::estimate(particles);
```

Showcasing more complex functionality:

```cpp
auto particles = beluga::TupleContainer<int, beluga::Weight>{};

// Type-erased models
beluga::AnyMotionModel motion_model = beluga::DifferentialDrive{...};
beluga::AnySensorModel sensor_model = beluga::LikelihoodField{...};

motion_model.update(...);
sensor_model.update(...);

[...]

PR #281, opened 2024-01-06
Add sample view

nahueespinosa:
### Proposed changes

This adds a (with-replacement) sample view that generates an infinite range compatible with particle filter algorithms.

Related to #279.

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/views/sample.hpp
+++ b/beluga/include/beluga/views/sample.hpp
@@ -20,9 +20,10 @@
 #include <range/v3/utility/random.hpp>
 #include <range/v3/view/common.hpp>
 
-#include <beluga/definitions.hpp>
 #include <beluga/type_traits/particle_traits.hpp>
 #include <beluga/views/particles.hpp>
+
+#include <beluga/detail/prologue.hpp>
 
 /**
  * \file
@@ -203,5 +204,7 @@ inline constexpr bool enable_borrowed_range<beluga::views::detail::sample_view<R
 /// \endcond
 
 }  // namespace ranges
+
+#include <beluga/detail/epilogue.hpp>
 
 #endif
--- a/beluga/include/beluga/views/sample.hpp
+++ b/beluga/include/beluga/views/sample.hpp
@@ -166,10 +166,8 @@ struct sample_fn {
   }
 
   /// Overload that returns a view closure to compose with other views.
-  template <
-      class URNG = typename ranges::detail::default_random_engine,
-      std::enable_if_t<!ranges::range<URNG>, int> = 0>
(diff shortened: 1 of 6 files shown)


beluga/include/beluga/views/sample.hpp
```diff
@@ -0,0 +1,210 @@
...
+  template <class URNG, std::enable_if_t<!ranges::range<URNG>, int> = 0>
+  constexpr auto operator()(URNG& engine) const {
+    return ranges::make_view_closure(ranges::bind_back(sample_fn{}, std::ref(engine)));
+  }
+};
+
+}  // namespace detail
+
+/// [Range adaptor object](https://en.cppreference.com/w/cpp/named_req/RangeAdaptorObject) that
+/// will ramdomly sample with replacement from an input range.
```
hidmic, line 179: @nahueespinosa typo: 
```suggestion
/// will randomly sample with replacement from an input range.
```

beluga/include/beluga/views/sample.hpp
```diff
@@ -0,0 +1,210 @@
...
+  [[nodiscard]] constexpr auto begin_cursor() { return cursor{distribution_, engine_, ranges::begin(range_)}; }
+
+  /// Return an unreachable sentinel since this is an infinite range.
+  [[nodiscard]] constexpr auto end_cursor() const noexcept { return ranges::unreachable_sentinel_t{}; }
+
+  Range range_;
+
+  // Using a shared resource as the current version of range-v3 does not support move-only views.
+  // This is to ensure the O(1) (cheaply) copyable semantic requirement for views.
+  std::shared_ptr<Distribution> distribution_;
```
hidmic, line 111: @nahueespinosa thought: we could escape dynamic allocation here by relying on the fact all STL distributions support near copy construction via parameter set getters and the corresponding constructor (see `param` methods). Unclear what the complexity of distribution instantiation is.

hidmic (approved):

beluga/include/beluga/views/sample.hpp
nahueespinosa: What I'm trying to avoid here is copying the vector of weights that `discrete_distribution` parameters hold:
https://github.com/gcc-mirror/gcc/blob/cddaff46adfd831d4836f29dc5e21f73eaed37b3/libstdc%2B%2B-v3/include/bits/random.h#L5499

hidmic: Yeah, I know that's the downside, and in this case, you'll heap allocate either way. My point is, near copy constructing distributions leaves the door open to stack allocated (and real-time friendly) implementations.

Anyways, I'm over-optimizing. Nevermind.

PR #282, opened 2024-01-07
Improve take_while_kld implementation

nahueespinosa:
### Proposed changes

Improves the take_while_kld view by:
- Adding overloads to handle particle ranges directly.
- Removing overloads that take hash ranges.
- Adding concept checks.

Related to #279.

#### Type of change

### Checklist


beluga/include/beluga/views/forward.hpp
```diff
@@ -0,0 +1,96 @@
...
+ * Restricts a range to model the forward range concept or lower.
+ */
+template <class Range>
+struct forward_view : public ranges::view_adaptor<forward_view<Range>, Range, ranges::range_cardinality<Range>::value> {
+ public:
+  /// Default constructor.
+  forward_view() = default;
+
+  /// Construct the view from an existing range.
+  constexpr explicit forward_view(Range range) : forward_view::view_adaptor{std::move(range)} {}
```
nahueespinosa, line 42: Codecov says that this line is not covered, probably because it is optimized away. :shrug:

https://app.codecov.io/gh/Ekumen-OS/beluga/pull/282/blob/beluga/include/beluga/views/forward.hpp#L42

beluga/test/beluga/views/test_forward.cpp
```diff
@@ -0,0 +1,101 @@
...
+  auto output = input | beluga::views::forward;
+
+  static_assert(!ranges::common_range<decltype(input)>);
+  static_assert(!ranges::common_range<decltype(output)>);
+
+  static_assert(ranges::viewable_range<decltype(input)>);
+  static_assert(ranges::viewable_range<decltype(output)>);
+
+  static_assert(!ranges::forward_range<decltype(input)>);
+  static_assert(!ranges::forward_range<decltype(output)>);
```
hidmic, line 63: @nahueespinosa meta: not sure I see why an infinite range wouldn't be a forward range (conceptually, regardless of whether rangesv3 allows it or not).

beluga/include/beluga/views/take_while_kld.hpp
```diff
@@ -95,58 +97,103 @@ namespace detail {
...
-    static_assert(ranges::input_range<States>);
+    static_assert(ranges::input_range<Range>);
     static_assert(ranges::input_range<Hashes>);
     static_assert(std::is_convertible_v<ranges::range_value_t<Hashes>, std::size_t>);
     const auto hash = [](const auto& p) { return std::get<1>(p); };
-    return ranges::views::zip(ranges::views::all(states), ranges::views::all(hashes)) |  //
-           ranges::views::take_while(beluga::kld_condition(min, epsilon, z), hash) |     //
-           ranges::views::take(max) |                                                    //
+    return ranges::views::zip(ranges::views::all(range), ranges::views::all(hashes)) |  //
+           beluga::views::forward |                                                     //
```
hidmic, line 126: @nahueespinosa IIUC we do this to constrain `take_while_kld` output to be a forward range (as `take_while` seems to mimic the given range traits). Is that true? If so, why? Order is arbitrary and orthogonal to KLD conditions.

hidmic (approved):

beluga/test/beluga/views/test_forward.cpp
nahueespinosa: A forward range allows to execute multi-pass algorithms and guarantees that two iterators to the same range can be compared against each other. That is not true for the `generate` view.

beluga/include/beluga/views/take_while_kld.hpp
nahueespinosa: Yes, that's true. Hmm, now that you say it, I think you're right. I'll remove the forward_view.

hidmic (approved):

beluga/include/beluga/views/take_while_kld.hpp
```diff
@@ -95,59 +94,45 @@ namespace detail {
...
+        // If the above is not possible, assume this is a particle range and invoke
+        // the hasher with the state element of each particle.
+        static_assert(is_particle_range_v<Range>);
+        static_assert(std::is_invocable_r_v<std::size_t, Hasher, beluga::state_t<ranges::range_value_t<Range>>>);
+        return ranges::compose(std::move(hasher), beluga::state);
+      }
+    }();
+
+    return ranges::views::all(std::forward<Range>(range)) |                                      //
+           ranges::views::take_while(beluga::kld_condition(min, epsilon, z), std::move(proj)) |  //
```
hidmic, line 134: @nahueespinosa aha! good catch.

hidmic (approved):

PR #285, opened 2024-01-09
Add support for tuple vector assignment and conversion

nahueespinosa:
### Proposed changes

Related to #279.

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/tuple_vector.hpp
+++ b/beluga/include/beluga/tuple_vector.hpp
@@ -1,4 +1,4 @@
-// Copyright 2022-2023 Ekumen, Inc.
+// Copyright 2022-2024 Ekumen, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


beluga/include/beluga/tuple_vector.hpp
```diff
@@ -123,13 +200,13 @@ class TupleContainer {
...
-  constexpr void push_back(value_type&& value) {
+  constexpr void push_back(value_type value) {
     push_back_impl(std::move(value), std::make_index_sequence<sizeof...(Types)>());
   }
 
   /// \overload
-  constexpr void push_back(const value_type& value) {
-    push_back_impl(value, std::make_index_sequence<sizeof...(Types)>());
+  constexpr void push_back(reference_type value) {
+    push_back_impl(std::move(value), std::make_index_sequence<sizeof...(Types)>());
```
hidmic, line 209: @nahueespinosa hmm, moving a mutable reference argument like this is unusual, even if a temporary tuple of references. Does it really make a difference?

hidmic (approved):
Simply flawless :muscle:

beluga/include/beluga/tuple_vector.hpp
nahueespinosa: I don't think it makes a difference. Let's stick to the principle of least surprise. Fixed in https://github.com/Ekumen-OS/beluga/pull/285/commits/854f94bf4fa288867385c0244879663177cb2039.

PR #287, opened 2024-01-12
Add assign action

nahueespinosa:
### Proposed changes

View-action composition in the range library is limited. This patch implements a new range adaptor object that can be appended to any view closure to evaluate it eagerly and assign the result to a given range, effectively converting any view into an action.

Related to #279.

The simple particle filter pipeline example becomes:

```cpp
struct Particle {
  int state;
  double weight;
};

auto particles = std::vector<Particle>{};
auto motion_model = ...;
auto sensor_model = ...;

particles |= beluga::actions::propagate(motion_model) |
             beluga::actions::reweight(sensor_model) |
             beluga::views::sample |
             ranges::views::take_exactly(1000) |
             beluga::actions::assign;
```

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/actions/assign.hpp
+++ b/beluga/include/beluga/actions/assign.hpp
@@ -1,0 +1,116 @@
+// Copyright 2024 Ekumen, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef BELUGA_ACTIONS_ASSIGN_HPP
+#define BELUGA_ACTIONS_ASSIGN_HPP
+
+#include <range/v3/action/action.hpp>
+#include <range/v3/functional/bind_back.hpp>
+#include <range/v3/range/conversion.hpp>
+#include <range/v3/view/view.hpp>
+
+namespace beluga::actions {
+
+/// \cond
+
+template <class T, class = void>
+struct is_range_closure : std::false_type {};
+
+template <class T>
+struct is_range_closure<ranges::actions::action_closure<T>> : std::true_type {};
+
+template <class T>
+struct is_range_closure<ranges::views::view_closure<T>> : std::true_type {};
+
+template <class T>
+inline constexpr bool is_range_closure_v = is_range_closure<T>::value;
+
+/// \endcond
+
+namespace detail {
+
+/// Implementation detail for an assign range adaptor object.
+struct assign_fn {
+  /// Overload that implements the assign algorithm.
+  template <
+      class Range,
+      class Fn,
+      std::enable_if_t<ranges::range<Range>, int> = 0,
+      std::enable_if_t<is_range_closure_v<Fn>, int> = 0>
+  constexpr auto operator()(Range& range, Fn fn) const -> Range& {
+    auto&& view = fn(range);
+    if constexpr (!std::is_same_v<Range, std::decay_t<decltype(view)>>) {
+      // If the result of invoking the closure is not the range itself,
+      // then we need to convert the view and assign it to the input range.
+      range = std::move(view) | ranges::to<Range>;
+    }
+    return range;
+  }
+
+  /// Hidden friend operator overload that enables action / view composition.
+  /**
+   * Enables the following expressions:
+   *   1) view_closure | assign
+   *   2) action_closure | view_closure | assign
+   *   3) action_closure | assign
+   *
+   * 1) Will create an action closure that can eagerly invoke `view_closure` and assign
+   *    the resulting range to the input range.
+   * 2) Will create a new action closure that can invoke `action_closure`, eagerly invoke
+   *    `view_closure`, and assign the resulting range to the input range.
+   * 3) Technically a no-op, it will just invoke the `action_closure`.
+   */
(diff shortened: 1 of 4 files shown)


beluga/include/beluga/actions/assign.hpp
```diff
@@ -0,0 +1,116 @@
...
+      class Range,
+      class Fn,
+      std::enable_if_t<ranges::range<Range>, int> = 0,
+      std::enable_if_t<is_range_closure_v<Fn>, int> = 0>
+  constexpr auto operator()(Range& range, Fn fn) const -> Range& {
+    auto&& view = fn(range);
+    if constexpr (!std::is_same_v<Range, std::decay_t<decltype(view)>>) {
+      // If the result of invoking the closure is not the range itself,
+      // then we need to convert the view and assign it to the input range.
+      range = std::move(view) | ranges::to<Range>;
```
hidmic, line 56: @nahueespinosa nit: does an rvalue `view` force move-construction of `Range` elements? Or do we need [`ranges::views::move`](https://ericniebler.github.io/range-v3/structranges_1_1views_1_1move__fn.html)?

beluga/test/beluga/actions/test_assign.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+TEST(AssignAction, ViewToActionCall) {
+  auto input = std::vector{1, 2, 3};
+  beluga::actions::assign(input, ranges::views::reverse);
+  ASSERT_TRUE(ranges::equal(input, std::vector{3, 2, 1}));
+}
+
+TEST(AssignAction, ViewToActionOperator) {
+  auto input = std::vector{1, 2, 3};
+  input <<= ranges::views::reverse;
```
hidmic, line 51: @nahueespinosa meta: I'll use the freedom of knowing you really know your sh*t to complain a bit :smiley: 

It is not that obvious what this does on first pass. It is also somewhat limiting for this to always be an in-place operation. Can we just switch to the stream operator? Then this becomes `input << input | ranges::views::reverse;` (verbose but clearer IMO) or even `output << input | ranges::views::reverse;` if we don't want that in-place operation. I acknowledge I may be leaving the scope of range action semantics here.

hidmic (commented):
Awesome!

beluga/include/beluga/actions/assign.hpp
nahueespinosa: That's a great question, here we move the view itself, not the elements. I think copying the elements by default is the sensible choice, since we don't know if the view will require access to the same element more than once. Our own sample view would be one such case. This is also what [`std::ranges::to`](https://en.cppreference.com/w/cpp/ranges/to#Notes) does by default too.

I added documentation explaining this and a test showing how to move in the elements in https://github.com/Ekumen-OS/beluga/pull/287/commits/905fcd01f634bdbc0b5c2d9658f7fc85b7d84f73.

beluga/test/beluga/actions/test_assign.cpp
nahueespinosa: > It is not that obvious what this does on first pass.

I think this is key, I wasn't sure whether to include the operator overloading or not. Removed in https://github.com/Ekumen-OS/beluga/pull/287/commits/1c5eefd0fcfe924c61fa139bbbd0a1c4f9ce9130.

Actions are kind of always in-place operations. The way of assigning to a different range would be:

```cpp
Container output;
output = input | ranges::to<Container>;
```

beluga/include/beluga/actions/assign.hpp
```diff
@@ -0,0 +1,98 @@
...
+/// can converts a view closure into an action closure.
+/**
+ * This can be appended to any view closure to evaluate it eagerly and assign the result
+ * to a given range, effectively converting any view into an action.
+ *
+ * The assignment of elements into the container may involve copy which can be less efficient than
+ * move because lvalue references are produced during the indirection call. Users can opt-in to use
+ * `ranges::views::move` to adapt the range in order for their elements to always produce an rvalue
+ * reference during the indirection call which implies move. In this case, make sure that the input
+ * views don't require accessing the elements in the adapted range more than once.
```
hidmic, line 92: :+1:

beluga/test/beluga/actions/test_assign.cpp
```diff
@@ -0,0 +1,90 @@
...
+  ASSERT_TRUE(ranges::equal(input, std::vector{3, 2}));
+}
+
+TEST(AssignAction, ActionViewActionComposition) {
+  auto input = std::vector{1, 2, 3};
+  input |= ranges::actions::drop(1) |                                      //
+           ranges::views::reverse |                                        //
+           beluga::actions::assign |                                       //
+           ranges::actions::drop(1) |                                      //
+           ranges::views::transform([](auto value) { return ++value; }) |  //
```
hidmic, line 79: @nahueespinosa nit^N: consider pushing the value outside the initial value set (`{1, 2, 3}`) to ensure this is doing what you think it's doing.

hidmic (approved):
LGTM

beluga/test/beluga/actions/test_assign.cpp
hidmic: > Actions are kind of always in-place operations.

Had the impression, yeah. FWIW `ranges::to` is (intentionally?) more explicit than it needs to be.

nahueespinosa: This paper discusses the details: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1206r1.pdf

Yeah, it seems to be intentional, adding range constructors was the original proposal for view materialization but `std::initializer_list` made it unworkable. :smile:  

```cpp
std::vector<int> foo = ....;
std::vector a{foo}; //constructs a std:vector<std::vector<int»
std::vector b(foo); //would construct a std::vector<int>
```

beluga/test/beluga/actions/test_assign.cpp
nahueespinosa: Done in https://github.com/Ekumen-OS/beluga/pull/287/commits/27860f8fe7bb8772b8dc9b5a9f562d987387c734!

PR #288, opened 2024-01-14
Fix sample view

nahueespinosa:
### Proposed changes

Make sure to set weight equal to 1 after sampling from a particle set.

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/tuple_vector.hpp
+++ b/beluga/include/beluga/tuple_vector.hpp
@@ -203,7 +203,7 @@ class TupleContainer<InternalContainer, std::tuple<Types...>> {
   }
 
   [[nodiscard]] constexpr auto all() const {
-    return std::apply([](const auto&... containers) { return beluga::views::zip(containers...); }, sequences_);
+    return std::apply([](auto&... containers) { return beluga::views::zip(containers...); }, sequences_);
   }
 
   [[nodiscard]] constexpr auto all() {
--- a/beluga/include/beluga/tuple_vector.hpp
+++ b/beluga/include/beluga/tuple_vector.hpp
@@ -203,19 +203,11 @@ class TupleContainer<InternalContainer, std::tuple<Types...>> {
   }
 
   [[nodiscard]] constexpr auto all() const {
-    return std::apply(
-        [](auto&... containers) {  //
-          return beluga::views::zip(containers...) | ranges::views::const_;
-        },
-        sequences_);
+    return std::apply([](const auto&... containers) { return beluga::views::zip(containers...); }, sequences_);
   }
 
   [[nodiscard]] constexpr auto all() {
-    return std::apply(
-        [](auto&... containers) {  //
-          return beluga::views::zip(containers...);
-        },
-        sequences_);
+    return std::apply([](auto&... containers) { return beluga::views::zip(containers...); }, sequences_);
   }
 };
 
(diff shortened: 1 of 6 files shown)


beluga/include/beluga/views/sample.hpp
```diff
@@ -183,6 +188,10 @@ struct sample_fn {
  * To use this, the input range must model the
  * [random_access_range](https://en.cppreference.com/w/cpp/ranges/random_access_range)
  * and [sized_range](https://en.cppreference.com/w/cpp/ranges/sized_range) concepts.
+ *
+ * This view implements multinomial resampling for a given range of particles. The core idea
+ * is to generate independently N random numbers and use them to select particles from the
+ * input range.
```
hidmic, line 194: @nahueespinosa meta: independent as in statistically independent :thinking:? Isn't it the same random variable $I \sim Mn(p_1, ..., p_k)$ of indices / iterators to the range? Perhaps you meant: 
```suggestion
 * This view implements multinomial resampling for a given range of particles.
 * The core idea is to draw random indices / iterators to the input particle range
 * from a [multinomial distribution](https://en.wikipedia.org/wiki/Multinomial_distribution)
 * parameterized after particle weights (and assumed uniform for non-weighted particle ranges).
```

beluga/test/beluga/test_tuple_vector.cpp
```diff
@@ -177,6 +179,20 @@ TEST(TupleVectorTest, TraitConsistency) {
...
+                ranges::range_value_t<ConstContainer>,  //
+                std::tuple<float, int>>);
+  static_assert(std::is_same_v<
+                ranges::range_reference_t<ConstContainer>,  //
+                ranges::common_tuple<const float&, const int&>>);
+  static_assert(std::is_same_v<
+                ranges::range_rvalue_reference_t<ConstContainer>,  //
+                ranges::common_tuple<const float&&, const int&&>>);
+
+  // Sadness... :(
```
hidmic, line 192: @nahueespinosa why so sad?

beluga/include/beluga/tuple_vector.hpp
```diff
@@ -203,19 +203,11 @@ class TupleContainer<InternalContainer, std::tuple<Types...>> {
   }
 
   [[nodiscard]] constexpr auto all() const {
-    return std::apply(
-        [](auto&... containers) {  //
```
hidmic, line 207: @nahueespinosa meta: now I wonder how this managed to compile. Did the implicit constness of `sequences_` not propagate to its elements?

hidmic (approved):
LGTM

beluga/include/beluga/views/sample.hpp
nahueespinosa: > independent as in statistically independent 🤔?

Independent compared to other resampling methods like systematic resampling or residual resampling.
https://bisite.usal.es/archivos/resampling_methods_for_particle_filtering_classification_implementation_and_strategies.pdf

But your suggestion looks great!

beluga/test/beluga/test_tuple_vector.cpp
nahueespinosa: I expected that the value type of the const view would be the same as the value type of the adapted container. But that's not the case. Added more context in https://github.com/Ekumen-OS/beluga/pull/288/commits/776e2b599b61775f267322fe909a571fc5932561.

beluga/include/beluga/tuple_vector.hpp
nahueespinosa: It did, the `const` I added here does nothing, but I thought it would make sense to add it. Now that you say it, I think I better just remove it. Done in https://github.com/Ekumen-OS/beluga/pull/288/commits/9d114352e16cb0cd14f1124c758e16e42c7b78bd.

PR #289, opened 2024-01-14
Add propagate and reweight actions

nahueespinosa:
### Proposed changes

Related to #279.

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/actions/propagate.hpp
+++ b/beluga/include/beluga/actions/propagate.hpp
@@ -1,0 +1,113 @@
+// Copyright 2024 Ekumen, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef BELUGA_ACTIONS_PROPAGATE_HPP
+#define BELUGA_ACTIONS_PROPAGATE_HPP
+
+#include <algorithm>
+#include <execution>
+
+#include <beluga/type_traits/particle_traits.hpp>
+#include <beluga/views/particles.hpp>
+
+#include <range/v3/action/action.hpp>
+#include <range/v3/view/common.hpp>
+
+namespace beluga::actions {
+
+namespace detail {
+
+/// Implementation detail for a propagate range adaptor object.
+struct propagate_base_fn {
+  /// Overload that implements the propagate algorithm.
+  /**
+   * \tparam ExecutionPolicy An [execution policy](https://en.cppreference.com/w/cpp/algorithm/execution_policy_tag_t).
+   * \tparam Range An [input range](https://en.cppreference.com/w/cpp/ranges/input_range) of particles.
+   * \tparam Model A callable that can compute the new state from the previous state.
+   * \param policy The execution policy to use.
+   * \param range An existing range of particles to apply this action to.
+   * \param model A callable instance to compute the states from the previous states.
+   */
+  template <
+      class ExecutionPolicy,
+      class Range,
+      class Model,
+      std::enable_if_t<std::is_execution_policy_v<std::decay_t<ExecutionPolicy>>, int> = 0,
+      std::enable_if_t<ranges::range<Range>, int> = 0>
+  constexpr auto operator()(ExecutionPolicy&& policy, Range& range, Model model) const -> Range& {
+    static_assert(beluga::is_particle_range_v<Range>);
+    auto states = range | beluga::views::states | ranges::views::common;
+    std::transform(
+        policy,              // rvalue policies are not supported in some STL implementations
+        std::begin(states),  //
+        std::end(states),    //
+        std::begin(states),  //
+        std::move(model));
+    return range;
+  }
+
+  /// Overload that re-orders arguments from a view closure.
+  template <
+      class Range,
+      class Model,
+      class ExecutionPolicy,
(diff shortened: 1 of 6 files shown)


beluga/include/beluga/actions/propagate.hpp
```diff
@@ -0,0 +1,113 @@
...
+  }
+};
+
+}  // namespace detail
+
+/// [Range adaptor object](https://en.cppreference.com/w/cpp/named_req/RangeAdaptorObject) that
+/// can update the state in a particle range using a motion model.
+/**
+ * This step generates a hyphotetical state based on the current particle state and controls,
+ * while leaving its importance weight unchanged.
```
hidmic, line 107: @nahueespinosa nit: perhaps consider:
```suggestion
 * This action updates particle states based on their current value and a state-transition function (or motion model).
 * Every other particle attribute (such as importance sampling weights) is left unchanged.
```

beluga/include/beluga/actions/reweight.hpp
```diff
@@ -0,0 +1,113 @@
...
+  }
+};
+
+}  // namespace detail
+
+/// [Range adaptor object](https://en.cppreference.com/w/cpp/named_req/RangeAdaptorObject) that
+/// can update the weights in a particle range using a sensor model.
+/**
+ * This step computes the importance factor or weight of each particle to incorporate measurements.
+ * The importance is proportional to the probability of seeing the measurement given the current particle state.
```
hidmic, line 107: @nahueespinosa nit: perhaps consider:
```suggestion
 * This action updates particle weights by importance weight multiplication. 
 * These importance weights are computed by a given measurement likelihood 
 * function (or sensor model) for current particle states.
```

hidmic (commented):
@nahueespinosa how do you imagine motion and sensor models will look like moving forward? Free functions for the simplest models? Stateful functors for the complex ones? I see pass-by-value semantics followed by move-construction. How will that affect usage for models with state?

nahueespinosa:
@hidmic In general I wouldn't divorce behavior and state unless we need to. I think we should be able to use stateful functors. But here I had a couple of potential implementation issues and I admittedly went for the lazy route.

The arguments of [`bind_back`](https://en.cppreference.com/w/cpp/utility/functional/bind_front#Notes) are always copied or moved, and must be wrapped in `std::ref` or `std::cref` if passing by reference is needed. That means there is no direct way to perfect-forward them using action closures.

In the overloads that create closures, if we're going to take a `const Model&` for instance, we should also overload `Model&&` (disabling lvalue reference binding) to make sure we make a copy and don't end up with a dangling reference to a temporary when we create our action closure in one line, and use it in another. Should we also consider mutable references? Constexpr ifs may be a way out of this madness.

That, plus the fact that I had to re-order the arguments to allow for the execution policy to go first, made me think I would need an huge amount of code to do this properly. And I couldn't think of a quick way of reducing duplication.

Before trying to implement this the hard way, I decided to give it a try to what the standard library does. It requires that function objects you pass to its algorithms be copyable. C++11 §25.1/10:

> [ Note: Unless otherwise specified, algorithms that take function objects as arguments are permitted to copy those function objects freely. Programmers for whom object identity is important should consider using a wrapper class that points to a noncopied implementation object such as reference_wrapper<T> (20.8.3), or some equivalent solution. —end note ]


That means that if we have a stateful (and expensive to copy) functor, we should use the following expression:
```cpp
input |= beluga::actions::reweight(std::cref(sensor_model));
```

I know that the standard decision is from a time where perfect forwarding didn't exist and places the burden on the user, so I'm open to try to do it the hard way if you think it's worth it. Maybe it won't be as difficult as I think it will be.

nahueespinosa:
I have an idea... Will come back later.

nahueespinosa:
I was drowning in a glass of water. Stateful functors are now supported. :sweat_smile:

beluga/test/beluga/actions/test_reweight.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+TEST(ReweightAction, Composition) {
+  auto input = std::vector{std::make_tuple(1, beluga::Weight(1.0))};
+  input |= beluga::actions::reweight([](auto) { return 4; }) |  //
+           beluga::views::sample |                              //
+           ranges::views::take_exactly(5) |                     //
+           beluga::actions::assign;
+  for (auto [value, weight] : input) {
+    std::cout << value << ", " << weight << std::endl;
```
hidmic, line 51: @nahueespinosa nit: a debugging leftover?

beluga/test/beluga/actions/test_propagate.cpp
```diff
@@ -0,0 +1,64 @@
...
+  ASSERT_TRUE(ranges::equal(beluga::views::states(input), std::vector{4, 4, 4, 4, 4}));
+}
+
+TEST(PropagateAction, StatefulModel) {
+  auto input = std::vector{std::make_tuple(5, beluga::Weight(1.0))};
+  auto model = [value = 0](int) mutable { return value++; };
+  input |= beluga::views::sample |           //
+           ranges::views::take_exactly(5) |  //
+           beluga::actions::assign |         //
+           beluga::actions::propagate(model);
```
hidmic, line 59: @nahueespinosa meta: you are going to kill me, but I honestly wonder if it wouldn't be best to simply force the user to use `std::ref` here. I mean, it is not clear from reading this that `model` is getting "borrowed" by `beluga::actions::propagate` and that it must outlive the action closure. If the action closure was to be returned from some function, omitting `std::move` would result in a dangling reference. WDYT?

nahueespinosa: Makes sense haha, will revert back to the original approach.

beluga/test/beluga/actions/test_reweight.cpp
nahueespinosa: Removed.

beluga/test/beluga/actions/test_propagate.cpp
nahueespinosa: Done.

hidmic (approved):
LGTM

PR #291, opened 2024-01-17
Motion and sensor models are not mixins

nahueespinosa:
### Proposed changes

Big patch to make motion and sensor models independent classes. It also removes utilities that are no longer needed.

Related to #279.

#### Type of change

💥 **Breaking change!** This removes some of the runtime support utilities we added for mixins. Please note that statically defined mixins are still fully supported (and will be until the migration is complete) and runtime dispatch through virtual interfaces is still there as well.

### Checklist

### DIFF
--- a/beluga_amcl/src/amcl_node.cpp
+++ b/beluga_amcl/src/amcl_node.cpp
@@ -715,7 +715,7 @@ std::unique_ptr<LaserLocalizationInterface2d> AmclNode::make_particle_filter(
       params.z_hit = get_parameter("z_hit").as_double();
       params.z_random = get_parameter("z_rand").as_double();
       params.sigma_hit = get_parameter("sigma_hit").as_double();
-      return beluga::LikelihoodFieldModel<beluga_ros::OccupancyGrid>{params, beluga_ros::OccupancyGrid{map}};
+      return beluga::LikelihoodFieldModel{params, beluga_ros::OccupancyGrid{map}};
     }
     if (name == kBeamSensorModelName) {
       auto params = beluga::BeamModelParam{};
@@ -726,7 +726,7 @@ std::unique_ptr<LaserLocalizationInterface2d> AmclNode::make_particle_filter(
       params.sigma_hit = get_parameter("sigma_hit").as_double();
       params.lambda_short = get_parameter("lambda_short").as_double();
       params.beam_max_range = get_parameter("laser_max_range").as_double();
-      return beluga::BeamSensorModel<beluga_ros::OccupancyGrid>{params, beluga_ros::OccupancyGrid{map}};
+      return beluga::BeamSensorModel{params, beluga_ros::OccupancyGrid{map}};
     }
     throw std::invalid_argument(std::string("Invalid sensor model: ") + std::string(name));
   };
--- a/beluga_amcl/src/amcl_nodelet.cpp
+++ b/beluga_amcl/src/amcl_nodelet.cpp
@@ -226,7 +226,7 @@ std::unique_ptr<LaserLocalizationInterface2d> AmclNodelet::make_particle_filter(
       params.z_hit = config_.laser_z_hit;
       params.z_random = config_.laser_z_rand;
       params.sigma_hit = config_.laser_sigma_hit;
-      return beluga::LikelihoodFieldModel<beluga_ros::OccupancyGrid>{params, beluga_ros::OccupancyGrid{map}};
+      return beluga::LikelihoodFieldModel{params, beluga_ros::OccupancyGrid{map}};
     }
     if (name == kBeamSensorModelName) {
       auto params = beluga::BeamModelParam{};
@@ -237,7 +237,7 @@ std::unique_ptr<LaserLocalizationInterface2d> AmclNodelet::make_particle_filter(
       params.sigma_hit = config_.laser_sigma_hit;
       params.lambda_short = config_.laser_lambda_short;
       params.beam_max_range = config_.laser_max_range;
-      return beluga::BeamSensorModel<beluga_ros::OccupancyGrid>{params, beluga_ros::OccupancyGrid{map}};
+      return beluga::BeamSensorModel{params, beluga_ros::OccupancyGrid{map}};
     }
     throw std::invalid_argument(std::string("Invalid sensor model: ") + std::string(name));
   };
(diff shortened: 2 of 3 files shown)


beluga/include/beluga/localization.hpp
```diff
@@ -70,30 +83,60 @@ class NullMixin : public Mixin {
...
+/// Wrapper to convert a motion model class into a mixin.
+template <typename Mixin, typename Model>
+struct MotionMixin : public Mixin, public Model {
+  using update_type = typename Model::update_type;
+
+  template <typename... Args>
+  constexpr explicit MotionMixin(Model model, Args&&... args)
+      : Model(std::move(model)), Mixin(static_cast<decltype(args)>(args)...) {}
+
+  void update_motion(const update_type& pose) { Model::update_motion(pose); }
```
hidmic, line 105: @nahueespinosa why method shadowing? Same below.

beluga_amcl/src/amcl_node.cpp
```diff
@@ -706,31 +699,23 @@ std::unique_ptr<LaserLocalizationInterface2d> AmclNode::make_particle_filter(
...
+  auto get_sensor_descriptor = [this, map](std::string_view name) -> SensorDescriptor {
     if (name == kLikelihoodFieldModelName) {
       auto params = beluga::LikelihoodFieldModelParam{};
       params.max_obstacle_distance = get_parameter("laser_likelihood_max_dist").as_double();
       params.max_laser_distance = get_parameter("laser_max_range").as_double();
       params.z_hit = get_parameter("z_hit").as_double();
       params.z_random = get_parameter("z_rand").as_double();
       params.sigma_hit = get_parameter("sigma_hit").as_double();
-      return LikelihoodField{params};
+      return beluga::LikelihoodFieldModel<beluga_ros::OccupancyGrid>{params, beluga_ros::OccupancyGrid{map}};
```
hidmic, line 718: @nahueespinosa nit: a deduction guideline would be nice. Same for other sensor models.

beluga_amcl/src/amcl_node.cpp
```diff
@@ -741,17 +726,27 @@ std::unique_ptr<LaserLocalizationInterface2d> AmclNode::make_particle_filter(
...
+              sampler_params,             //
+              limiter_params,             //
+              motion_model,               //
+              sensor_model,               //
+              resample_on_motion_params,  //
+              resample_interval_params,   //
+              selective_resampling_params);
+        },
+        get_motion_descriptor(get_parameter("robot_model_type").as_string()),
+        get_sensor_descriptor(get_parameter("laser_model_type").as_string()));
```
hidmic, line 749: @nahueespinosa so much simpler :heart:

hidmic (commented):
Overall LGTM

beluga/include/beluga/localization.hpp
nahueespinosa: Because this doesn't compile :smiling_face_with_tear:

```cpp
struct Interface {
    virtual void do_the_thing() = 0;
};

struct Base {
    void do_the_thing() {}
};

struct Derived : public Interface, public Base {};
```

And something like making models inherit from their specific interface doesn't work unless we use virtual inheritance. Which in turns breaks mixins since you can't perform static downcasts.

beluga_amcl/src/amcl_node.cpp
nahueespinosa: Implicitly-generated deduction guides do the job actually.
Fixed in https://github.com/Ekumen-OS/beluga/pull/291/commits/deeaaa598ac33c92ae21c3a7d59d28d0e5a06ff5.

beluga/include/beluga/localization.hpp
hidmic: Hmm, I'm confused. Why would we need virtual inheritance? If motion mixins inherit from motion models, and motion models implement a motion model interface, that then becomes the motion mixin interface as well, not directly inherited but obtained through a type constrain on the motion model class, there's no diamond problem. Or that's how I picture it.

nahueespinosa: > If motion mixins inherit from motion models, and motion models implement a motion model interface, that then becomes the motion mixin interface as well

That doesn't work like that, the compiler doesn't piece together that the motion mixin inherited members are actually implementing pure interface members inherited by the whole mixin.

```cpp
struct Interface { virtual void do_the_thing() = 0; };

struct MotionModel { void do_the_thing() {} };

struct OtherMixins : public Interface {};

struct MotionMixin : public OtherMixins, public MotionModel {};

struct MostDerived : public MotionMixin {};
```

```bash
<source>:17:33: error: invalid cast to abstract class type 'MostDerived'
   17 |     auto instance = MostDerived{};
      |                                 ^
<source>:14:8: note:   because the following virtual functions are pure within 'MostDerived':
   14 | struct MostDerived : public MotionMixin {};
      |        ^~~~~~~~~~~
```

While doing this, I thought of this potential solution:
```cpp
struct Interface { virtual void do_the_thing() = 0; };

struct MotionModel : virtual public Interface { void do_the_thing() {} };

struct OtherMixins : virtual public Interface {};

struct MotionMixin : public OtherMixins, public MotionModel {};

struct MostDerived : public MotionMixin {};
```
But that breaks compile time static casts for self members.

hidmic: > That doesn't work like that, the compiler doesn't piece together that the motion mixin inherited members are actually implementing pure interface members inherited by the whole mixin.

Right. What I'm proposing is closer to your second example, on the basis that there will not be any mixins inheriting `Interface` directly, only through `MotionModel`. Thus virtual inheritance is not necessary.

nahueespinosa: And then we cast the particle filter mixin pointer in the user code to the appropriate interface depending of if we want to update the sensor measurement or we want to _resample_ for instance?

hidmic: Hmm, but that will make it impossible to define a particle filter interface. Nevermind. You are right, there is no way around shadowing.

nahueespinosa: Sadly right... The worst kind of right.

nahueespinosa: I think it's not worth it. The future (ranges) looks greener and this is just a transition patch.

hidmic (approved):
Mixins are getting uglier. It may be the contrast with the new stuff.

beluga/include/beluga/localization.hpp
hidmic: Not even suggesting it (posts are getting interleaved :sweat_smile:). That'd be awful. This patch is OK as-is.).

PR #293, opened 2024-01-20
Run system tests with parallel execution enabled

nahueespinosa:
#### Type of change

### Checklist

### DIFF
--- a/beluga_system_tests/test/test_system_new.cpp
+++ b/beluga_system_tests/test/test_system_new.cpp
@@ -123,8 +123,6 @@ auto particle_filter_test(
 
   using Particle = std::tuple<Sophus::SE2d, beluga::Weight>;
 
-  [[maybe_unused]] static thread_local auto engine = std::mt19937{std::random_device()()};
-
   auto hasher = beluga::spatial_hash<Sophus::SE2d>{0.1, 0.1, 0.1};
 
   // Use the initial distribution to initialize particles.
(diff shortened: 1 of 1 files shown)


beluga_system_tests/test/test_system_new.cpp
```diff
@@ -181,14 +180,19 @@ auto particle_filter_test(
...
     /**
-     * particles |= beluga::actions::propagate(motion(odom)) |
-     *              beluga::actions::reweight(sensor(measurement));
+     * particles |= beluga::actions::propagate(std::execution::par, motion(odom)) |
+     *              beluga::actions::reweight(std::execution::par, sensor(measurement));
      */
     motion.update_motion(odom);
     sensor.update_sensor(std::move(measurement));
-    particles |=
-        beluga::actions::propagate([&motion](const auto& state) { return motion.apply_motion(state, engine); }) |
```
nahueespinosa, line 190: Just for a quick test, I tried adding `std::execution::par`, but this code made all the worker threads use the same engine instantiated above, causing the process to hang.

This patch does it the right way, so I figured I better submit this before someone else runs into the same issue.

hidmic: Good find and +1 to the fix, but I think we should make sure this is seldom required. Perhaps turning these PRNGs into thread local singletons and using them implicitly by default i.e. models, distributions, views, etc., would be templated on some engine type, invoking the corresponding singleton when necessary. I don't love it either, but with no control over what execution policies do (and don't) copy and when, the alternatives are way more convoluted.

hidmic (approved):

beluga_system_tests/test/test_system_new.cpp
nahueespinosa: Yeah, currently our random views have default thread_local engine. Doing something similar for models should be enough for the average user. Maybe we don't let users override the engine for motion and sensor models since it's tricky to get right, or we let them do it at their own risk, I'm not which is best.

PR #295, opened 2024-01-22
Add adaptive probability estimator

nahueespinosa:
### Proposed changes

Related to #279.

#### Type of change

### Checklist


beluga/include/beluga/algorithm/adaptive_probability_estimator.hpp
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * The addition of random samples allows the filter to recover in case it converged to
+ * the wrong estimate, adding an additional level of robustness.
+ *
+ * This estimator averages the total weight of the particles and computes the ratio
+ * between a short-term and a long-term average over time.
+ *
+ * See Probabilistic Robotics \cite thrun2005probabilistic, Chapter 8.3.3.
+ */
+class AdaptiveProbabilityEstimator {
```
nahueespinosa, line 40: Maybe this can be called `RecoveryProbabilityEstimator` since adaptive is used multiple times for different things in particle filter literature.

hidmic: Perhaps `ThrunRecoveryProbabilityEstimator`? Particle filter literature is very prolific, it won't be the last time we have different authors using the same name for different things, or even multiple implementations of the same concept

beluga/include/beluga/algorithm/adaptive_probability_estimator.hpp
```diff
@@ -0,0 +1,100 @@
...
+    fast_filter_.reset();
+    probability_ = 0.0;
+  }
+
+  /// Update the estimation based on a particle range.
+  template <class Range>
+  constexpr void update(Range&& range) {
+    static_assert(ranges::sized_range<Range>);
+    static_assert(beluga::is_particle_range_v<Range>);
+    const auto size = static_cast<double>(range.size());
```
hidmic, line 69: @nahueespinosa nit: consider checking for zero size in its integer form, and only casting when computing a weight average. Checking floating point numbers for equality is ill-defined and compilers sometimes emit warnings when you do.

beluga/include/beluga/algorithm/adaptive_probability_estimator.hpp
```diff
@@ -0,0 +1,100 @@
...
+    if (slow_average == 0.0) {
+      probability_ = 0.0;
+      return;
+    }
+
+    probability_ = std::clamp(1.0 - fast_average / slow_average, 0.0, 1.0);
+  }
+
+  /// Returns the estimated random state probability to be used by the filter.
+  constexpr double operator()() const noexcept { return probability_; }
```
hidmic, line 90: @nahueespinosa meta: do we need to track the probability? Why not return it on update? Moreover, having a consistent API with exponential filters would be nice.

beluga/include/beluga/algorithm/adaptive_probability_estimator.hpp
nahueespinosa: Done!

beluga/include/beluga/algorithm/adaptive_probability_estimator.hpp
nahueespinosa: Makes sense, fixed!

beluga/include/beluga/algorithm/thrun_recovery_probability_estimator.hpp
```diff
@@ -0,0 +1,96 @@
...
+      reset();
+      return 0.0;
+    }
+
+    const double total_weight = ranges::accumulate(beluga::views::weights(range), 0.0);
+    const double average_weight = total_weight / static_cast<double>(size);
+    const double fast_average = fast_filter_(average_weight);
+    const double slow_average = slow_filter_(average_weight);
+
+    if (slow_average == 0.0) {
```
hidmic, line 82: @nahueespinosa nit: consider softening the equality comparison ie.:
```suggestion
    if (std::abs(slow_average) < std::numeric_limits<double>::epsilon()) {
```

hidmic (approved):
LGTM

PR #296, opened 2024-01-27
Extend sample view to sample from random distributions

nahueespinosa:
### Proposed changes

This saves lines of code by extending the sample view implementation to convert any random distribution (a callable that takes a URNG as an input argument) into a range.

It also makes the implementation easier to follow by removing the use of default arguments (which play poorly with overloads).

Related to #279.

#### Type of change

### Checklist


hidmic (approved):
Nothing to add. Flawless.

PR #298, opened 2024-01-28
Extend multivariate distribution to support different result types

nahueespinosa:
### Proposed changes

As the title says, this patch extends the multivariate normal distribution type to support row vectors, SO2, and SE2 elements as different representations of the result type.

Related to #279.

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/random/multivariate_distribution_traits.hpp
+++ b/beluga/include/beluga/random/multivariate_distribution_traits.hpp
@@ -17,7 +17,6 @@
 
 #include <Eigen/Core>
 #include <sophus/se2.hpp>
-#include <sophus/se3.hpp>
 
 /**
  * \file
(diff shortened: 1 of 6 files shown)


beluga/include/beluga/random/multivariate_distribution_traits.hpp
```diff
@@ -0,0 +1,168 @@
...
+
+  /// The covariance matrix type.
+  using covariance_type = typename Eigen::Matrix<scalar_type, 3, 3>;
+
+  /// Convert from result to vector representation.
+  [[nodiscard]] static constexpr vector_type to_vector(const result_type& t) { return t.log(); }
+
+  /// Convert from vector to result representation.
+  [[nodiscard]] static constexpr result_type from_vector(const vector_type& v) {
+    return Sophus::SO3<scalar_type>::exp(v);
```
hidmic, line 134: @nahueespinosa hmm, this caught my attention, then sent me down a rabbit hole. I think this is not correct. If we just project a gaussian through the exponential map, we are not even going to get a unimodal distribution. That tangent space vector $\xi \in \mathbb{R}^3$ is an axis-angle representation (see [here](https://cvg.cit.tum.de/_media/members/demmeln/nurlanov2021so3log.pdf)). If we sample from a standard normal distribution, that axis could flip (the plot in section 2.2.1, [here](https://mediatum.ub.tum.de/doc/1273706/195643554083.pdf) is quite telling). It gets worse for non-zero mean distributions -- I'm not even sure it will pass the internal [unit norm checks](https://docs.ros.org/en/noetic/api/sophus/html/so3_8hpp_source.html#l00571) in Sophus.

Sampling spherical distributions is tricky. We can get away with it for SO(2) because the tangent space is $\mathbb{R}$ and unconstrained (Sophus wraps around angles on exponentiation). For SO(3), the analogue to the normal distribution seems to be the [Fisher-Bingham or Kent distribution](https://www.kent.ac.uk/smsas/personal/ak96/fgibbs_12.pdf). Unfortunately, it uses a different parameterization (direction and concentration parameters) and it is non-trivial to sample. To make matters worse, at least in ROS, SO(3) covariances are typically expressed in roll-pitch-yaw coordinates (I guess because they are intuitive :man_shrugging:).

So I would suggest we remove this for now. We can work on a proper spherical distribution later on. We could inspire ourselves in [OMPL](https://ompl.kavrakilab.org/SO3StateSpace_8cpp_source.html#l00119)'s approach (which I don't fully grasp, with those magic constants and all), or we could sample roll-pitch-yaw from the marginal gaussian, wrap them around component-wise, and then compose the corresponding rotations with the mean. Or we could try to implement that Fisher-Bingham distribution, which opens up Beluga to interesting attitude estimation applications but gets into [MCMC territory](https://arxiv.org/pdf/1310.8110.pdf).

hidmic (changes requested):

beluga/include/beluga/random/multivariate_distribution_traits.hpp
nahueespinosa: Argh, thank you for the comprehensive explanation. You rock!

Removed in https://github.com/Ekumen-OS/beluga/pull/298/commits/a9aa4dcc6183ffdced560f43b76359ffa4e35aee.

Will look more into this when I have some cycles.

hidmic (approved):

PR #319, opened 2024-02-10
Add uniform grid distribution

nahueespinosa:
### Proposed changes

Related to #279.

#### Type of change

### Checklist

### Additional comments

Existing sensor models were not changed, that can probably be done in a follow-up PR.

### DIFF
--- a/beluga/include/beluga/random/uniform_grid_distribution.hpp
+++ b/beluga/include/beluga/random/uniform_grid_distribution.hpp
@@ -1,0 +1,81 @@
+// Copyright 2024 Ekumen, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef BELUGA_RANDOM_UNIFORM_GRID_DISTRIBUTION_HPP
+#define BELUGA_RANDOM_UNIFORM_GRID_DISTRIBUTION_HPP
+
+#include <random>
+
+#include <range/v3/utility/random.hpp>
+#include <sophus/se2.hpp>
+
+#include <beluga/sensor/data/occupancy_grid.hpp>
+
+namespace beluga {
+
+/// Primary template for a uniform grid distribution.
+/**
+ * \tparam T The result type for sampling from the distribution.
+ */
+template <class T>
+class UniformGridDistribution;
+
+/// Specialization of uniform grid distribution for Sophus::SE2d.
+template <>
+class UniformGridDistribution<Sophus::SE2d> {
+ public:
+  /// Constructs a uniform grid distribution based on the provided occupancy grid.
+  /**
+   * \tparam Derived A type derived from BaseOccupancyGrid2.
+   * \param grid The occupancy grid from which free states will be computed.
+   */
+  template <class Derived>
+  constexpr explicit UniformGridDistribution(const BaseOccupancyGrid2<Derived>& grid)
+      : free_states_{compute_free_states(static_cast<const Derived&>(grid))},
+        distribution_{0, free_states_.size() - 1} {
+    assert(free_states_.size() > 0);
+  }
+
+  /// Generates a random 2D pose.
+  /**
+   * This function generates a random pose by sampling a random rotation
+   * from SO2 space and a random translation from the precomputed free states
+   * based on the provided occupancy grid.
+   *
+   * \tparam URNG The type of the random number generator.
+   * \param engine The random number generator engine (default is a default generator).
+   * \return A random Sophus::SE2d pose.
+   */
+  template <class URNG = typename ranges::detail::default_random_engine>
+  [[nodiscard]] Sophus::SE2d operator()(URNG& engine = ranges::detail::get_random_engine()) {
+    return {Sophus::SO2d::sampleUniform(engine), free_states_[distribution_(engine)]};
+  }
+
+ private:
+  std::vector<Eigen::Vector2d> free_states_;                 ///< Vector containing free states.
+  std::uniform_int_distribution<std::size_t> distribution_;  ///< Uniform distribution for indices.
+
+  template <class Grid>
+  static std::vector<Eigen::Vector2d> compute_free_states(const Grid& grid) {
+    return grid.coordinates_for(grid.free_cells(), Grid::Frame::kGlobal) | ranges::to<std::vector>;
+  }
+};
+
+/// Deduction guide for 2D occupancy grids.
+template <class Derived>
+UniformGridDistribution(const BaseOccupancyGrid2<Derived>&) -> UniformGridDistribution<Sophus::SE2d>;
+
+}  // namespace beluga
+
+#endif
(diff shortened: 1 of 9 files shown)


beluga_system_tests/test/test_system_new.cpp
```diff
@@ -170,13 +173,10 @@ auto particle_filter_test(
...
-      auto make_random_particle = [&sensor] {
-        static thread_local auto engine = std::mt19937{std::random_device()()};
-        return beluga::make_from_state<Particle>(sensor.make_random_state(engine));
-      };
-
       particles |= beluga::views::sample |
-                   beluga::views::random_intersperse(make_random_particle, random_state_probability) |
+                   beluga::views::random_intersperse(
+                       ranges::compose(beluga::make_from_state<Particle>, std::ref(map_distribution)),
+                       random_state_probability) |
```
nahueespinosa, line 179: meta: We can make `random_intersperse` smarter so we can write this:

```suggestion
                   beluga::views::random_intersperse(std::ref(map_distribution), random_state_probability) |
```

I'm not super convinced that we should though.

beluga/test/beluga/random/test_uniform_grid_distribution.cpp
```diff
@@ -0,0 +1,97 @@
...
+  constexpr double kResolution = 1.0;
+  const auto grid = beluga::testing::StaticOccupancyGrid<3, 3>{
+      {true, false, true,   //
+       false, true, false,  //
+       true, false, true},
+      kResolution,
+  };
+  auto distribution = beluga::UniformGridDistribution{grid};
+
+  auto output = ranges::views::generate(distribution) | ranges::views::take_exactly(kSize);
```
hidmic, line 67: @nahueespinosa meta: I was expecting to see `beluga::views::sample` instead.

beluga/include/beluga/random/uniform_grid_distribution.hpp
```diff
@@ -0,0 +1,81 @@
...
+   * This function generates a random pose by sampling a random rotation
+   * from SO2 space and a random translation from the precomputed free states
+   * based on the provided occupancy grid.
+   *
+   * \tparam URNG The type of the random number generator.
+   * \param engine The random number generator engine (default is a default generator).
+   * \return A random Sophus::SE2d pose.
+   */
+  template <class URNG = typename ranges::detail::default_random_engine>
+  [[nodiscard]] Sophus::SE2d operator()(URNG& engine = ranges::detail::get_random_engine()) {
```
hidmic, line 61: @nahueespinosa hmm, why provide a default URNG? STL distributions don't, and we already have `beluga::views::sample` to inject the ranges-v3 specific URNG.

beluga/include/beluga/random/uniform_grid_distribution.hpp
```diff
@@ -0,0 +1,81 @@
...
+class UniformGridDistribution<Sophus::SE2d> {
+ public:
+  /// Constructs a uniform grid distribution based on the provided occupancy grid.
+  /**
+   * \tparam Derived A type derived from BaseOccupancyGrid2.
+   * \param grid The occupancy grid from which free states will be computed.
+   */
+  template <class Derived>
+  constexpr explicit UniformGridDistribution(const BaseOccupancyGrid2<Derived>& grid)
+      : free_states_{compute_free_states(static_cast<const Derived&>(grid))},
```
hidmic, line 45: @nahueespinosa meta: this makes it not uniform, strictly speaking. Don't have a better name for the distribution though :)

beluga_system_tests/test/test_system_new.cpp
hidmic: +1 to having `beluga::views::random_intersperse` accept a distribution instead of a callable. It's not a substantial detour, API wise, and for a library of Monte Carlo methods, pretty much everything is a distribution.

beluga/include/beluga/random/uniform_grid_distribution.hpp
hidmic: Also, could the distribution yield particles instead of particle state? Perhaps it's two distributions, one composing the other.

hidmic (commented):
First pass.

beluga/include/beluga/random/uniform_grid_distribution.hpp
nahueespinosa: Yeah, FreeSpaceGridDistribution distribution maybe?

beluga_system_tests/test/test_system_new.cpp
nahueespinosa: Hmm, I think `beluga::views::random_intersperse` should take callables. We can also take distributions, but the thing here is that those distributions/callables should generate values that are compatible with the input range.

In this case, `beluga::views::sample` is generating particles (as it should) since it's sampling from a particle range.
The question is, should we make `beluga::views::random_intersperse` check if the callable does not generate particles but state values compatible with the input particles and call `beluga::make_from_state` internally?

beluga/include/beluga/random/uniform_grid_distribution.hpp
nahueespinosa: > STL distributions don't, and we already have beluga::views::sample to inject the ranges-v3 specific URNG.

That's fair.

> Also, could the distribution yield particles instead of particle state?

It can, but it's not consistent with `MultivariateNormalDistribution` for instance. In my mind, distributions generate states that we use to initialize particles, there is no need to make that relationship stronger than that in my opinion.

> Perhaps it's two distributions, one composing the other.

Could you elaborate on that?

nahueespinosa: See https://github.com/Ekumen-OS/beluga/pull/319/commits/85f562321b322545c73cde9256e91784de745b6d.

beluga_system_tests/test/test_system_new.cpp
hidmic: I like the current state of affairs. No need for `random_intersperse` to do more than it should.

beluga_system_tests/test/test_system_new.cpp
```diff
@@ -170,13 +173,8 @@ auto particle_filter_test(
...
       }
 
-      auto make_random_particle = [&sensor] {
-        static thread_local auto engine = std::mt19937{std::random_device()()};
-        return beluga::make_from_state<Particle>(sensor.make_random_state(engine));
-      };
-
       particles |= beluga::views::sample |
-                   beluga::views::random_intersperse(make_random_particle, random_state_probability) |
+                   beluga::views::random_intersperse(std::ref(map_distribution), random_state_probability) |
```
hidmic, line 177: Just to be sure, we are passing by reference to avoid unnecessarily expensive copies, right?

hidmic (approved):
LGTM but for a few minor comments.

beluga_system_tests/test/test_system_new.cpp
nahueespinosa: Exactly!

beluga/include/beluga/random/uniform_grid_distribution.hpp
nahueespinosa: Done in https://github.com/Ekumen-OS/beluga/pull/319/commits/f308c57f81651e46182ce99d45bcf0b9b0eafa24!

hidmic:
@nahueespinosa I need this patch to wrap up the functional form sensor models :)

hidmic:
@nahueespinosa BTW we are going to need a distribution for landmark maps as well :eyes:

nahueespinosa:
@hidmic Ah, yes. Noted, shouldn't be too hard.

hidmic:
> Ah, yes. Noted, shouldn't be too hard.

@nahueespinosa Will you, or should I?

nahueespinosa:
@hidmic I have no time today, if you have some cycles, please go ahead!

PR #320, opened 2024-02-13
Add generic circular array container

hidmic:
### Proposed changes

This patch introduces a generic circular array container, ancillary to the upcoming functional form of motion models.

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/circular_array.hpp
+++ b/beluga/include/beluga/circular_array.hpp
@@ -1,0 +1,539 @@
+// Copyright 2024 Ekumen, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef BELUGA_CIRCULAR_ARRAY_HPP
+#define BELUGA_CIRCULAR_ARRAY_HPP
+
+#include <array>
+#include <cstdint>
+#include <iterator>
+#include <stdexcept>
+#include <type_traits>
+#include <utility>
+
+#include <beluga/utility/indexing_iterator.hpp>
+
+/**
+ * \file
+ * \brief Implementation of a generic circular array container.
+ */
+
+namespace beluga {
+
+/// Feature flags for circular arrays.
+enum class CircularArrayFeatureFlags : int {
+  kNone = 0x00,
+  kRolloverOnWrite = 0x01,    ///<! If enabled, older values in the array are overwritten by
+                              /// newer values if the array has reached its maximum size already.
+  kExtrapolateOnRead = 0x02,  ///<! If enabled, the back value is extrapolated for constant
+                              /// accesses up to the array maximum size.
+  kLayoutReversal = 0x04      ///<! If enabled, the circular array memory layout is reversed so that
+                              /// values can be pushed to the front rather than the back of the array.
+};
+
+/// Bitwise OR operator overload to combine two feature flags in a single mask-like flag.
+inline constexpr CircularArrayFeatureFlags operator|(CircularArrayFeatureFlags lflag, CircularArrayFeatureFlags rflag) {
+  return static_cast<CircularArrayFeatureFlags>(
+      static_cast<typename std::underlying_type<CircularArrayFeatureFlags>::type>(lflag) |
+      static_cast<typename std::underlying_type<CircularArrayFeatureFlags>::type>(rflag));
+}
+
+/// Bitwise AND operator overload to check of the presence of a feature `flag` in a feature `mask`.
+inline constexpr bool operator&(CircularArrayFeatureFlags mask, CircularArrayFeatureFlags flag) {
+  return static_cast<bool>(
+      static_cast<typename std::underlying_type<CircularArrayFeatureFlags>::type>(mask) &
+      static_cast<typename std::underlying_type<CircularArrayFeatureFlags>::type>(flag));
+}
+
+/// An implementation of generic, non-threadsafe circular array.
+/**
+ * Modelled after `std::array`. Its maximum size is fixed at compile-time. Most operations are O(1)
+ * but for a few that exhibit worse case O(N) (not considering the complexity of move-construction
+ * and move-assignment of array elements of type T). Additionally, it features optional indexing
+ * reversal (making it a LIFO rather than a FIFO data structure), optional back value extrapolation
+ * for constant read accesses, and optional rollover on write accesses (i.e. overwriting the oldest
+ * value when pushing to an array instance that is full).
+ *
+ * \tparam T Element type.
+ *   It must be [default constructible](https://en.cppreference.com/w/cpp/named_req/DefaultConstructible),
+ *   [move constructible](https://en.cppreference.com/w/cpp/named_req/MoveConstructible), and
+ *   [move assignable](https://en.cppreference.com/w/cpp/named_req/MoveAssignable).
+ * \tparam N Maximum size. It must be a positive integer.
+ * \tparam F Array feature flags, to enable optional functionality. It defaults to none.
+ */
+template <typename T, std::size_t N, CircularArrayFeatureFlags F = CircularArrayFeatureFlags::kNone>
+class CircularArray {
+ public:
+  /// Value type of the array.
+  using value_type = T;
+  /// Size type of the array.
+  using size_type = std::size_t;
+  /// Size difference type of the array.
+  using difference_type = std::ptrdiff_t;
+  /// Value reference type of the array.
+  using reference = value_type&;
+  /// Constant value reference type of the array.
+  using const_reference = const value_type&;
+  /// Value pointer type of the array.
+  using pointer = value_type*;
+  /// Constant value pointer type of the arra.y
+  using const_pointer = const value_type*;
+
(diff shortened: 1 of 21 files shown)


beluga/include/beluga/circular_array.hpp
```diff
@@ -0,0 +1,539 @@
...
+ *
+ * \tparam T Element type.
+ *   It must be [default constructible](https://en.cppreference.com/w/cpp/named_req/DefaultConstructible),
+ *   [move constructible](https://en.cppreference.com/w/cpp/named_req/MoveConstructible), and
+ *   [move assignable](https://en.cppreference.com/w/cpp/named_req/MoveAssignable).
+ * \tparam N Maximum size. It must be a positive integer.
+ * \tparam F Array feature flags, to enable optional functionality. It defaults to none.
+ */
+template <typename T, std::size_t N, CircularArrayFeatureFlags F = CircularArrayFeatureFlags::kNone>
+class CircularArray {
```
nahueespinosa, line 76: Funnily enough, this needs to define an `allocator_type` for `ranges::to` to work in range-v3 10.0:

https://github.com/ericniebler/range-v3/blob/0.10.0/include/range/v3/range/conversion.hpp#L262

Maybe add `using allocator_type = void;` somewhere down below.

beluga/include/beluga/circular_array.hpp
```diff
@@ -0,0 +1,539 @@
...
+}
+
+/// Swaps arrays `a` and `b`.
+/**
+ * This is an `std::swap` overload that relies on argument-dependent lookup (ADL).
+ */
+template <typename T, std::size_t N, CircularArrayFeatureFlags F, CircularArrayFeatureFlags G>
+constexpr void swap(CircularArray<T, N, F>& a, CircularArray<T, N, G>& b) {
+  a.swap(b);
+};
```
nahueespinosa, line 521: The linter is complaining about this.

```suggestion
}
```

nahueespinosa (commented):
@hidmic This is looking amazing. I didn't finish the first pass yet, but I figured I'd leave some comments so you don't have to debug `ranges::to`.

beluga/include/beluga/circular_array.hpp
```diff
@@ -0,0 +1,539 @@
...
+   * \param first Iterator to the beginning of a range.
+   * \param last Iterator to the end of a range.
+   * \throws std::length_error If the range does not fit in the array
+   * (and the rollover on write feature is not enabled).
+   */
+  template <
+      typename Iterator,
+      typename = std::enable_if_t<std::is_same_v<T, typename std::iterator_traits<Iterator>::value_type>>>
+  CircularArray(Iterator first, Iterator last) {
+    if constexpr (F & CircularArrayFeatureFlags::kLayoutReversal) {
```
nahueespinosa, line 123: Based on the documentation I would expect this test to pass, but it doesn't :thinking::

```cpp
TEST(CircularArrayTest, RolloverOnWriteWithLayoutReversalFromRange) {
  constexpr auto kFlags =
      beluga::CircularArrayFeatureFlags::kLayoutReversal | beluga::CircularArrayFeatureFlags::kRolloverOnWrite;
  auto input = std::array{1, 2, 3};
  auto array = beluga::CircularArray<int, 2, kFlags>{input.begin(), input.end()};
  EXPECT_THAT(array, ::testing::ElementsAre(3, 2));  // output is {1, 2}
}
```

We can maybe add this and similar tests for this constructor to clarify the expected behavior.

beluga/include/beluga/circular_array.hpp
nahueespinosa, line 1: meta: Should we have this file and the TupleVector implementation in a subfolder (i.e. `beluga/containers`)?

beluga/include/beluga/utility/indexing_iterator.hpp
```diff
@@ -0,0 +1,205 @@
...
+  bool operator!=(const IndexingIterator& other) const noexcept { return !(*this == other); }
+
+ private:
+  Indexable* indexable_{nullptr};
+  Index cursor_{};
+};
+
+/// Rewinds `iterator` a given `offset`, yielding a modified copy.
+template <class Indexable, class Index>
+[[nodiscard]] IndexingIterator<Indexable, Index> operator+(
```
nahueespinosa, line 197: Implementing this as a hidden friend so it is only found by ADL would be much nicer to the compiler (and error messages for unrelated operator+ overloading errors).

beluga/include/beluga/circular_array.hpp
hidmic: Haha, I love the `HACKHACK` comment :joy:

nahueespinosa (commented):
@hidmic Just a couple more comments. Again, this is awesome!

beluga/include/beluga/utility/indexing_iterator.hpp
nahueespinosa: Wish we could do the same for the enum pipe operator overloads.

hidmic: Good idea. Done in https://github.com/Ekumen-OS/beluga/pull/320/commits/d451b5add74832b2397a3625c1a58d4565b409eb.

beluga/include/beluga/circular_array.hpp
hidmic: Done in https://github.com/Ekumen-OS/beluga/pull/320/commits/d451b5add74832b2397a3625c1a58d4565b409eb.

beluga/include/beluga/circular_array.hpp
hidmic: Added a couple tests, with and without layout reversal in https://github.com/Ekumen-OS/beluga/pull/320/commits/d451b5add74832b2397a3625c1a58d4565b409eb.

Layout reversal only dictates to which end of the array you can push and from which end of the array you can pop. Everything else (constructors, iterators, indices, accessors) still operates left to right, front to back.

beluga/include/beluga/circular_array.hpp
hidmic: Yes! Fixed in https://github.com/Ekumen-OS/beluga/pull/320/commits/e1e3e0684cce1d489e0d87b469a2afa86f960477.

beluga/include/beluga/circular_array.hpp
hidmic: Done in https://github.com/Ekumen-OS/beluga/commit/d451b5add74832b2397a3625c1a58d4565b409eb.

beluga/include/beluga/circular_array.hpp
```diff
@@ -0,0 +1,539 @@
...
+   *
+   * \param first Iterator to the beginning of a range.
+   * \param last Iterator to the end of a range.
+   * \throws std::length_error If the range does not fit in the array
+   * (and the rollover on write feature is not enabled).
+   */
+  template <
+      typename Iterator,
+      typename = std::enable_if_t<std::is_same_v<T, typename std::iterator_traits<Iterator>::value_type>>>
+  CircularArray(Iterator first, Iterator last) {
```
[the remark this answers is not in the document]
hidmic: Changed in https://github.com/Ekumen-OS/beluga/commit/d451b5add74832b2397a3625c1a58d4565b409eb.

beluga/include/beluga/utility/indexing_iterator.hpp
hidmic: > Wish we could do the same for the enum pipe operator overloads.

Do we need to? I mean, other than because it looks cleaner. Enum classes cannot derive from each other and are not convertible to each other unless explicit provisions are put in place.

nahueespinosa: See how the enum operator pipe overload is listed as a candidate when `ranges::to` fails to compile. That's unfortunate. To be fair, a couple of overloads in the std namespace are listed as well, but not that many.

beluga/include/beluga/circular_array.hpp
nahueespinosa: Got it!

nahueespinosa (approved):
LGTM!

beluga/include/beluga/containers/circular_array.hpp
```diff
@@ -0,0 +1,543 @@
...
+   * \param first Iterator to the beginning of a range.
+   * \param last Sentinel for the end of a range.
+   * \throws std::length_error If the range does not fit in the array
+   * (and the rollover on write feature is not enabled).
+   */
+  template <
+      typename Iterator,
+      typename Sentinel,
+      typename = std::enable_if_t<std::is_same_v<T, typename std::iterator_traits<Iterator>::value_type>>>
+  CircularArray(Iterator first, Sentinel const last) {
```
nahueespinosa: I think clang-tidy is being silly here.

```suggestion
  CircularArray(Iterator first, Sentinel last) {  // NOLINT(readability-non-const-parameter)
```

beluga/include/beluga/utility/indexing_iterator.hpp
hidmic: I see. Yeah, I agree it adds unnecessary clutter. I guess we could also drop the strongly typed enums and use plain integer flags. I was on the fence about that too.

beluga/include/beluga/containers/circular_array.hpp
hidmic: It's funny though, it only shows on the Noetic container.

nahueespinosa: :shrug:

beluga/include/beluga/utility/indexing_iterator.hpp
nahueespinosa: Range-v3 does something very similar with range categories: https://github.com/ericniebler/range-v3/blob/97452bb3eb74a73fc86504421a6a27c92bce6b99/include/range/v3/view/any_view.hpp#L43C16-L43C24

I think it's fine the way it is.

hidmic:
All green, going in!

PR #321, opened 2024-02-13
Migrate motion models to functional form

hidmic:
### Proposed changes

This patch migrates motion models to a functional form. Combined with #320, this affords:

```c++
beluga::actions::propagate(std::execution::par, motion(control_action_window << odom))
```

when propagating particles. 

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/localization.hpp
+++ b/beluga/include/beluga/localization.hpp
@@ -16,6 +16,7 @@
 #define BELUGA_MIXIN_LOCALIZATION_HPP
 
 #include <beluga/algorithm/estimation.hpp>
+#include <beluga/containers/circular_array.hpp>
 #include <beluga/mixin.hpp>
 #include <beluga/motion.hpp>
 #include <beluga/sensor.hpp>
@@ -95,14 +96,32 @@ using LaserLocalizationInterface2d = beluga::compose_interfaces<
 
 /// Wrapper to convert a motion model class into a mixin.
 template <typename Mixin, typename Model>
-struct MotionMixin : public Mixin, public Model {
-  using update_type = typename Model::update_type;
+class MotionMixin : public Mixin {
+ public:
+  using update_type = std::tuple_element_t<0, typename Model::control_type>;
+  using state_type = typename Model::state_type;
 
   template <typename... Args>
   constexpr explicit MotionMixin(Model model, Args&&... args)
-      : Model(std::move(model)), Mixin(static_cast<decltype(args)>(args)...) {}
+      : Mixin(static_cast<decltype(args)>(args)...), model_(std::move(model)) {}
 
-  void update_motion(const update_type& pose) { Model::update_motion(pose); }
+  template <class Generator>
+  [[nodiscard]] state_type apply_motion(const state_type& state, Generator& gen) {
+    if (state_sampling_function_.has_value()) {
+      return (*state_sampling_function_)(state, gen);
+    }
+    return state;
+  }
+
+  void update_motion(const update_type& pose) { state_sampling_function_.emplace(model_(control_window_ << pose)); }
+
+ private:
+  Model model_;
+  static constexpr auto kWindowSize = std::tuple_size_v<typename Model::control_type>;
+  RollingWindow<update_type, kWindowSize> control_window_;
+  using StateSamplingFunction =
+      decltype(std::declval<Model>()(std::declval<RollingWindow<update_type, kWindowSize>&>()));
+  std::optional<StateSamplingFunction> state_sampling_function_;
 };
 
 /// Wrapper to convert a sensor model class into a mixin.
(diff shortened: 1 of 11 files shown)


beluga/include/beluga/localization.hpp
```diff
@@ -95,14 +96,34 @@ using LaserLocalizationInterface2d = beluga::compose_interfaces<
...
+    state_sampling_function_.emplace(model_(control_action_window_ << pose));
+  }
+
+ private:
+  Model model_;
+  static constexpr auto kWindowSize = std::tuple_size_v<typename Model::control_action_type>;
+  RollingWindow<update_type, kWindowSize> control_action_window_;
+  using StateSamplingFunction =
+      decltype(std::declval<Model>()(std::declval<RollingWindow<update_type, kWindowSize>&>()));
+  std::optional<StateSamplingFunction> state_sampling_function_;
```
hidmic, line 126: FWIW I don't love this, but it'll go away when mixins do.

beluga/include/beluga/motion/differential_drive_model.hpp
```diff
@@ -83,69 +87,56 @@ class DifferentialDriveModel {
...
-                          params_.translation_noise_from_rotation *
-                              (rotation_variance(first_rotation) + rotation_variance(second_rotation)))};
-        second_rotation_params_ = DistributionParam{
-            second_rotation.log(), std::sqrt(
-                                       params_.rotation_noise_from_rotation * rotation_variance(second_rotation) +
-                                       params_.rotation_noise_from_translation * distance_variance)};
-      }
-    }
-    last_pose_ = pose;
+  template <class ControlAction, typename = common_tuple_type_t<ControlAction, control_action_type>>
```
hidmic, line 98: This is the trick that allows tuple-like containers like the circular array to be used in place for standard tuples. This is arguably too complex. I convinced myself that was OK on the basis that (a) this is a core library abstraction, and (b) users don't need to go this far when implementing their models. They can fix all types and that will work too.

beluga_system_tests/test/test_system_new.cpp
```diff
@@ -141,24 +141,18 @@ auto particle_filter_test(
...
      */
-    motion.update_motion(odom);
     sensor.update_sensor(std::move(measurement));
-    particles |= beluga::actions::propagate(
-                     std::execution::par,
-                     [&motion](const auto& state) {
-                       static thread_local auto engine = std::mt19937{std::random_device()()};
-                       return motion.apply_motion(state, engine);
-                     }) |
+    particles |= beluga::actions::propagate(std::execution::par, motion(control_action_window << odom)) |
```
hidmic, line 155: This is the best I could come up with that is still a one-liner. It is in line with what `Eigen` does to populate matrices, but I know it may be somewhat controversial.

beluga/include/beluga/motion/differential_drive_model.hpp
nahueespinosa: > They can fix all types and that will work too.

+1

beluga_system_tests/test/test_system_new.cpp
nahueespinosa: Looks great to me!

nahueespinosa (approved):
@hidmic Love it.

beluga/include/beluga/motion/differential_drive_model.hpp
```diff
@@ -64,15 +68,15 @@ struct DifferentialDriveModelParam {
...
  *
  * See Probabilistic Robotics \cite thrun2005probabilistic Chapter 5.4.2.
  */
 class DifferentialDriveModel {
  public:
-  /// Update type of the motion model, same as the state_type in the odometry model.
-  using update_type = Sophus::SE2d;
-  /// State type of a particle.
+  /// Current and previous odometry estimates as motion model control action.
+  using control_action_type = std::tuple<Sophus::SE2d, Sophus::SE2d>;
```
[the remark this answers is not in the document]
hidmic: > I was expecting [previous, current] instead of [current, previous], but this makes sense with the rolling window usage. I wonder if others will have the same expectations.

So, I think in discrete difference equations, like $y[n] = b_0 x[n] + b_1 x[n - 1] + ...$. That's why $u = (x[n], x[n - 1])$ makes sense to me, but I really don't mind. We can drop layout reversal for rolling windows and get it in reverse.

> nit^n: What do you think about control_type instead of control_action_type? Just putting it out there, I'm fine with the current choice.

I was toying with `control_type` at the beginning too. At the end, I switch to `control_action_type` because it seemed more accurate to me. But again, I don't mind strongly about it.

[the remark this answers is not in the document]
nahueespinosa: > We can drop layout reversal for rolling windows and get it in reverse.

I don't think that's necessary, and it makes sense to me that we call push_front in rolling windows.

> But again, I don't mind strongly about it.

The only thing I'm thinking is that `control_type` looks nice and consistent with `state_type` IMO. But I'll leave it up to you.

nahueespinosa (approved):
LGTM!

beluga/include/beluga/motion/differential_drive_model.hpp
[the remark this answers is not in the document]
hidmic: Switched to `control_type`.

nahueespinosa (approved):

PR #322, opened 2024-02-16
Add uniform distribution with bounding regions

nahueespinosa:
### Proposed changes

A follow-up to #319, implementing multivariate uniform distributions constrained by a bounding region. This is required for landmark maps support.

Related to #279.

#### Type of change

### Checklist


beluga/include/beluga/random/multivariate_uniform_distribution.hpp
```diff
@@ -0,0 +1,193 @@
...
+template <class T, class Constraint>
+class MultivariateUniformDistribution;
+
+/// Struct representing an axis-aligned bounding region in 2D space.
+struct BoundingRegion2d {
+  double x_min;  ///< Minimum x coordinate of the map.
+  double x_max;  ///< Maximum x coordinate of the map.
+  double y_min;  ///< Minimum y coordinate of the map.
+  double y_max;  ///< Maximum y coordinate of the map.
+};
```
hidmic, line 46: @nahueespinosa meta: should we be using [`Eigen::AlignedBox`](https://eigen.tuxfamily.org/dox/classEigen_1_1AlignedBox.html) instead? Same for 3D bounding boxes.

beluga/test/beluga/random/test_multivariate_uniform_distribution.cpp
```diff
@@ -26,19 +26,48 @@
...
-TEST(UniformFreeSpaceGridDistribution, SingleSlot) {
+TEST(MultivariateUniformDistribution, BoundingRegion2d) {
+  constexpr double kTolerance = 0.01;
+  auto region = beluga::BoundingRegion2d{};
+  region.x_min = 3.00;
+  region.x_max = region.x_min + kTolerance;
+  region.y_min = -2.00;
+  region.y_max = region.y_min + kTolerance;
+  auto distribution = beluga::MultivariateUniformDistribution{region};
+  auto engine = std::mt19937{std::random_device()()};
```
hidmic, line 39: @nahueespinosa meta: should we fix the seed for tests to be reproducible? Same elsewhere.

hidmic (commented):
Awesome work @nahueespinosa ! Left a couple questions.

beluga/include/beluga/random/multivariate_uniform_distribution.hpp
nahueespinosa: Of course! Done in https://github.com/Ekumen-OS/beluga/pull/322/commits/56abf62e4755af5bec6d5aa192bbec25f024b1c2.

beluga/test/beluga/random/test_multivariate_uniform_distribution.cpp
nahueespinosa: I don't think so. I found this article related to the usage of seeds in ML projects: https://misailo.cs.illinois.edu/papers/icst22.pdf

I found the recommendations useful:
- Use fixed seeds only for tests checking for exact reproducibility of some functionality in their code.
- Randomize and log the seeds in other tests for nondeterministic algorithms to allow both reproducibility of failures and diverse executions.
- Use Test Re-runs on failure instead of setting seeds to mitigate random failures.
- Determine optimal test settings such as hyper-parameters for the algorithm(s) under test and/or assertion bounds to minimize flakiness.

I think we should treat flakiness as a bug. Logging the used seed on failure is a nice feature, but it exceeds the scope of this PR.

hidmic: Interesting read. I'm satisfied with the reasoning.

> Logging the used seed on failure is a nice feature, but it exceeds the scope of this PR.

Fair enough. I suspect we'll add it when and if this comes back to haunt us.

hidmic (approved):
LGTM

PR #324, opened 2024-02-17
Relocate `make_random_state()` method

hidmic:
### Proposed changes

Connected to #279. Follow-up to #322. This is a long due change, and as I was reworking sensor models I figured it was appropriate. This patch removes `make_random_state()` from sensor models. As we still support mixins, it relocates `make_random_state()` to the `RandomStateGenerator` mixin, which now takes both state and map as template parameters.

#### Type of change

💥 **Breaking change!** Sensor models no longer have nor require a `make_random_state()` method.

### Checklist

### Additional comments

I'm not super fond of these changes to `RandomStateGenerator`, but they are not unreasonable in context and it is arguably a temporary workaround. 

We also have to update https://github.com/Ekumen-OS/beluga-demos to match.


nahueespinosa (approved):
:raised_hands: LGTM!

hidmic:
Hmm, I suspect coverage worsened because this patch removes more code than it adds (and thus it changed proportions). How do you want to go about it @nahueespinosa?

nahueespinosa:
@hidmic Let's go ahead and merge. I don't see anything standing out in the coverage reports.

PR #326, opened 2024-02-17
Use AlignedBox3d as LandmarkMapBoundaries

hidmic:
### Proposed changes

Follow-up to #322 and #324. This patch turns `LandmarkMapBoundaries` into an `Eigen::AlignedBox3d` alias.

#### Type of change

💥 **Breaking change!** This will allow us to use `LandmarkMap` limits as constraint for `MultivariateUniformDistribution` instances.

### Checklist

### Additional comments

We have to update https://github.com/Ekumen-OS/beluga-demos too.


nahueespinosa (approved):
LGTM

PR #329, opened 2024-02-25
Fix landmark models probability aggregation formula

glpuga:
### Proposed changes

The sum-of-cubes-plus-1 formula works really bad for landmark models, causing them almost not throw particles that don't really fit what the sensors see, because the high and low values of the probability is not really very different if the number of landmarks is low.

#### Type of change

### Checklist


nahueespinosa (approved):
LGTM

Issue #309, opened 2024-02-09
Add nD-NDT maps to Beluga


hidmic:
### Feature description

Connected to #93. Before we can implement NDT based sensor models in Beluga, we need an efficient data structure for NDT maps of arbitrary dimension.

### Implementation considerations

 https://github.com/cogsys-tuebingen/cslibs_ndt may be of interest.

serraramiro1:
What about https://github.com/Tessil/robin-map ? This is the library `Kiss-ICP` uses and I had very good results with.
It contains `std::unordered_map` drop-in replacements depending on your needs.

Here're some benchmarks:
https://tessil.github.io/2016/08/29/benchmark-hopscotch-map.html

serraramiro1:
@hidmic 
I'm not completely sure if we need spatial indexing here.

Spatial indexing is usually good for complex queries, like intersections between complex shapes.

In this case we can leverage the map being discretized.
I'm thinking of:

NDT map represented as `Discretized cell -> Normal Distribution` hash map.

Then for getting likelihood for a point `{x, y}` is as easy as 
```
discretized_cell = {floor(x / resolution), floor(y / resolution)};
if discretized_cell not in map:
      return MINIMUM_LIKELIHOOD
return map[discretized_cell].likelihood(x,y)
```
Hashing integers is almost free and I don't think doing spatial indexing will outperform this. 

I might be overlooking something tho.

Thoughts?

hidmic:
IIRC NDT maps also use some shifted copies to deal with boundary effects, but yeah, that can be sorted out with cell hashes. Eventually we may want to be smart about how we run over the pointcloud to maximize cache efficiency (hash maps are O(1) but memory I/O is not). I'll try and take a look at at `robin_map` to see what it can offer in that sense. For now, the approach makes sense to me if @nahueespinosa is OK with the extra dependency (has is it been released anywhere already?).

nahueespinosa:
If the idea is to use the drop-in replacements for `std::unordered_map`, then I would support implementing a class with the necessary accessor methods where the underlying map type is a template parameter. This way we don't have to add the extra dependency to `beluga` core, and instead defer that to the concrete implementation packages.

hidmic:
As of #336 and https://github.com/Ekumen-OS/beluga/pull/338, we kinda have a first 2D variant. I'll keep this one open until we land a 3D variant.

serraramiro1:
The 3D variant was introduced with #422

PR #335, opened 2024-03-12
Add HDF5 dependency to beluga core

serraramiro1:
### Proposed changes
Needed to work in #309 as our NDT generated maps are in HDF5 format. 

#### Type of change

### Checklist

### DIFF
--- a/beluga/package.xml
+++ b/beluga/package.xml
@@ -15,6 +15,7 @@
   <buildtool_depend>cmake</buildtool_depend>
 
   <depend>eigen</depend>
+  <depend>libhdf5-dev</depend>
   <depend>range-v3</depend>
   <depend>sophus</depend>
   <depend>tbb</depend>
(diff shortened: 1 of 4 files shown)


beluga/package.xml
```diff
@@ -15,6 +15,7 @@
   <buildtool_depend>cmake</buildtool_depend>
 
   <depend>eigen</depend>
+  <depend>libhdf5-dev</depend>
```
nahueespinosa, line 18: Is this not available in Focal or something? :smiling_face_with_tear:

serraramiro1: It is! It seems like the old version doesn't export the `hdf5::*` variables, but we should be good to go using the other one I just switched to.
https://cmake.org/cmake/help/latest/module/FindHDF5.html

serraramiro1: It is! https://packages.ubuntu.com/search?keywords=libhdf5-dev

I used another variable to link against the libraries based on:
https://cmake.org/cmake/help/latest/module/FindHDF5.html
That seems to be part of all the distributed versions.

nahueespinosa: Sweet!

nahueespinosa (approved):
LGTM

Issue #153, opened 2023-03-29
Investigate why the scan likelihood formula does not follow the model in Prob. Rob.


glpuga:
## Description

Both QuickMCL and AMCL seem to have for a long time used a formula for aggregation of the likelihood `p` that does not match the model proposed in Probabilistic Robotics (Table 6.3).

The explanation seems to be only based on the empiric evidence that "it works". See

- https://github.com/VorpalBlade/quickmcl/blob/955a148806196fd6f48dd7aa80bf6d92f4e3cbce/src/quickmcl/map_likelihood.cpp#L99-L102
- https://github.com/ros-planning/navigation2/blob/81d7056f73235573407cb2ea833dbbec275fe986/nav2_amcl/src/sensors/laser/likelihood_field_model.cpp#L121-L124
- Probabilistic Robotics, Table 6.3

For the sake of equivalence, we currently implement this formula in Beluga too.

However, we should investigate how performance changes if the Prob.Rob. formula is used instead, and what the relative merits of each option are.

## Definition of done

- A baseline performance is captured against a large enough dataset with the same formula used by QuickMCL and AMCL.
- The experiment is repeated, using instead the aggregation formula proposed by the model as described in Prob. Robotics.
- The differences in performance are analyzed to evaluate the options.

hidmic:
I will eventually get back to this and try out [Log-PF](https://www.hindawi.com/journals/jece/2018/5763461/) ideas. Money on a numerical issue with likelihood functions and floating point representations.

PR #338, opened 2024-03-25
Add NDT sensor model

serraramiro1:
### Proposed changes

Adds a new sensor model, based on NDT transforms, that performs importances weighting by transforming the measurements

#### Type of change

### Checklist

### Additional comments


serraramiro1:
@hidmic PTAL when possible, it's still a draft but might be good to get the review process going.

A few concerns I have.

Citing the NDT MCL paper:
> In practice we approximate Eq. 4
by finding the cell from the map that corresponds to the mean
µ¯i = Rkµi + tk and search the local neighborhood in order
to find the closest normal distribution to µ¯i 

I'm not doing nearest neighbor search here, but we could preprocess the grid to make that work or modify the conversion grid <-> NDT map script to take that into account, likelihoods fall so abruptly when you go away from the mean that I don't expect it to matter much but we would certainly have some boundary effects with the current approach.

The way they're suggesting the scaling of the likelihood with `d1` and `d2` I don't love.
Likelihoods are not bounded, and can be any positive number. The current version would favor perfect hits to narrower NDT cells to perfect hits to cells with bigger covariances..

I'd like to try a different version where

```
weight = unscaled_likelihood_at(measurement) / unscaled_measurement_at(mean)
```
This way, our weights would have an upper bound of 1 and would make our lives easier. 

Also I'd consider using the log likelihood instead of the likelihood to make the weighting function less spiky and prevent us from giving poor scores to very-close hits to a narrow distribution.

hidmic:
> I'm not doing nearest neighbor search here, but we could preprocess the grid to make that work or modify the conversion grid <-> NDT map script to take that into account, 

IIUC nearest neighbors here is just a performance optimization. I do wonder about the detour in the algorithm from what [Sarineen et. al.](https://ieeexplore.ieee.org/document/6696380) propose. We are not applying NDT to the measurement like [Biber et. al.](https://ieeexplore.ieee.org/document/1249285) describe (i.e. discretizing space, fitting normal distributions within each cell) before computing likelihoods. Why?

> likelihoods fall so abruptly when you go away from the mean that I don't expect it to matter much but we would certainly have some boundary effects with the current approach.

And that can be problematic, numerically speaking. Working with degenerate, point-mass distributions won't make it any better (like we do here instead of using NDT in full). Even working with NDT may not be enough. Both Biber et. al. and [Schulz et. al.](https://ieeexplore.ieee.org/abstract/document/8593602) suggest having multiple map layers, offseted to deal with boundary effects. We are not doing that either.

> The way they're suggesting the scaling of the likelihood with d1 and d2 I don't love.

Yeah, there's some handwaving to those scalings. $d_2$ I can understand though. By artificially inflating covariances you can compensate for the limited sample size issues (i.e. the issues that particle filters when distributions are too narrow).

> Likelihoods are not bounded, and can be any positive number. The current version would favor perfect hits to narrower NDT cells to perfect hits to cells with bigger covariances..

I don't follow you here.

> This way, our weights would have an upper bound of 1 and would make our lives easier.

We do [`normalize`](https://github.com/Ekumen-OS/beluga/blob/main/beluga/include/beluga/actions/normalize.hpp) though, typically after reweighting.

> Also I'd consider using the log likelihood instead of the likelihood to make the weighting function less spiky and prevent us from giving poor scores to very-close hits to a narrow distribution.

That's something [I've been wanting to try](https://github.com/Ekumen-OS/beluga/issues/153#issuecomment-1936583340) for a while. I have a strong suspicion that much of the heuristic that the other sensor models rely on when computing likelihoods can be traced back to numerical issues. Log likelihood and log odds representation may help. I would defer to a different PR though.

beluga/include/beluga/sensor/ndt_sensor_model.hpp
```diff
@@ -0,0 +1,234 @@
...
+  Eigen::Matrix2Xd centered = points_view.colwise() - mean;
+  // Use sample covariance.
+  Eigen::Matrix2Xd cov = (centered * centered.transpose()) / static_cast<double>(points_view.cols() - 1);
+  cov(0, 0) = std::max(cov(0, 0), kMinVariance);
+  cov(1, 1) = std::max(cov(1, 1), kMinVariance);
+  return NDTCell{mean, cov};
+}
+
+/// Ostream overload mostly for debugging purposes.
+inline std::ostream& operator<<(std::ostream& os, const NDTCell& cell) {
```
hidmic, line 98: @serraramiro1 consider relocating operator overloads to class / struct scope as inline friend functions, so as to limit them to ADL resolution (which means less verbosity from the compiler when an error occurs).

beluga/include/beluga/sensor/ndt_sensor_model.hpp
```diff
@@ -0,0 +1,234 @@
...
+
+  /// Returns the L2 likelihood scaled by 'd1' and 'd2' set in the parameters for this instance for 'measurement, or
+  /// 'params_.min_likelihood' if the cell corresponding to 'measurement' doesn't exist in the map.
+  [[nodiscard]] double likelihood_at(const NDTCell& measurement) const {
+    const auto maybe_cell = cells_data_.data_near(measurement.mean);
+    if (!maybe_cell.has_value()) {
+      return params_.minimum_likelihood;
+    }
+    return maybe_cell->likelihood_at(measurement, params_.d1, params_.d2);
+  }
```
hidmic, line 182: @serraramiro1 does this need to be a public method? Does it need to be a sensor model method?

beluga/test/beluga/sensor/test_ndt_model.cpp
```diff
@@ -0,0 +1,142 @@
...
+  std::vector map_data{
+      Eigen::Vector2d{0.1, 0.2},  Eigen::Vector2d{0.112, 0.22}, Eigen::Vector2d{0.15, 0.23},
+      Eigen::Vector2d{0.1, 0.24}, Eigen::Vector2d{0.16, 0.25},  Eigen::Vector2d{0.1, 0.26},
+  };
+  auto cells = to_cells(map_data, map_res);
+
+  typename sparse_grid_t::map_type map_cells_data;
+  for (const auto& cell : cells) {
+    map_cells_data[(cell.mean.array() / map_res).cast<int>()] = cell;
+    std::cerr << cell;
```
hidmic, line 127: @serraramiro1 nit: leftover?

beluga/include/beluga/sensor/ndt_sensor_model.hpp
```diff
@@ -0,0 +1,234 @@
...
+  /// Returns a state weighting function conditioned on 2D lidar hits.
+  /**
+   * \param points 2D lidar hit points in the reference frame of particle states.
+   * \return a state weighting function satisfying \ref StateWeightingFunctionPage
+   *  and borrowing a reference to this sensor model (and thus their lifetime are bound).
+   */
+  [[nodiscard]] auto operator()(measurement_type&& points) const {
+    return [this, cells = to_cells(points, cells_data_.resolution())](const state_type& state) -> weight_type {
+      return std::transform_reduce(
+          cells.cbegin(), cells.cend(), 1.0, std::plus{},
```
hidmic, line 169: @serraramiro1 why do we start at 1?

hidmic (commented):
Second pass. Overall looks reasonable.

beluga/include/beluga/sensor/ndt_sensor_model.hpp
```diff
@@ -0,0 +1,234 @@
...
+
+/// Loads a sparse grid from an hdf5 file, with the following structure:
+// - "resolution": () resolution for the discrete grid (cells are resolution x
+//   resolution m^2 squares).
+// - "cells": (NUM_CELLS, 2) that contains the cell coordinates.
+// - "means": (NUM_CELLS, 2) that contains the 2d mean of the normal distribution
+//   of the cell.
+// - "covariances": (NUM_CELLS, 2, 2) Contains the covariance for each cell.
+template <typename SparseGridT>
+SparseGridT load_from_hdf5(const std::filesystem::path& path_to_hdf5_file) {
```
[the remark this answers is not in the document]
serraramiro1: I've made a few modifications, PTAL.
I didn't move it to the sparse grid header because it contains NDT specific details.

beluga/include/beluga/sensor/ndt_sensor_model.hpp
serraramiro1: Nice one. Done.

beluga/include/beluga/sensor/ndt_sensor_model.hpp
serraramiro1: It makes testing way easier and thorough, in my opinion.

I'm down to make it private + friend test if needed but I'd rather keep some version of it.

beluga/include/beluga/sensor/ndt_sensor_model.hpp
serraramiro1: I just copied this from `likelihood_sensor_model.hpp`, I figured starting from 0 might cause weird normalization / recovery behaviors.

WDYT?

serraramiro1:
@hidmic ready to re-review.
Note that I couldn't fully extend this to 3D because of sparse grids being only-2D.
Making them support NDimensional data would involve going back to `BaseGrid2` and I rather not go that deep for now.

I'll create a ticket for this effort and leave it here if you're OK with it.

beluga/include/beluga/sensor/ndt_sensor_model.hpp
hidmic: > I figured starting from 0 might cause weird normalization / recovery behaviors.

Why? It's funny, you can trace this back to 20yo AMCL code, which gives no reasons for these deviations. 

I'd rather not propagate it unless we have a good reason to.

beluga/test/beluga/sensor/data/test_ndt_cell.cpp
```diff
@@ -0,0 +1,73 @@
+// Copyright 2023 Ekumen, Inc.
```
hidmic, line 1: @serraramiro1 nit: `2023` -> `2024`, here and elsewhere.

beluga/test/beluga/sensor/data/test_ndt_cell.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include <sophus/se2.hpp>
+
+namespace beluga {
+
+namespace {
+Eigen::Matrix2Xd get_diagonal_covariance(double x_var = 0.5, double y_var = 0.8) {
+  return Eigen::Vector2d{x_var, y_var}.asDiagonal();
+}
+
+}  // namespace
```
hidmic, line 34: @serraramiro1 nit: 
```suggestion
namespace {

Eigen::Matrix2Xd get_diagonal_covariance(double x_var = 0.5, double y_var = 0.8) {
  return Eigen::Vector2d{x_var, y_var}.asDiagonal();
}

}  // namespace
```

beluga/include/beluga/sensor/ndt_sensor_model.hpp
```diff
@@ -0,0 +1,244 @@
...
+ * \tparam SparseGridT Type representing a sparse NDT grid, as a specialization of 'beluga::SparseValueGrid'.
+ */
+template <typename SparseGridT>
+class NDTSensorModel {
+ public:
+  static_assert(std::is_same_v<SparseValueGrid<typename SparseGridT::map_type>, SparseGridT>);
+  /// NDT Cell type.
+  using ndt_cell_type = typename SparseGridT::mapped_type;
+  static_assert(
+      ndt_cell_type::num_dim == 2 || ndt_cell_type::num_dim == 3,
```
hidmic, line 129: @serraramiro1 should we constrain this to 2D only until the rest of the implementation catches up?

hidmic (approved):
Overall LGTM. The coverage check is in a weird state.

beluga/test/beluga/sensor/data/test_ndt_cell.cpp
serraramiro1: Addressed

beluga/test/beluga/sensor/data/test_ndt_cell.cpp
serraramiro1: Done

PR #342, opened 2024-04-12
Remove sensor model variant from AMCL pipeline

serraramiro1:
### Proposed changes

_Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request.
If it fixes a bug or resolves a feature request, be sure to link to that issue._

#### Type of change

### Checklist

### DIFF
--- a/beluga_ros/include/beluga_ros/amcl.hpp
+++ b/beluga_ros/include/beluga_ros/amcl.hpp
@@ -15,6 +15,8 @@
 #ifndef BELUGA_ROS_AMCL_HPP
 #define BELUGA_ROS_AMCL_HPP
 
+#include <beluga/sensor/beam_model.hpp>
+#include <beluga/sensor/likelihood_field_model.hpp>
 #include <optional>
 #include <utility>
 #include <variant>
@@ -26,6 +28,17 @@
 #include <range/v3/view/take_exactly.hpp>
 
 namespace beluga_ros {
+
+using sensor_model_variant = std::variant<
+    beluga::LikelihoodFieldModel<beluga_ros::OccupancyGrid>,  //
+    beluga::BeamSensorModel<beluga_ros::OccupancyGrid>>;
+
+using motion_model_variant = std::variant<
+    beluga::DifferentialDriveModel,     //
+    beluga::OmnidirectionalDriveModel,  //
+    beluga::StationaryModel>;
+
+using execution_policy_variant = std::variant<std::execution::sequenced_policy, std::execution::parallel_policy>;
 
 /// Struct containing parameters for the Adaptive Monte Carlo Localization (AMCL) implementation.
 struct AmclParams {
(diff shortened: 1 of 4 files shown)


serraramiro1:
@hidmic With this and unifying the measurement types we could add the NDT sensor model to the `sensor_model_variant` and reuse this piece of code for the NDT pipeline (after implementing `  beluga::MultivariateUniformDistribution<Sophus::SE2d, NDTMapT>`).

LMK what you think

beluga_ros/include/beluga_ros/amcl.hpp
```diff
@@ -45,33 +58,32 @@ struct AmclParams {
...
-      beluga::OmnidirectionalDriveModel,  //
-      beluga::StationaryModel>;
+  using sensor_model_type = SensorModel;
 
-  using sensor_model_variant = std::variant<
-      beluga::LikelihoodFieldModel<beluga_ros::OccupancyGrid>,  //
-      beluga::BeamSensorModel<beluga_ros::OccupancyGrid>>;
-
-  using execution_policy_variant = std::variant<std::execution::sequenced_policy, std::execution::parallel_policy>;
+  using map_type = typename sensor_model_type::map_type;
```
hidmic, line 73: @serraramiro1 meta: do this two need to be coupled like this? I mean, there will be a constraint between the two, but do we have to derive it like this instead of allowing the user to set both?

beluga_ros/include/beluga_ros/amcl.hpp
```diff
@@ -45,33 +58,32 @@ struct AmclParams {
...
+   * \param sensor_model An instance of the sensor model.
    * \param params Parameters for AMCL implementation.
    * \param execution_policy Variant of execution policy.
    */
   Amcl(
-      beluga_ros::OccupancyGrid map,
+      map_type map,
       motion_model_variant motion_model,
-      sensor_model_variant sensor_model,
+      sensor_model_type sensor_model,
```
hidmic, line 86: @serraramiro1 meta: I get the idea of bubbling up variants to calling code and keeping this class fully statically typed instead. What I'm not very fond of is the mix of having some of it fully statically typed (i.e. the sensor model) and some of it using algebraic data types (i.e. the motion model). IMO it should be either or but not both.
Of course, if you go fully statically typed you will have a combinatorial explosion in your `AnyAMCL` variant definition. 

All that to say that I'm not entirely convinced this is the right path (but I can be swayed too). Have you considered introducing a "filter initializer" object instead of passing the map to the filter constructor? Have you considered templating the variant types themselves? If you are worried about enforcing constraints between types, that can be arranged.

serraramiro1: You're totally right.
Based on the offline sync we had:

- I'll create a new fully typed generic pipeline that can live in `beluga`
- Create a ticket to consolidate the `beluga_ros` pipeline with the new one in beluga

PR #343, opened 2024-04-17
Add a core AMCL pipeline

serraramiro1:
### Proposed changes

Adds a fully typed generic AMCL pipeline that can be used with any motion model, sensor model, random state generator, etc.

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/algorithm/amcl_core.hpp
+++ b/beluga/include/beluga/algorithm/amcl_core.hpp
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef AMCL_CORE_HPP
-#define AMCL_CORE_HPP
+#ifndef BELUGA_ALGORITHM_AMCL_CORE_HPP
+#define BELUGA_ALGORITHM_AMCL_CORE_HPP
 
 #include <execution>
 #include <optional>
@@ -30,7 +30,6 @@
 namespace beluga {
 
 /// Struct containing parameters for the Adaptive Monte Carlo Localization (AMCL) implementation.
-
 struct AmclParams {
   /// Min distance in meters between updates.
   double update_min_d = 0.25;
@@ -78,7 +77,6 @@ template <
     typename WeightT = beluga::Weight,
     class ParticleType = std::tuple<typename SensorModel::state_type, WeightT>,
     class ExecutionPolicy = std::execution::sequenced_policy>
-
 class Amcl {
   static_assert(
       std::is_same_v<ExecutionPolicy, std::execution::parallel_policy> or
@@ -159,29 +157,29 @@ class Amcl {
    * weights are adjusted accordingly. Also, according to the configured resampling policy, the particles
    * are resampled to maintain diversity and prevent degeneracy.
    *
-   * \param state_at_measurement State at the measurement time.
+   * \param control_action Control action.
    * \param measurement Measurement data.
    * \return An optional pair containing the estimated pose and covariance after the update,
    *         or std::nullopt if no update was performed.
    */
-  auto update(state_type state_at_measurement, measurement_type measurement) -> std::optional<estimation_type> {
+  auto update(state_type control_action, measurement_type measurement) -> std::optional<estimation_type> {
     if (particles_.empty()) {
       return std::nullopt;
     }
 
-    if (!update_policy_(state_at_measurement) && !force_update_) {
+    if (!update_policy_(control_action) && !force_update_) {
       return std::nullopt;
     }
 
-    particles_ |= beluga::actions::propagate(
-                      execution_policy_, motion_model_(control_action_window_ << state_at_measurement)) |  //
-                  beluga::actions::reweight(execution_policy_, sensor_model_(std::move(measurement))) |    //
-                  beluga::actions::normalize(execution_policy_);
+    particles_ |=
+        beluga::actions::propagate(execution_policy_, motion_model_(control_action_window_ << control_action)) |  //
+        beluga::actions::reweight(execution_policy_, sensor_model_(std::move(measurement))) |                     //
+        beluga::actions::normalize(execution_policy_);
 
     const double random_state_probability = random_probability_estimator_(particles_);
 
     if (resample_policy_(particles_)) {
-      auto random_state = ranges::compose(beluga::make_from_state<particle_type>, get_random_state_fn());
+      auto random_state = ranges::compose(beluga::make_from_state<particle_type>, get_random_state_generator());
 
       if (random_state_probability > 0.0) {
         random_probability_estimator_.reset();
@@ -205,16 +203,15 @@ class Amcl {
   /// Force a manual update of the particles on the next iteration of the filter.
   void force_update() { force_update_ = true; }
 
+ private:
   /// Gets a callable that will produce a random state.
-  decltype(auto) get_random_state_fn() const {
+  decltype(auto) get_random_state_generator() const {
     if constexpr (std::is_invocable_v<random_state_generator_type>) {
       return random_state_generator_;
     } else {
       return random_state_generator_(particles_);
     }
   }
-
- private:
   beluga::TupleVector<particle_type> particles_;
 
   AmclParams params_;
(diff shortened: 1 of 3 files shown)


beluga/include/beluga/algorithm/amcl_core.hpp
```diff
@@ -0,0 +1,240 @@
...
+   * is required, the motion model and sensor model updates are applied to the particles, and the particle
+   * weights are adjusted accordingly. Also, according to the configured resampling policy, the particles
+   * are resampled to maintain diversity and prevent degeneracy.
+   *
+   * \param state_at_measurement State at the measurement time.
+   * \param measurement Measurement data.
+   * \return An optional pair containing the estimated pose and covariance after the update,
+   *         or std::nullopt if no update was performed.
+   */
+  auto update(state_type state_at_measurement, measurement_type measurement) -> std::optional<estimation_type> {
```
hidmic, line 167: @serraramiro1 meta: hmm, this should be taking a control action like generic motion models do. Using a window of state measurements is something specific to a kind of motion models. But pulling the control window out is inconvenient for the user, so perhaps this class should be specialized to include the window when a specific kind of motion model is in use. 

That said, I'm OK with waiting until we have a motion model that doesn't use a control window to implement that.

beluga/include/beluga/algorithm/amcl_core.hpp
```diff
@@ -0,0 +1,240 @@
...
+  void force_update() { force_update_ = true; }
+
+  /// Gets a callable that will produce a random state.
+  decltype(auto) get_random_state_fn() const {
+    if constexpr (std::is_invocable_v<random_state_generator_type>) {
+      return random_state_generator_;
+    } else {
+      return random_state_generator_(particles_);
+    }
+  }
```
hidmic, line 215: @serraramiro1 nit: consider making this private. Also, `get_random_state_generator()` seems more adequate.

beluga/include/beluga/algorithm/amcl_core.hpp
```diff
@@ -0,0 +1,240 @@
...
+   * state.
+   * \param spatial_hasher A spatial hasher instance capable of computing a hash out of a particle state.
+   * \param params Parameters for AMCL implementation.
+   * \param execution_policy Policy to use when processing particles.
+   */
+  Amcl(
+      MotionModel motion_model,
+      SensorModel sensor_model,
+      RandomStateGenerator random_state_generator,
+      spatial_hasher_type spatial_hasher,
```
hidmic, line 111: @serraramiro1 meta: we need to sort out the mixed casing we have for types here and there.

beluga/include/beluga/algorithm/amcl_core.hpp
```diff
@@ -0,0 +1,240 @@
...
+ * \tparam ExecutionPolicy Execution policy for particles processing.
+ */
+template <
+    class MotionModel,
+    class SensorModel,
+    class RandomStateGenerator,
+    typename WeightT = beluga::Weight,
+    class ParticleType = std::tuple<typename SensorModel::state_type, WeightT>,
+    class ExecutionPolicy = std::execution::sequenced_policy>
+
```
hidmic, line 81: @serraramiro1 nit: extra blank?

beluga/include/beluga/algorithm/amcl_core.hpp
```diff
@@ -0,0 +1,240 @@
...
+ * - t can be called with `const beluga::TupleVector<ParticleType>&> returning a callable
+ *   that can be called without arguments and return an instance of 'SensorModel::state_type'.
+ * \tparam WeightT Type to represent a weight of a particle.
+ * \tparam ParticleType Full particle type, containing state, weight and possibly
+ * other information .
+ * \tparam ExecutionPolicy Execution policy for particles processing.
+ */
+template <
+    class MotionModel,
+    class SensorModel,
```
hidmic, line 76: @serraramiro1 meta: as an afterthought, we could hide runtime type selection in perfect forwarding models. E.g. have a template motion model that takes a variant of motion models and forwards method and operator invocations using `std::visit()`. We would need some extra wrappers to deal with some lambda returns, but that's perfectly doable.

Then, this AMCL implementation would be the only one.

hidmic (commented):
I like it :100:

beluga/include/beluga/algorithm/amcl_core.hpp
serraramiro1: I see, I think I got confused by this taking in a `state_type` instead of what seems to be a more generic `control_type`.
DYT we should exclude the rolling window here as it seems model-specific and take in a `control_type`?

hidmic: Yes but it's kinda inconvenient because then you have to store the window elsewhere. The nice thing would be to have a specialization that includes the window for the motion model that use one. It's some work though. I won't argue against leaving the TODO behind.

serraramiro1: Also just noticed that the `on_motion` policy needs a state (at least for the current motion models). So even if we pass a `control_type` we'd need a way to extract the pose out of it :-1:

hidmic: Well, `on_motion` could integrate control actions or AMCL could track prior state estimates. There is room for generalization, but we don't have to generalize now.

serraramiro1: Done!

beluga/include/beluga/algorithm/amcl_core.hpp
serraramiro1: Yeah, our `snake_case` class names is all over the place.
I'll leave this as is for now if that's OK

beluga/include/beluga/algorithm/amcl_core.hpp
serraramiro1: Done!

beluga/include/beluga/algorithm/amcl_core.hpp
```diff
@@ -0,0 +1,237 @@
...
+    initialize(beluga::MultivariateNormalDistribution{pose, covariance});
+  }
+
+  /// Update the map used for localization.
+  void update_map(map_type map) { sensor_model_.update_map(std::move(map)); }
+
+  /// Update particles based on motion and sensor information.
+  /**
+   * This method performs a particle filter update step using motion and sensor data. It evaluates whether
+   * an update is necessary based on the configured update policy and the force_update flag. If an update
```
hidmic, line 155: @serraramiro1 nit^2: consider 
```suggestion
   * an update is necessary based on the configured update policy. If an update is required or if it was forced via `force_update()`, ...
```
instead of discussing private state.

hidmic (approved):
LGTM but for a few last comments.

beluga/test/beluga/algorithm/test_amcl_core.cpp
```diff
@@ -0,0 +1,180 @@
...
+#include "beluga/sensor/beam_model.hpp"
+#include "beluga/test/static_occupancy_grid.hpp"
+#include "beluga/views/particles.hpp"
+
+#include <gmock/gmock.h>
+#include <range/v3/view/take_last.hpp>
+#include <sophus/se2.hpp>
+
+using beluga::testing::StaticOccupancyGrid;
+namespace beluga {
```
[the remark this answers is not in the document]
serraramiro1: Done!

beluga/include/beluga/algorithm/amcl_core.hpp
```diff
@@ -0,0 +1,237 @@
...
+    if (particles_.empty()) {
+      return std::nullopt;
+    }
+
+    if (!update_policy_(control_action) && !force_update_) {
+      return std::nullopt;
+    }
+
+    particles_ |=
+        beluga::actions::propagate(execution_policy_, motion_model_(control_action_window_ << control_action)) |  //
```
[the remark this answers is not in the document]
serraramiro1: Done!

beluga/include/beluga/algorithm/amcl_core.hpp
serraramiro1: Done!

PR #347, opened 2024-05-06
Add NDT AMCL node and tests.

### LINKED ISSUE #312: Implement 2D NDT-MCL ROS 2 node
### Feature description

Connected to #93. Depends on #311. Precisely what the title says. Implement a ROS 2 node for 2D localization by feeding laser scans and a map to a 2D-NDT sensor model. Use a planar lidar on a Kobuki robot (in simulation or in the real-world) to test it.

serraramiro1:
### Proposed changes

Adds a new node to leverage the NDT sensor model.\
Closes #312 
#### Type of change

### Checklist


beluga_amcl/src/ndt_amcl_node.cpp
```diff
@@ -0,0 +1,915 @@
...
+void NdtAmclNode::timer_callback() {
+  if (!particle_filter_) {
+    return;
+  }
+
+  if (particle_cloud_pub_->get_subscription_count() == 0) {
+    return;
+  }
+
+  // Particle weights from the filter may or may not be representative of the
```
hidmic, line 691: @serraramiro1 @glpuga nit: we should relocate this particle cloud computation to `beluga_ros`.

beluga_amcl/test/test_ndt_amcl_node.cpp
```diff
@@ -0,0 +1,734 @@
...
+    tester_node_->create_particle_cloud_subscriber();
+    return spin_until(
+        [this] { return tester_node_->latest_particle_cloud().has_value(); }, 1000ms, ndt_amcl_node_, tester_node_);
+  }
+
+ protected:
+  std::shared_ptr<AmclNodeUnderTest> ndt_amcl_node_;
+  std::shared_ptr<TesterNode> tester_node_;
+};
+
```
hidmic, line 295: @serraramiro1 nit: consider relocating all test helpers to a separate header so we can reuse them for regular AMCL and NDT based AMCL tests.

beluga_amcl/src/ndt_amcl_node.cpp
```diff
@@ -0,0 +1,915 @@
...
+
+constexpr std::string_view kDifferentialModelName = "differential_drive";
+constexpr std::string_view kOmnidirectionalModelName = "omnidirectional_drive";
+constexpr std::string_view kStationaryModelName = "stationary";
+constexpr std::string_view kNav2DifferentialModelName = "nav2_amcl::DifferentialMotionModel";
+constexpr std::string_view kNav2OmnidirectionalModelName = "nav2_amcl::OmniMotionModel";
+
+}  // namespace
+
+NdtAmclNode::NdtAmclNode(const rclcpp::NodeOptions& options)
```
hidmic, line 65: @serraramiro1 meta: it would seem like there is some room to deduplicate vast amounts of code between this node and the regular AMCL one with little effort. Many methods are verbatim copies.

hidmic (commented):
First pass. CI does not look happy.

serraramiro1:
> First pass. CI does not look happy.

Although this is a regression in terms of coverage, I feel like the line coverage looks sane. 
Do you mind TAL and providing insight whether the current state is enough or not?

I feel like >90% is good enough for nodes

beluga_amcl/src/ndt_amcl_node.cpp
hidmic: I meant the whole computation. We have copies of this in several places.

beluga_amcl/src/ndt_amcl_node.cpp
hidmic: I like the changes, but I will say we still have a lot of duplicate code though. I'm OK with deferring to a follow-up ticket if we tackle this right after higher priority stuff (instead of forgetting about it).

hidmic:
> Do you mind TAL and providing insight whether the current state is enough or not?

If `codecov` line coverage report is accurate, a test with non-empty laserscan and a non-empty particle cloud would exercise the missing code paths. You could trick it by de-duplicating code too.

> I feel like >90% is good enough for nodes

I'm reluctant to let this one slip.

beluga_amcl/src/ndt_amcl_node.cpp
serraramiro1: Done.

beluga_amcl/src/ndt_amcl_node.cpp
serraramiro1: Created https://github.com/Ekumen-OS/beluga/issues/355

beluga_amcl/include/beluga_amcl/common.hpp
```diff
@@ -0,0 +1,466 @@
+// Copyright 2024 Ekumen, Inc.
```
hidmic, line 1: @serraramiro1 nit: perhaps name this file so it is clear it is ROS 2 specific?

beluga_amcl/test/cmake/ament.cmake
```diff
@@ -14,6 +14,20 @@
 
 find_package(ament_cmake_gmock REQUIRED)
 
+file(COPY test_data DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
```
hidmic, line 17: @serraramiro1 nit: did you consider passing the path to test data as a compiler definition?

hidmic (approved):
LGTM though I still see a sensible decrease in test coverage.

beluga_amcl/include/beluga_amcl/common.hpp
serraramiro1: Done

beluga_amcl/test/cmake/ament.cmake
serraramiro1: I didn't know that was a thing, tbh. 
I don't see a lot of value on changing this and this is the way it's being done in `beluga`, so I'll keep it as-is.

serraramiro1: I didn't know that was a thing, tbh. 
I don't see a lot of value on changing this and this is the way it's being done in `beluga`, so I'll keep it as-is.

hidmic:
> /home/runner/work/beluga/beluga/beluga_amcl/include/beluga_amcl/ndt_amcl_node.hpp:91: error: Compound beluga_amcl::NdtAmclNode is not documented. (warning treated as error, aborting now)

build-docs job isn't happy :)

PR #352, opened 2024-05-07
Clang tidy doesn't sort includes

serraramiro1:
### Proposed changes
The way `clang-format` sorts includes conflicts with `clang-tidy`'s, and can only be fixed by the user including empty lines between include blocks in a non intuitive manner.

This PR removes the sorting behavior of `clang-tidy` and will delegate that functionality solely to `clang-format`.
#### Type of change

### Checklist

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -7,7 +7,6 @@ Checks: >
   bugprone-*,
   google-*,
   llvm-header-guard,
-  llvm-include-order,
   misc-*,
   modernize-*,
   performance-*,


hidmic (changes requested):
The problem with this is that we don't run `clang-tidy` as a pre-commit hook (and doing so is going to be tricky with all those `build_commands.json` files). Can't we disable the offending `clang-tidy` check instead?

hidmic (approved):
LGTM. We can ignore `codecov`'s check.

PR #353, opened 2024-05-07
Allow coverage to drop by 1%

nahueespinosa:
### Proposed changes

As the title says, this reduces the amount of false positives in detecting test coverage regressions.

#### Type of change

### Checklist

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -13,12 +13,19 @@ component_management:
       - type: project  # Measure overall project coverage.
         # The minimum coverage ratio to send a success status is the base
         # commit coverage (pull request base or parent commit).
-        target: auto
-        # Allow the coverage to drop by this percentage.
-        threshold: 1%
+        target: 95%
         # If the patch coverage is 100% and there are no unexpected changes,
         # pass the project status.
         removed_code_behavior: fully_covered_patch
+
+      - type: patch  # Measure lines adjusted in the pull request.
+        # The minimum coverage ratio to send a success status is the base
+        # commit coverage (pull request base or parent commit).
+        target: auto
+        # Allow the coverage to drop by this percentage.
+        threshold: 1%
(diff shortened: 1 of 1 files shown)


.codecov.yml
```diff
@@ -14,6 +14,8 @@ component_management:
         # The minimum coverage ratio to send a success status is the base
         # commit coverage (pull request base or parent commit).
         target: auto
+        # Allow the coverage to drop by this percentage.
+        threshold: 1%
```
hidmic, line 18: @nahueespinosa can we have a strict lower bound? Like, all good with small drops but going below 95% (or 90%, whichever we prefer) is a no-no.

nahueespinosa: AIUI, looking at https://docs.codecov.com/docs/commit-status, we can't do both checks at the same time.
In https://github.com/Ekumen-OS/beluga/pull/353/commits/2d15264375cf2ea704166d374c1828aa9c618ee2, I changed the configuration to have a strict lower bound for project level coverage, and use the target branch as reference in patch level coverage (with a 1% threshold). I think this might work.

nahueespinosa: IMO, leaving only a lower limit without patch verification may result in a slow decrease towards that limit, and people having to add coverage for the old code in order to unblock themselves.

hidmic: > IMO, leaving only a lower limit without patch verification may result in a slow decrease towards that limit, and people having to add coverage for the old code in order to unblock themselves.

That's a very good point.

hidmic (approved):
LGTM though I'm not entirely sure this is working. `beluga_ros` code coverage is ~90%.

nahueespinosa:
`beluga_ros` limit is set to 90% as a special case for now, so this works as expected. Let's merge and see how it goes in other PRs.

Issue #306, opened 2024-02-09
Write tutorial on extending Beluga 


hidmic:
### Feature description

Precisely what the title says. Despite the fact extensibility is one of Beluga's major selling points, we have zero documentation on how one is supposed to do that. Any canonical dynamic system (particle, pendulum) could be used as a basis to demonstrate how algorithms and models are implemented in Beluga.

Make sure to at least cover:

- [x] A sensor model implementation
- [x] A motion model implementation
- [ ] A sampling method implementation
- [x] An estimate method implementation

hidmic:
@nahueespinosa is it time to circle back to #358 already?

nahueespinosa:
@hidmic I have no bandwidth for this, but yes, I guess so.

PR #358, opened 2024-05-13
Add guide on extending Beluga

nahueespinosa:
### Proposed changes

Partially addresses #306.

#### Type of change

### Checklist

### DIFF
--- a/docs/guides/extending-beluga.md
+++ b/docs/guides/extending-beluga.md
@@ -1,3 +1,264 @@
 # Extending Beluga
 
-Coming soon!
+## Introduction
+
+This guide demonstrates how to extend the capabilities of the Beluga library by adding custom models and algorithms. The focus will be on understanding the existing codebase and introducing minimal code examples that provide a foundation for building more complex and useful extensions.
+
+Here’s how a minimal particle filter algorithm could be structured using Beluga. For the purpose of this guide, we will ignore how the particle set is initialized, and will focus only on the core of the algorithm.
+
+```cpp
+constexpr auto kMaxParticles = 1'000;
+
+// Define and initialize key components (parameters omitted for simplicity)
+auto particles = ...;
+auto motion_model = ...;
+auto sensor_model = ...;
+
+for (auto&& [control, measurement] : datapoints) {
+
+  // Update
+  particles |=
+    beluga::actions::propagate(motion_model(control)) |
+    beluga::actions::reweight(sensor_model(measurement)) |
+    beluga::actions::normalize();
+
+  // Resampling
+  particles |=
+    beluga::views::sample |
+    ranges::views::take_exactly(kMaxParticles) |
+    beluga::actions::assign;
+
+  // Estimation
+  const auto estimate = beluga::estimate(particles);
+}
+```
+
+Any part of this code can be customized or extended by adhering to specific constraints which we will discuss. In the following sections, we will explore the customization of different components, starting with the motion and sensor models.
+
+## Motion Model
+
+In particle filter algorithms, motion models are crucial for predicting the next state of a particle based on its current state and the control inputs. By customizing the motion model, developers can tailor the library to more accurately reflect the dynamics of specific systems or environments. This section explains how motion models work and how they can be implemented and adapted to meet specific needs.
+
+### Definition
+
+A motion model, also known as a state transition model, describes the posterior distribution that a particle assumes when executing an action command or control. It is represented by the conditional density:
+
+$$
+    p(x_t | u_t, x_{t-1})
+$$
+
+Where $u_t$ is the control or action, $x_t$ is the current state of the particle, and $x_{t-1}$ is the previous state.
+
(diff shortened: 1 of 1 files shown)


docs/guides/extending-beluga.md
```diff
@@ -1,3 +1,264 @@
...
+  }
+
+ private:
+  double stddev_;
+};
+```
+
+The implementation of the `MotionModel` demonstrates a straightforward yet powerful way to integrate custom motion dynamics into the Beluga library. This library already provides implementations that can be used as reference or starting point for custom models. See for example:
+
+- `beluga::DifferentialDriveModel`: Sampled odometry model for a differential drive robot.
```
hidmic, line 119: FWIW these should be links, but I have to fix [`doxylink`](https://github.com/sphinx-contrib/doxylink) first. It chokes on Beluga code right now, it is too complex for `doxylink`'s implementation :roll_eyes:

hidmic (commented):
@nahueespinosa meta: this is great! I'm a bit divided about content organization though. There are paragraphs that seem better suited for the `Key Concepts` page, paragraphs that read like the 1D particle filter tutorial, paragraphs that discuss API usage, and paragraphs that do indeed show how to extend Beluga (and only some of its concepts).

Would it make sense to split / merge with #340? What do you have in mind?

nahueespinosa:
> There are paragraphs that seem better suited for the Key Concepts page, paragraphs that read like the 1D particle filter tutorial, paragraphs that discuss API usage, and paragraphs that do indeed show how to extend Beluga (and only some of its concepts).

@hidmic Honestly, I see that too. Here are some thoughts:

1. Model definitions do belong to the key concepts page. I can move them and link them here when appropriate.
2. I think the tutorial could be a follow-up to this guide, since you need to extend Beluga in order to implement the most basic particle filter simulation. I would like to see the tutorial using the models designed here and maybe show how tweaking certain aspects of the implementation yields different results.
3. IMO, the API usage paragraphs contain the minimum required information to follow the guide. Even if that information becomes available in another section of the documentation, I think having a refresher here isn't a bad idea.
4. When you say "only some of its concepts", do you mean that we're missing model concepts? This PR is meant to cover the first two bullet points in #306. So, let me know if there is something missing here.

hidmic:
@nahueespinosa :+1: to (1) and (3). OK with (4), and yes, that's part of what I had in mind. About (2), in my mind the distinction between a tutorial on how to use Beluga and a guide on how to extend Beluga is in the audience and purpose of each. The tutorial is useful for users that need particle filtering. The guide is useful for developers that want to contribute to Beluga or libraries derived from Beluga. The user can use an ad-hoc lambda for a motion model. The developer should use a templated functor and ideally stay general. The user could use a free function as location estimator. The developer should specialize `estimate` overloads. And so on.

nahueespinosa:
> About (2), in my mind the distinction between a tutorial on how to use Beluga and a guide on how to extend Beluga is in the audience and purpose of each.

@hidmic I think that's a fair point. In any case, we need to describe what we're about to implement here (or reference it from somewhere else at least), so the paragraphs that read like the 1D particle filter tutorial are needed. I don't have a strong opinion on whether we should keep them here. But I'm inclined to create a new page (that can expand on the key concepts) that formulates the problem of 1D particle filtering, and which this guide and tutorial can refer to. WDYT?

hidmic:
@nahueespinosa Hmm, thinking out loud, someone reading the guide is likely to have read the tutorial first (or at least that's the logical progression). Would it make sense for the guide to refer to the tutorial? Like a follow-up document, discussing how one would go about improving and generalizing the 1D particle filter to be library material. I see a lot of synergy between @Alondruck's patch and yours there. WDYT?

nahueespinosa:
@hidmic Sure, we could establish that dependency and block this until the tutorial lands.  For now, I'll create another PR adding a few more key concepts.

hidmic:
Is this ready for a rebase?

nahueespinosa:
> Is this ready for a rebase?

@hidmic We still need to move the key-concept content out of this section. Also waiting for the tutorial to land so this extension guide can refer to it.

docs/guides/extending-beluga.md
```diff
@@ -1,3 +1,264 @@
...
+The resulting implementation could look like this:
+
+```cpp
+struct SensorModel {
+ public:
+  using state_type = double;
+  using weight_type = double;
+  using measurement_type = std::vector<double>;
+
+  SensorModel(double sigma, std::vector<double> map) : sigma_{sigma}, map_{std::move(map)} {}
```
Alondruck, line 191: nit: just to follow the same style as the motion model struct add the `explicit` to the constructor

docs/guides/extending-beluga.md
```diff
@@ -1,3 +1,264 @@
...
+
+  auto operator()(measurement_type detections) const {
+    return [this, detections](state_type position) -> weight_type {
+      return std::transform_reduce(
+        std::begin(detections), std::end(detections), 0.0, std::plus{},
```
Alondruck, line 196: Just a question, is `std::plus{}` correct? I used `std::multiplies{}` for the calculation in the tutorial

nahueespinosa:
Closing this one as it seems to keep creating confusion about what goes where.

PR #359, opened 2024-05-13
Fix pseudocode rendering

nahueespinosa:
#### Type of change

### Checklist

### DIFF
--- a/docs/concepts/key-concepts.md
+++ b/docs/concepts/key-concepts.md
@@ -1,6 +1,3 @@
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/pseudocode@2.4.1/build/pseudocode.min.css">
-<script src="https://cdn.jsdelivr.net/npm/pseudocode@2.4.1/build/pseudocode.min.js"></script>
-
 # Key concepts
 
 ## Monte Carlo Localization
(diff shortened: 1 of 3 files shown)


nahueespinosa:
Tested locally:
image

docs/concepts/key-concepts.md
```diff
@@ -1,6 +1,3 @@
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/pseudocode@2.4.1/build/pseudocode.min.css">
-<script src="https://cdn.jsdelivr.net/npm/pseudocode@2.4.1/build/pseudocode.min.js"></script>
-
```
hidmic, line 3: @nahueespinosa meta: this is dicey. I tried and failed to use `sphinx.pseudocode`. Since it's just a wrapper for `pseudocode.js`, I switched to raw HTML. This patch uses the broken extension and raw HTML. If the latter gets out of sync with the former, this will break again.

nahueespinosa: I see. Ok, https://github.com/Ekumen-OS/beluga/pull/359/commits/081ff53d5a81fe63243be8f63ab0cbde9c011a39 should fix the output then.

nahueespinosa: Note that this is what the extension does:

https://github.com/xxks-kkk/sphinxcontrib-pseudocode/blob/c9ee00ab02f7538735b062548fb006b8b48dd84f/sphinxcontrib/pseudocode.py#L142

nahueespinosa: Not sure if we should use [`sphinxcontrib-katex`](https://pypi.org/project/sphinxcontrib-katex/).

hidmic: I'm not entirely sure what sort of interactions between KaTex and MathJax may arise. I don't want to know either :joy:

hidmic (approved):

PR #363, opened 2024-05-15
Add conversion script from PLY to NDT

nahueespinosa:
### Proposed changes

Closes #348.

#### Type of change

- [ ] 🐛 Bugfix (change which fixes an issue)
- [x] 🚀 Feature (change which adds functionality)
- [ ] 📚 Documentation (change which fixes or extends documentation)

### Checklist

- [x] Lint and unit tests (if any) pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] All commits have been signed for [DCO](https://developercertificate.org/)

PR #364, opened 2024-05-15
Revert "Add conversion script from PLY to NDT"

nahueespinosa:
Reverts Ekumen-OS/beluga#363


hidmic (approved):
I see three alternatives here:

1. We migrate `beluga_ros` tools to C++ so we can use OpenCV and PCL.
2. We relocate `beluga_ros` tools out of the main repository.
3. We go back to @nahueespinosa custom PLY reader.

(2) is a hassle but is future-proof. (1) is a pain but lets us release the tools. (3) just pushes the problem until the next time we need some fancy Python package that is only on PyPI. Pick your poison.

nahueespinosa:
None of them look particularly appealing, but I'd go for 1.

@serraramiro1 @glpuga Feel free to chime in. For context, we had to revert this since `python-plyfile` is only available on PyPI, so we can't create a deb package with that.

PR #372, opened 2024-05-19
Use `geometry_msgs::msg::PoseArray` instead of `nav2_msgs::msg::ParticleCloud`

hidmic:
### Proposed changes

Precisely what the title says. This patch goes back to plain old  `geometry_msgs/msg/PoseArray` messages when publishing the particle cloud, effectively removing Beluga AMCL dependency on `nav2_msgs` (unlikely to land in time for ROS 2 Jazzy).

#### Type of change

### Checklist

### Additional comments

I mulled over the exact API a few times, and eventually decided against a `make_*` function that does not return a smart pointer in favor of a more procedural approach to building ROS messages. I don't love it either, but it gives better control over message instantiation, it's less cumbersome to generalize than functional alternatives, and it's easy to keep compatible for both ROS 1 and ROS 2 (e.g. ROS 2 supports a builder API for messages we could piggyback but ROS 1 doesn't :happy-cry:).

Also, note this patch replaces some of the nice functionality we had for reconstructing a representative particle cloud with plain resampling. I'll be adding that back soon, using `MarkerArray` messages instead.


nahueespinosa (approved):
LGTM

nahueespinosa (approved):

PR #386, opened 2024-05-24
Support 32-bit hashes

nahueespinosa:
### Proposed changes

This change improves compatibility with 32-bit platforms (`armhf`).

#### Type of change

### Checklist

### Additional comments

Trying to fix https://build.ros.org/job/Nbin_ufhf_uFhf__beluga__ubuntu_focal_armhf__binary/4/console.

### DIFF
--- a/beluga/include/beluga/algorithm/spatial_hash.hpp
+++ b/beluga/include/beluga/algorithm/spatial_hash.hpp
@@ -47,6 +47,7 @@ constexpr std::size_t floor_and_fibo_hash(double value) {
     if constexpr (std::is_same_v<std::size_t, std::uint64_t>) {
       return 11400714819323198485LLU;  // golden ratio for 64 bits
     } else {
+      static_assert(std::is_same_v<std::size_t, std::uint32_t>);
       return 2654435769U;  // golden ratio for 32 bits
     }
   }();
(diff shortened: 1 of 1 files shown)


beluga/include/beluga/algorithm/spatial_hash.hpp
```diff
@@ -43,18 +43,27 @@ namespace detail {
...
 template <std::size_t N, std::size_t I>
 constexpr std::size_t floor_and_fibo_hash(double value) {
-  static_assert(std::is_same_v<std::size_t, std::uint64_t>);
-  constexpr auto kFib64 = 11400714819323198485LLU;  // golden ratio for 64 bits
+  constexpr auto kFib = []() {
+    if constexpr (std::is_same_v<std::size_t, std::uint64_t>) {
+      return 11400714819323198485LLU;  // golden ratio for 64 bits
+    } else {
+      return 2654435769U;  // golden ratio for 32 bits
+    }
```
glpuga, line 51: I don't feel too strongly about it, but maybe failing if size_t is neither 32 nor 64 bits long. I mean, there are probably no relevant architectures (to us) today where this is not true, but at least when this stops being true one day, we'll catch the need to update this quickly.

glpuga (approved):

hidmic (approved):

beluga/include/beluga/algorithm/spatial_hash.hpp
nahueespinosa: Done! See https://github.com/Ekumen-OS/beluga/pull/386/commits/a3540e2c2515f70e25bb9c46eccb51670da1e41c.

nahueespinosa: 128-bit? :eye:

glpuga: Registers in all "off-the-shelf" processors have been 128 bits (or even 256) for several years now, to support vectorization. In a lot of cases those registers share register space with the regular general purpose ones to save sillicon space. I think it's a matter of time before the general purpose registers/instructions go 128 bits wide, even if not needed for physical memory addressing, just for the sake of simplifying sillicon.

Issue #305, opened 2024-02-09
Write tutorial on using Beluga


hidmic:
### Feature description

Precisely what the title says. While Beluga's API is well documented, the only reference on how to make use of the library is a ROS node and some tests. That works against adoption. We need a golden example that builds up in complexity, showcasing Beluga's features. We need to guide newcomers on reproducing that example, one step at a time.

nahueespinosa:
Following up on the conversation on how to implement the golden example, the idea is to implement the one-dimensional hallway example for Monte Carlo Localization presented in Chapter 8.3 in Thrun et. al. using the Beluga core library without ROS dependencies.

![Image](https://github.com/Ekumen-OS/beluga/assets/33042669/3cf5de0b-8a8c-4b10-96a9-0da5212f98e4)

Here is a list of milestones that can be used as reference:

1. Implement the program skeleton, focusing on the creation of a particle set (states and weights) structure and basic visualization capabilities.
    - This can be serializing to a file and then post-processing in Python, or simply writing the output to the terminal in a way that resembles the example in the book.
    - It is key to represent both the position (state) of the particle and the weight (could be a line's height, a circle's diameter, or something else).
2. Develop the initialization function for the particle set to sample uniformly from a specified interval.
    - See [`std::uniform_real_distribution`](https://en.cppreference.com/w/cpp/numeric/random/uniform_real_distribution) and [`beluga::views::sample`](https://github.com/Ekumen-OS/beluga/blob/143e1ae4c062b9a7f8719bccf138e5eff5b7f5f7/beluga/test/beluga/views/test_sample.cpp#L153).
4. Implement the sensor model capable of reweighting particles based on the map landmarks (doors) and the current position.
    - This can use a simplified version of the landmark model proposed in Chapter 6.6.
    - $w = e ^ {{- d^2} / {2 \sigma}}$ where d: _distance to the door_
    - See [`beluga::sensor`](https://github.com/Ekumen-OS/beluga/blob/main/beluga/include/beluga/sensor.hpp#L28) and [`beluga::actions::reweight`](https://github.com/Ekumen-OS/beluga/blob/main/beluga/test/beluga/actions/test_reweight.cpp#L27)
5. Develop the motion model to update the positions of the particles, including a routine for moving particles at a constant rate in a given direction.
[...]

PR #387, opened 2024-05-24
Add scripts for beluga_tutorial visualizations

Alondruck:
### Proposed changes

Related to #305, split from #340.

#### Type of change

### Checklist


beluga_tutorial/src/visualization.py
```diff
@@ -0,0 +1,164 @@
...
+        '--record-file-path',
+        type=str,
+        help='Absolute path to the record file generated by the beluga_tutorial example code',
+        required=True,
+    )
+
+    parser.add_argument(
+        '-m',
+        '--manual-control',
+        type=bool,
```
nahueespinosa, line 114: Use `action='store_true'` for on/off flags.

beluga_tutorial/src/visualization.py
```diff
@@ -0,0 +1,164 @@
...
+                lambda event: on_key(
+                    event, yaml_data, axs, current_frame_container, num_frames
+                ),
+            )
+        else:
+
+            def update(current_frame):
+                plot_stages(yaml_data, axs, current_frame)
+
+            animation.FuncAnimation(
```
nahueespinosa, line 151: The animation does not seem to work in my machine.

Alondruck: Yes, now in doesn't work in my machine also, there is a problem because `animation.FuncAnimation()` function throwing me the following warning: `UserWarning: Animation was deleted without rendering anything. This is most likely not intended. To prevent deletion, assign the Animation to a variable, e.g. `anim`, that exists until you output the Animation using plt.show() or anim.save().`
At the same time, the linter or the pre-commit doens't let me commit a code with unused variables. Maybe I can to use some type of indicator that `anim` is an unused variable.

nahueespinosa: Linters usually have a way of preventing them from changing certain parts of the code you know are correct.
Which one of the [linters](https://github.com/Ekumen-OS/beluga/blob/main/.pre-commit-config.yaml) is preventing you from adding that line?

beluga_tutorial/src/visualization.py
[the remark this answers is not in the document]
Alondruck: Maybe you refer to move the file to `beluga_tutorial/scripts/`?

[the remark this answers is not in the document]
nahueespinosa: Ah yes, that's what I meant.

beluga_tutorial/src/visualization.py
```diff
@@ -0,0 +1,164 @@
...
+    args = parser.parse_args(argv)
+
+    yaml_data = read_yaml_file(args.record_file_path)
+
+    if yaml_data:
+        fig, axs = plt.subplots(6, 1)
+        num_frames = len(yaml_data['simulation_records'])
+
+        if args.manual_control:
+            current_frame_container = [0]
```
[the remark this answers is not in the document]
Alondruck: I wanted to pass the value by reference, now I'm pushing another solution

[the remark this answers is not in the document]
nahueespinosa: In Python everything is passed by reference.

beluga_tutorial/src/visualization.py
Alondruck: this linter `flake8`. I'm trying to do the following but the linter doesn't like the `_anim` (I added the underscore because I saw this in some place but apparently it doesn't work)
```python
_anim = animation.FuncAnimation(
            fig,
            update,
            frames=num_frames,
            blit=False,
            interval=args.interval_ms,
            repeat=False,
        )
```

nahueespinosa: This might be useful:
https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors

beluga_tutorial/scripts/visualize.py
```diff
@@ -0,0 +1,162 @@
...
+def main(argv=None) -> int:
+    parser = argparse.ArgumentParser(description=globals()['__doc__'])
+
+    parser.add_argument(
+        '-p',
+        '--record-file-path',
+        type=str,
+        help='Absolute path to the record file generated by the beluga_tutorial example code',
+        required=True,
+    )
```
nahueespinosa, line 112: Consider making this a positional argument, so we can pass the path directly without `-p`.

beluga_tutorial/scripts/visualize.py
```diff
@@ -0,0 +1,162 @@
...
+        anim = animation.FuncAnimation(  # noqa: F841
+            fig,
+            update,
+            frames=num_frames,
+            blit=False,
+            interval=args.interval_ms,
+            repeat=False,
+        )
+
+    plt.show()
```
nahueespinosa, line 158: I would suggest this to avoid messing with function object attributes manually, which is not good practice.
```python
    fig, axs = plt.subplots(6, 1)
    num_frames = len(yaml_data['simulation_records'])

    def plot_stages_update(current_frame: int) -> None:
        """Helper function to update plots."""
        plot_stages(yaml_data, axs, current_frame)

    if args.manual_control:
        current_frame = 0
        plot_stages_update(current_frame)

        def on_key(event):
            """Callback function to handle key events."""
            nonlocal current_frame

            if event.key == 'right':
                current_frame += 1
            elif event.key == 'left':
                current_frame -= 1

            current_frame %= num_frames
            plot_stages_update(current_frame)

        fig.canvas.mpl_connect('key_press_event', on_key)

    else:
        anim = animation.FuncAnimation(  # noqa: F841
            fig,
            plot_stages_update,
            frames=num_frames,
            blit=False,
            interval=args.interval_ms,
            repeat=False,
        )
```

Alondruck:
@nahueespinosa @hidmic The propagate and reweight stages actually represent the same graph, as both are histograms of the same distribution of the set of particles. Do we need to show the reweight stage in a different way, or is it sufficient for what we expect from the tutorial?

nahueespinosa:
> Do we need to show the reweight stage in a different way, or is it sufficient for what we expect from the tutorial?

If we opt for the histogram display, then I think it makes more sense to show: the _current_, _prediction_ (currently called propagation), and _update_ (reweight + resample) stages. This would match the description on the [key concepts](https://ekumen-os.github.io/beluga/concepts/key-concepts.html) page about the two main stages: _prediction_ and _update_.

beluga_tutorial/scripts/visualize.py
```diff
@@ -0,0 +1,168 @@
...
+
+This script uses the matplotlib to visualize the data generated
+by beluga_tutorial example code.
+"""
+
+import yaml
+import matplotlib.pyplot as plt
+import matplotlib.animation as animation
+import numpy as np
+import argparse
```
nahueespinosa, line 27: Order alphabetically. Also, should any of these modules be included in the package.xml?

Alondruck: Do you use some specific formatter when coding in python?
Every time I run Docker and the script, it doesn't ask me to install some of those libraries, but I'll try again

nahueespinosa: Docker has its own set of dependencies installed, and some other packages do install these modules.
The formatter is installer as a pre-commit but it doesn't enforce this.

Whenever you make a list of things on multiple lines, it's good to sort them alphabetically because it makes it much easier to inspect the list to see if there's a specific item there and detect duplicate entries as they grow.

nahueespinosa (approved):
LGTM

PR #393, opened 2024-05-29
Adapt beluga_tutorial code to key concepts

Alondruck:
### Proposed changes
Adapt the main code to the names used by the key concepts.

#### Type of change

### Checklist

### Additional comments
This PR will wait for #390 to be merged into the main branch.

### DIFF
--- a/beluga_tutorial/src/main.cpp
+++ b/beluga_tutorial/src/main.cpp
@@ -224,10 +224,10 @@ int run(const std::filesystem::path& path) {
     particles |= beluga::actions::propagate(std::execution::seq, motion_model);
     record.prediction = particles;
 
-    particles |= beluga::actions::reweight(std::execution::seq, sensor_model) | beluga::actions::normalize;
-
-    particles |= beluga::views::sample |                                        //
-                 ranges::views::take_exactly(parameters.number_of_particles) |  //
+    particles |= beluga::actions::reweight(std::execution::seq, sensor_model) |  //
+                 beluga::actions::normalize |                                    //
+                 beluga::views::sample |                                         //
+                 ranges::views::take_exactly(parameters.number_of_particles) |   //
                  beluga::actions::assign;
     record.update = particles;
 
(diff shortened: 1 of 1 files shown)


Alondruck:
@nahueespinosa @hidmic I realized that we are not using the default values of the Parameters struct. Should we use them, or is it fair to require the user to provide a path to the parameters file?

In the explanation of the tutorial, I'll delete all the code related to YAML, and the user will use the default parameters. So, I want to know your thoughts on whether we also need to let the user use the default parameters here.

beluga_tutorial/src/main.cpp
```diff
@@ -224,15 +222,14 @@ int run(const std::filesystem::path& path) {
...
     particles |= beluga::actions::propagate(std::execution::seq, motion_model);
-    record.propagate = particles;
+    record.prediction = particles;
 
     particles |= beluga::actions::reweight(std::execution::seq, sensor_model) | beluga::actions::normalize;
-    record.reweight = particles;
 
     particles |= beluga::views::sample |                                        //
                  ranges::views::take_exactly(parameters.number_of_particles) |  //
                  beluga::actions::assign;
```
nahueespinosa, line 231: nit: We can define the update stage like this:

```
particles |= beluga::actions::reweight(std::execution::seq, sensor_model) |  //
             beluga::actions::normalize |                                    //
             beluga::views::sample |                                         //
             ranges::views::take_exactly(parameters.number_of_particles) |   //
             beluga::actions::assign;
```

nahueespinosa (commented):
> I want to know your thoughts on whether we also need to let the user use the default parameters here.

No strong opinion on this one. We could make the parameter_path argument optional, loading the default values if no yaml file is provided, but I also don't mind that we always ignore the default values.

beluga_tutorial/src/main.cpp
Alondruck: Maybe in this code, we can write the update stage this way. However, in the explanation of the code, I still want to separate the stages for explanatory purposes.

nahueespinosa (approved):
LGTM!

Issue #316, opened 2024-02-09
Benchmark Beluga NDT-MCL against Cartographer ROS


hidmic:
### Feature description

Connected to #93. Depends on #312 and #313. In addition to tests and micro benchmarks, NDT-MCL nodes' performance must be compared with that of other solution. Cartographer ROS is a fairly mature solution for 2D and 3D localization in the ROS ecosystem we can use to compare ourselves against.

### Implementation considerations

Pick a pair of suitable datasets and benchmark 2D vs 2D, 3D vs 3D using LAMBKIN.

serraramiro1:
@hidmic @nahueespinosa @glpuga 
Based on our conversations, and us wanting to diverge a bit from `nav2_amcl` and the difficulties of making "fair" comparisons between different sensor models and map representations, I've changed this ticket to track the benchmarking  of the model against a groundtruth signal instead. LMK what you think.

PR #395, opened 2024-05-31
Add NDT benchmark report. Adjust default parameter values for the NDT sensor model.

serraramiro1:
### Proposed changes

Adds a new benchmark report for the `NDT` sensor model for magazino datasets.
Matches NDT sensor model default parameters in the example launch files to match the ones that got empirically tested.
Related to #316.
#### Type of change

### Checklist


serraramiro1:
@hidmic I'd love your thoughts on how to best present this benchmark given that it doesn't follow the current one's format (we can't re-generate it from this repo alone).

hidmic:
We were talking precisely about this with @glpuga and @nahueespinosa earlier today. A report pulled out of nowhere is not ideal. Having old reports in our repository is not either. We need the benchmarking code out, that's on me. Until then, we should probably keep this elsewhere. Thoughts?

nahueespinosa:
Yeah, adding a pdf to this repo might not be the best. I think it makes more sense to move in the opposite direction until we can open source our benchmarking code. We were talking about removing the `beluga_benchmark` package even.

I'd keep the default parameter changes though.

serraramiro1:
Will create a new one so we keep this closed PR as-is for post-facto analysis.

PR #402, opened 2024-06-18
Release Beluga 2.0.2

nahueespinosa:
#### Type of change

### Checklist

### Additional comments

To be rebased-merged, not squashed.


hidmic (commented):
The one thing I don't love is that we are backporting what is effectively a work in progress with NDT. It's not the first patch release that does this. There is a cost in cherry-picking though. It will make it harder to backport later on. Thoughts?

nahueespinosa:
No strong opinion. We can do it the proper way. Hopefully, we won't be backporting too frequently. I don't know.

hidmic:
OK, let's move forward, hopefully we can land the missing bits soon enough.

hidmic (approved):

PR #403, opened 2024-06-25
Add SO3 and SE3 support for multivariate normal distribution

### LINKED ISSUE #398: Expand `MultivariateNormalDistribution` to SE3
### Feature description
We need a SE3 variant so that we can use this to:

- Initialize SE3 particles given an initial pose and a covariance
- Possibly as part of the implementation of the 3D variant of `OmnidirectionalDriveModel`

### Implementation considerations

Coming up with a proper normally distributed rotation in SO3 is specially complicated. There's already been some discussion in https://github.com/Ekumen-OS/beluga/pull/298#discussion_r1468925250.

serraramiro1:
### Proposed changes
Adds support for SE3 and SO3 elements in multivariate normal distribution.
Closes #398. 

# NOTE:
Rotations in Lie groups don't have infinite support, and thus the term 'normal' distribution is inherently incorrect.
This approach just follows the `SO2/SE2` approach which faces the same problems as the 3D case.

See some analysis below on how the resulting rotations in SO3 distribute close to normally  for reasonably small covariances:

The plot below shows KDE histograms and Q-Q plots for a diagonal covariance matrix of 
```
{1e-2, 1e-2, 1e-2}
```
and drawing 1000 samples using the approach that was implemented in this PR.

image

Running the [shapiro](https://en.wikipedia.org/wiki/Shapiro%E2%80%93Wilk_test) test shows that the rotation components of the samples seem drawn from a normal distribution with a `p-value` of 0.03.

Note the difference with a very big covariance matrix of 10 in all axes.
image

This setup fails the shapiro test completely.

All of this to say, that expanding to SE3 has the same caveats as the SE2 version, and thus this seems like a reasonable approach to generating monomodal distributions in SO3 and SO2 when under reasonably controlled situations (reasonable covariances).

Also, this previous dicussion is worth taking a look:
https://github.com/Ekumen-OS/beluga/pull/298#discussion_r1468925250

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/random/multivariate_distribution_traits.hpp
+++ b/beluga/include/beluga/random/multivariate_distribution_traits.hpp
@@ -17,6 +17,8 @@
 
 #include <Eigen/Core>
 #include <sophus/se2.hpp>
+#include <sophus/se3.hpp>
+#include <sophus/so3.hpp>
 
 /**
  * \file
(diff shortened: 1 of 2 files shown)


beluga/include/beluga/random/multivariate_distribution_traits.hpp
```diff
@@ -110,6 +112,58 @@ struct multivariate_distribution_traits<T, std::enable_if_t<std::is_base_of_v<So
...
+
+  /// The covariance matrix type.
+  using covariance_type = typename Eigen::Matrix<scalar_type, 3, 3>;
+
+  /// Convert from result to vector representation.
+  [[nodiscard]] static constexpr vector_type to_vector(const result_type& t) { return t.log(); }
+
+  /// Convert from vector to result representation.
+  [[nodiscard]] static constexpr result_type from_vector(const vector_type& v) {
+    return Sophus::SO3<scalar_type>::exp(v);
```
nahueespinosa, line 135: nit: Maybe we can add a block comment with a summary of the PR notes here for posterity.

nahueespinosa (approved):
LGTM!

beluga/include/beluga/random/multivariate_distribution_traits.hpp
nahueespinosa: Something along this lines?

```
    // Projecting a Gaussian through the exponential map does not yield a unimodal distribution.
    // Rotations in Lie groups don't have infinite support, making the term 'normal' distribution incorrect.
    // This approach follows the SO2/SE2 method, facing similar issues in 3D.
    // For small covariances, the resulting rotations in SO3 approximate a normal distribution.
```

serraramiro1: Done!

nahueespinosa (approved):

hidmic (approved):
LGTM though eventually a generalized Shapiro-Wilkins test would be nice addition.

PR #404, opened 2024-06-25
Add planar 3D support for `DifferentialDriveModel`

### LINKED ISSUE #399: Expand `DifferentialDriveModel` to SE3
### Feature description
Exactly as the title says. We need this for 3D part of #93.

Possibly linked to #398

serraramiro1:
### Proposed changes
Provides functionality to use SE3d states for the Differential Drive model, by performing some Z flattening and using the underlying 2D motion model.
Closes #399 .

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/3d_embedding.hpp
+++ b/beluga/include/beluga/3d_embedding.hpp
@@ -1,0 +1,36 @@
+// Copyright 2024 Ekumen, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef BELUGA_3D_EMBEDDING_HPP
+#define BELUGA_3D_EMBEDDING_HPP
+
+#include <sophus/se2.hpp>
+#include <sophus/se3.hpp>
+
+namespace beluga {
+
+/// Transforms a SE3 transform into a SE2 transform, by flattening the Z axis.
+inline Sophus::SE2d To2d(const Sophus::SE3d& tf) {
+  return Sophus::SE2d{tf.angleZ(), tf.translation().head<2>()};
+}
+
+/// Embed a SE2 transform into 3D space with zero Z translation and only rotation about the Z axis.
+inline Sophus::SE3d To3d(const Sophus::SE2d& tf) {
+  return Sophus::SE3d{
+      Sophus::SO3d::rotZ(tf.so2().log()), Eigen::Vector3d{tf.translation().x(), tf.translation().y(), 0.0}};
+}
+
+}  // namespace beluga
+
+#endif
--- a/beluga/include/beluga/motion/differential_drive_model.hpp
+++ b/beluga/include/beluga/motion/differential_drive_model.hpp
@@ -15,15 +15,16 @@
 #ifndef BELUGA_MOTION_DIFFERENTIAL_DRIVE_MODEL_HPP
 #define BELUGA_MOTION_DIFFERENTIAL_DRIVE_MODEL_HPP
 
-#include <optional>
 #include <random>
+#include <sophus/se3.hpp>
 #include <tuple>
-#include <type_traits>
 
 #include <beluga/type_traits/tuple_traits.hpp>
 
+#include <beluga/3d_embedding.hpp>
 #include <sophus/se2.hpp>
 #include <sophus/so2.hpp>
+#include <type_traits>
 
 /**
  * \file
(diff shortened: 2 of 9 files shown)


beluga/include/beluga/motion/3d_model_embedding.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+/// Wrapper for 2D motion models, in order to embed them in 3D space.
+/**
+ * This class satisfies \ref MotionModelPage.
+ *
+ * \tparam MotionModel The 2D motion model to be embedded into 3D space.
+ */
+
+template <class MotionModel>
+class Embed3D {
```
hidmic, line 43: @serraramiro1 meta: I like that `Embed3D` is general but it's not general enough -- it works with a very specific subset of all possible motion models. Sure, we could constrain the type, but I wonder if it isn't easier to generalize the motion models themselves e.g. have `DifferentialDriveModel` accept either SE2 or SE3 states. It will lead to some duplication compared to this but it makes it easier to deal with the peculiarities of each motion model. Think of motion models that take velocities as control actions.

serraramiro1: Good point, didn't think of that.
How about this new version?
Basically I don't want to test things twice, for 2d and 3d, that's why I want to use this "general" version even though it doesn't cover all the cases.

hidmic: I understand the advantages but I'm not very fond of seemingly-general-but-not-quite abstractions. If we only need and are willing to cover the case of a 3D diff drive model, then that's what we should shoot for. Summoning @glpuga and @nahueespinosa for feedback.

nahueespinosa: > Basically I don't want to test things twice.

And I think this is valid.

> If we only need and are willing to cover the case of a 3D diff drive model, then that's what we should shoot for.

I agree with this.

I have two ideas here:
- Idea number 1:
     - Inherit from the original 2D models and use `To2d` and `To3d` within the 3D class like we did here.
     - Bear with the duplication between similar motion models (diff, omni, stationary), or just implement the diff model and let's worry about this in the future.
     - Do a basic sanity test for the 3D case of each model.

- Idea number 2 (crazy):
    - Write `To2d` and `To3d` overloads for tuples of SE(n) objects.
    - Write `To2d` and `To3d` overloads for callables taking SE(n) as arguments.
    - Do (1) but with minimal duplication.
    - Profit.

hidmic: > Why would we want to support different states than sophus SE types?

Partial specialization on `Sophus::SE*` types gives us the generality we need and affords additional flexibility on state type choices to downstream users. 

Anyhow, what you did is rather close to what I had in mind, so I'm good with it too.

beluga/include/beluga/motion/differential_drive_model.hpp
```diff
@@ -146,6 +173,11 @@ class DifferentialDriveModel {
     return delta * delta;
   }
 };
+/// Alias for a 2D differential drive model, for convinience.
```
nahueespinosa, line 176: nit: add newline above.

beluga/include/beluga/3d_embedding.hpp
```diff
@@ -0,0 +1,36 @@
...
+}
+
+/// Embed a SE2 transform into 3D space with zero Z translation and only rotation about the Z axis.
+inline Sophus::SE3d To3d(const Sophus::SE2d& tf) {
+  return Sophus::SE3d{
+      Sophus::SO3d::rotZ(tf.so2().log()), Eigen::Vector3d{tf.translation().x(), tf.translation().y(), 0.0}};
+}
+
+}  // namespace beluga
+#endif
```
nahueespinosa, line 36: nit: add newline above.

beluga/include/beluga/3d_embedding.hpp
```diff
@@ -0,0 +1,36 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef BELUGA_3D_EMBEDDING_HPP
+#define BELUGA_3D_EMBEDDING_HPP
+
+#include <beluga/type_traits/tuple_traits.hpp>
```
nahueespinosa, line 18: Unused?

beluga/include/beluga/motion/differential_drive_model.hpp
```diff
@@ -68,16 +69,23 @@ struct DifferentialDriveModelParam {
...
 /// Sampled odometry model for a differential drive.
 /**
+ * Supports 2D and (flattened) 3D state types.
  * This class satisfies \ref MotionModelPage.
  *
  * See Probabilistic Robotics \cite thrun2005probabilistic Chapter 5.4.2.
+ *
+ * \tparam NDim State dimension for the state representation. Only 2 and 3 are supported.
  */
+template <int NDim = 2>
```
nahueespinosa, line 79: meta: I would prefer specializing with the state type, rather than the number of dimensions, for consistency with other 2D and 3D classes. And may also come in handy since it is directly related to the value type of the particle range.

beluga/include/beluga/motion/differential_drive_model.hpp
```diff
@@ -96,7 +104,28 @@ class DifferentialDriveModel {
...
+    }
+  }
+
+ private:
+  using control_type_2d = std::tuple<Sophus::SE2d, Sophus::SE2d>;
+  using control_type_3d = std::tuple<Sophus::SE3d, Sophus::SE3d>;
+
+  template <class Control, typename = common_tuple_type_t<Control, control_type_3d>>
+  [[nodiscard]] auto sampling_fn_3d(const Control& action) const {
+    const auto& [current_odom, previous_odom] = action;
```
nahueespinosa, line 121: nit: Using this for the internals might save a little in terms of template instantiations.
```suggestion
  [[nodiscard]] auto sampling_fn_3d(const Sophus::SE3d& current_odom, const Sophus::SE3d& previous_odom) const {
```

nahueespinosa (commented):
The approach looks good! Maybe the PR title and description need an update.

PR #405, opened 2024-07-02
Change scalar `beluga::estimate()` to yield variance

hidmic:
### Proposed changes

This isn't as much of a fix as it is a minor API contract correction for consistency with the rest. Every other `beluga::estimate` overload trades on variances, but the scalar one introduced in #340 didn't. This patch amends that.

#### Type of change

### Checklist


hidmic:
An argument can be made in favor of generalizing estimation methods to operate with different notions of dispersion, and I do think that's where we are going. Until then, I'd rather be consistent, but I can be convinced otherwise.

nahueespinosa (approved):
LGTM

Alondruck (approved):

PR #408, opened 2024-07-05
Add deserialization support for 3D NDT maps.

serraramiro1:
### Proposed changes

Support loading 3D NDT maps from disk.

#### Type of change

### Checklist

### Additional comments

The code used to generate the hdf5 file will be added in a separate PR, I've used a slightly modified version of this PR that got reverted:
https://github.com/Ekumen-OS/beluga/pull/363


beluga/include/beluga/sensor/ndt_sensor_model.hpp
```diff
@@ -239,20 +239,26 @@ class NDTSensorModel {
...
+/// - "covariances": (NUM_CELLS, 2 / 3, 2 / 3) Contains the covariance for each cell.
 ///
 ///  \tparam NDTMapRepresentation A specialized SparseValueGrid (see sensor/data/sparse_value_grid.hpp), where
-///  mapped_type == NDTCell2d, that will represent the NDT map as a mapping from 2D cells to NDTCells.
+///  mapped_type == NDTCell2d / NDTCell3d, that will represent the NDT map as a mapping from 2D / 3D cells to NDTCells.
 template <typename NDTMapRepresentationT>
-NDTMapRepresentationT load_from_hdf5_2d(const std::filesystem::path& path_to_hdf5_file) {
-  static_assert(std::is_same_v<typename NDTMapRepresentationT::mapped_type, NDTCell2d>);
-  static_assert(std::is_same_v<typename NDTMapRepresentationT::key_type, Eigen::Vector2i>);
+NDTMapRepresentationT load_from_hdf5(const std::filesystem::path& path_to_hdf5_file) {
```
hidmic, line 253: @serraramiro1 meta: code looks good overall, just a few general thoughts about API design. 

`T beluga::io::load_from_hdf5<T>(const std::filesystem::path& path)` suggests a generality it doesn't have. It also doesn't make it easy to add support for other `Ts`. I think that something like `T beluga::io::load<T>(const H5::H5File& hdf5_file)` wrapping `beluga::io::H5Loader<T>{}.load(hdf5_file)` instead has a number of advantages:

- `beluga::io::H5Loader<T>` can be both partially and fully specialized, here or in a  downstream project.
- `T beluga::io::load<T>(const H5::H5File& hdf5_file)` is as general as it looks, and could be overloaded for other storage types, here or in a downstream project.
- Taking `H5::H5File` correctly places the responsibility for file path checks on the `hdf5` library. It also simplifies testing (no need to use temporary files of any sort).

serraramiro1: I think taking `H5::H5File` makes sense, I'll make that change.

However, for your other points, I don't think I understand.
IIUC we don't need to support more than 2D and 3D case, so why would we generalise this further?.

The current approach is general when it comes to mappings from voxel indices (Eigen::Vector2i / 3i) to NDT cells (2d or 3d), as in the user can provide any map type and that's supported, but not general when it comes to the voxel types nor NDT cell types, which IMO, is fine.

I think we are covered here, what other generalizations do you foresee that the proposed path would solve?

hidmic: > I think we are covered here, what other generalizations do you foresee that the proposed path would solve?

NDT maps are not the first kind of map nor will be the last kind of map Beluga uses. Without going too far, you could have dual NDT maps and/or sparse feature grids. I imagine we would use HDF5 for those too. 

If at all possible, I would try to keep the library reusable and extensible downstream.

serraramiro1: Makes sense, I originally thought you were talking about not supporting all possible NDT maps (which I think it's not the case), and not the general HDF5 IO stuff.
I'll try to refactor things a bit to be more general. Could we merge this as-is in the meantime, though?

The non-general version is already merged in main for the 2D case.

serraramiro1: Created #409 so we keep track of this work.

hidmic: I can live with that :)

hidmic (approved):

PR #415, opened 2024-07-17
Rework `beluga_benchmark` using LAMBKIN

hidmic:
### Proposed changes

Precisely what the title says.

#### Type of change

### Checklist


hidmic:
FYI @glpuga this needs some love still but the bulk of it is there.

hidmic:
> @hidmic should we should reactivate or drop this.
> 
> I tend lean on the second option, since as long as we publish results, and replicability is possible through the lambkin repository, I don't really see a lot of advantage to having a beluga benchmarks package

Ideally, neither. We've discussed the implications of bringing LAMBKIN here and it's not great, yet developer experience for running Beluga benchmarks is terrible. It makes following a data driven development path extremely rough. 

We are discussing @LaBruma how and where to go from here.

glpuga:
> Ideally, neither.

I meant, having  12 months paused PR is not great marketing. Regardless of what we decide to do about benchmarking, we should tilt this PR one way or the other.

PR #419, opened 2024-08-01
Add implicit boundaries `LandmarkMap` constructor

hidmic:
### Proposed changes

This patch de-duplicates (and reuses) some of the code over at https://github.com/Ekumen-OS/beluga-demos.

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/sensor/data/landmark_map.hpp
+++ b/beluga/include/beluga/sensor/data/landmark_map.hpp
@@ -51,6 +51,12 @@ class LandmarkMap {
       : landmarks_(std::move(landmarks)), map_boundaries_(std::move(boundaries)) {}
 
   /// @brief Constructor with implicit map boundaries (computed from landmarks).
+  /// @important Note that computing map boundaries from landmarks will effectively
+  /// constrain the traversable region to the volume delimited by such landmarks.
+  /// In many cases this is undesirable e.g. when landmarks are scattered in some
+  /// region of space, when landmarks are distributed on some flat surface such
+  /// as a wall, when landmarks are located within a small area at some height
+  /// for visibility, etc. Use with care.
   /// @param landmarks List of landmarks that can be expected to be detected.
   explicit LandmarkMap(landmarks_set_position_data landmarks) : landmarks_(std::move(landmarks)) {
     if (!landmarks_.empty()) {
@@ -58,8 +64,8 @@ class LandmarkMap {
       map_boundaries_.max() = landmarks_[0].detection_position_in_robot;
       for (const auto& landmark : ranges::views::tail(landmarks_)) {
         const auto& position = landmark.detection_position_in_robot;
-        map_boundaries_.min() = map_boundaries_.min().cwiseMin(position).eval();
-        map_boundaries_.max() = map_boundaries_.max().cwiseMax(position).eval();
+        map_boundaries_.min() = map_boundaries_.min().cwiseMin(position);
+        map_boundaries_.max() = map_boundaries_.max().cwiseMax(position);
       }
     }
   }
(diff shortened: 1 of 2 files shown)


beluga/include/beluga/sensor/data/landmark_map.hpp
```diff
@@ -46,9 +47,23 @@ class LandmarkMap {
...
+  /// @brief Constructor with implicit map boundaries (computed from landmarks).
+  /// @param landmarks List of landmarks that can be expected to be detected.
+  explicit LandmarkMap(landmarks_set_position_data landmarks) : landmarks_(std::move(landmarks)) {
+    if (!landmarks_.empty()) {
+      map_boundaries_.min() = landmarks_[0].detection_position_in_robot;
+      map_boundaries_.max() = landmarks_[0].detection_position_in_robot;
+      for (const auto& landmark : ranges::views::tail(landmarks_)) {
+        const auto& position = landmark.detection_position_in_robot;
+        map_boundaries_.min() = map_boundaries_.min().cwiseMin(position).eval();
+        map_boundaries_.max() = map_boundaries_.max().cwiseMax(position).eval();
```
nahueespinosa, line 62: nit^n: Can we avoid a temporary here since the aliasing is harmless?

```suggestion
        map_boundaries_.min() = map_boundaries_.min().cwiseMin(position);
        map_boundaries_.max() = map_boundaries_.max().cwiseMax(position);
```

nahueespinosa (approved):
LGTM!

beluga/include/beluga/sensor/data/landmark_map.hpp
```diff
@@ -46,9 +47,23 @@ class LandmarkMap {
   /// @brief Constructor.
   /// @param boundaries Limits of the map.
   /// @param landmarks List of landmarks that can be expected to be detected.
-  LandmarkMap(const LandmarkMapBoundaries& boundaries, landmarks_set_position_data landmarks)
+  explicit LandmarkMap(const LandmarkMapBoundaries& boundaries, landmarks_set_position_data landmarks)
```
glpuga, line 50: The explicit boundaries are kind of a necessary evil. The only reason this exists is to initialize the `beluga::MultivariateUniformDistribution uniform_state_distribution_`, and the thing with Landmark maps, which does not happen in the case of OccupancyMaps, is that the support of the map may be much smaller than the traversable region.

- a field with scattered apriltags (the robot can still navigate on the outside of the region delimited by the furthest landmarks.
- all landmarks painted on the surface of a single flat wall (the landmarks delimit a zero area region, with the robots on the outside).
- four apriltags on top of a single pole, looking different ways (again zero region, and all "random" states will be a single point).

This needs not cause a failure under normal behavior, but no random states will be generated outside of the area "enclosed" be landmarks.

There's definitely a subset of cases where the apriltags region is a superset of the traversable region (the demo in beluga-demos, for instance), but at the very best a word of warning should be given in the comments for users that decide to use this constructor.

hidmic: Agreed. I'll rephrase the above for users to know what equating boundaries to map support means in practice.

hidmic: See https://github.com/Ekumen-OS/beluga/pull/419/commits/76f6c1be9e6074ebdf3bcf759d8a5e91434c7e8a.

beluga/include/beluga/sensor/data/landmark_map.hpp
hidmic: See https://github.com/Ekumen-OS/beluga/pull/419/commits/76f6c1be9e6074ebdf3bcf759d8a5e91434c7e8a.

glpuga (approved):
LGTM

PR #421, opened 2024-08-07
Drop Ubuntu user from Noble containers

hidmic:
### Proposed changes

This patch removes the `ubuntu` user from Noble containers, in order to avoid `fixuid` from clashing with it. Other than system level users, there should only be a `developer` user in development containers.

#### Type of change

### Checklist


hidmic:
> I got an error in the latter when building the workspace (flatland_server does not build). This is not related to the Dockerfile I think, so should I rise an issue to check it?

It is unrelated, yes. `flatland` is not that well maintained so we have to figure out what to do with it moving forward. CC @nahueespinosa.

PR #422, opened 2024-08-08
Add NDT AMCL 3D node.

### LINKED ISSUE #311: Implement nD-NDT sensor model in Beluga
### Feature description

Connected to #93. Depends on #309. Precisely what the title says. Craft a likelihood function by matching a point cloud with an NDT map. Good performance is crucial, so make sure to micro benchmark it. Optimizations for the 2D and/or 3D case are also valid.

### LINKED ISSUE #313: Implement 3D NDT-MCL ROS 2 node
### Feature description

Connected to #93. Depends on #311. Precisely what the title says. Implement a ROS 2 node for 2D localization by feeding point clouds and a map to a 3D-NDT sensor model. Use a depth camera on a Kobuki robot (in simulation or in the real-world) to test it.

serraramiro1:
### Proposed changes

Adds a new node for 3D NDT AMCL.
Closes #422 
Closes #311
Closes #313 

#### Type of change

### Checklist


serraramiro1:
@hidmic ready for a first pass.
Also, this seems to not work on Jazzy, I'll TAL in the meantime.

I'll try to bring the coverage for the node a little bit higher.

serraramiro1:
Running jazzy's container is giving me :

```
developer@23e2d3be126e:/ws$ colcon build --symlink-install --packages-up-to=beluga_amcl
Starting >>> beluga
--- stderr: beluga
CMake Error at test/CMakeLists.txt:30 (find_package):
  By not providing "Findbenchmark.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "benchmark", but CMake did not find one.

  Could not find a package configuration file provided by "benchmark" with
  any of the following names:

    benchmarkConfig.cmake
    benchmark-config.cmake

  Add the installation prefix of "benchmark" to CMAKE_PREFIX_PATH or set
  "benchmark_DIR" to a directory containing one of the above files.  If
  "benchmark" provides a separate development package or SDK, be sure it has
  been installed.


---
Failed   <<< beluga [1.02s, exited with code 1]

Summary: 0 packages finished [1.19s]
  1 package failed: beluga
  1 package had stderr output: beluga
  2 packages not processed
 ```
  
 Possibly related to #421 @hidmic do you think you can try to repro? I checked out `main` and getting the same

File: beluga_amcl/include/beluga_amcl/ndt_amcl_node_3d.hpp [hunk 1]
```diff
@@ -0,0 +1,155 @@
...
+ */
+
+namespace beluga_amcl {
+
+/// Underlying map representation for the NDT sensor model.
+using NDTMapRepresentation =
+    beluga::SparseValueGrid3<std::unordered_map<Eigen::Vector3i, beluga::NDTCell3d, beluga::detail::CellHasher<3>>>;
+
+/// Type of a particle-dependent random state generator.
+using RandomStateGenerator = std::function<typename beluga::NDTSensorModel<NDTMapRepresentation>::state_type()>;
```
hidmic, line 57: @serraramiro1 nit: consider aliasing `typename beluga::NDTSensorModel<NDTMapRepresentation>::state_type`, we use it more than once.

beluga_amcl/src/ndt_amcl_node_3d.cpp
```diff
@@ -0,0 +1,503 @@
...
+    descriptor.floating_point_range[0].from_value = 0;
+    descriptor.floating_point_range[0].to_value = 1000;
+    descriptor.floating_point_range[0].step = 0;
+    declare_parameter("d2", rclcpp::ParameterValue(0.6), descriptor);
+  }
+
+  if (get_parameter("autostart").as_bool()) {
+    auto autostart_delay = std::chrono::duration<double>(get_parameter("autostart_delay").as_double());
+    autostart_timer_ = create_wall_timer(autostart_delay, std::bind(&NdtAmclNode3D::autostart_callback, this));
+  }
```
hidmic, line 153: @serraramiro1 nit: why don't we reuse this block?

beluga_amcl/src/ndt_amcl_node_3d.cpp
```diff
@@ -0,0 +1,503 @@
...
+    shutdown();
+  }
+  RCLCPP_WARN(get_logger(), "Auto configured successfully");
+  current_state = activate();
+  if (current_state.id() != State::PRIMARY_STATE_ACTIVE) {
+    RCLCPP_WARN(get_logger(), "Failed to auto activate, shutting down");
+    shutdown();
+  }
+  RCLCPP_INFO(get_logger(), "Auto activated successfully");
+}
```
hidmic, line 170: @serraramiro1 meta: FWIW this looks like a perfectly reasonable autostart sequence for every node out there.

beluga_amcl/src/ndt_amcl_node_3d.cpp
```diff
@@ -0,0 +1,503 @@
...
+  covariance.coeffRef(3, 3) = get_parameter("initial_pose.covariance_roll").as_double();
+  covariance.coeffRef(4, 4) = get_parameter("initial_pose.covariance_pitch").as_double();
+  covariance.coeffRef(5, 5) = get_parameter("initial_pose.covariance_yaw").as_double();
+
+  return std::make_pair(pose, covariance);
+}
+
+auto NdtAmclNode3D::get_motion_model() const -> MotionModelVariant {
+  const auto name = get_parameter("robot_model_type").as_string();
+  if (name == kDifferentialModelName || name == kNav2DifferentialModelName) {
```
hidmic, line 248: @serraramiro1 meta: I would just drop `kNav2DifferentialModelName` compatibility here. This is not good ol' AMCL.

beluga_amcl/src/ndt_amcl_node_3d.cpp
```diff
@@ -0,0 +1,503 @@
...
+  params.d1 = get_parameter("d1").as_double();
+  params.d2 = get_parameter("d2").as_double();
+  const auto map_path = get_parameter("map_path").as_string();
+  RCLCPP_INFO(get_logger(), "Loading map from %s.", map_path.c_str());
+
+  return beluga::NDTSensorModel<NDTMapRepresentation>{
+      params, beluga::io::load_from_hdf5<NDTMapRepresentation>(get_parameter("map_path").as_string())};
+}
+
+auto NdtAmclNode3D::get_execution_policy() const -> ExecutionPolicyVariant {
```
hidmic, line 271: @serraramiro1 nit: this looks reusable too.

beluga_amcl/src/ndt_amcl_node_3d.cpp
```diff
@@ -0,0 +1,503 @@
...
+  }
+
+  const auto update_start_time = std::chrono::high_resolution_clock::now();
+
+  pcl::PointCloud<pcl::PointXYZ> pcl;
+  std::vector<Eigen::Vector3d> measurement;
+  pcl::fromROSMsg(*laser_scan, pcl);
+  measurement.reserve(pcl.points.size());
+
+  // TODO (serraramiro1): Can this be more efficient?
```
hidmic, line 381: @serraramiro1 well, we could use a [`PointCloud2Iterator`](https://github.com/ros2/common_interfaces/blob/rolling/sensor_msgs/include/sensor_msgs/point_cloud2_iterator.hpp) for each coordinate field, then zip them, then transform them to `Eigen::Vector3d` and apply the transform one by one. We still have to apply the transform for each point, but at least we won't be bringing PCL as a dependency.

hidmic: Another intrusive thought. Point clouds will always be super dense, but NDT maps are sparse. Would it more efficient to transform the map instead? Food for thought, wouldn't try that here.

beluga_amcl/src/ndt_amcl_node_3d.cpp
```diff
@@ -0,0 +1,503 @@
...
+    auto [base_pose_in_map, base_pose_covariance] = new_estimate.value();
+    tf2::toMsg(base_pose_in_map, message.pose.pose);
+
+    static constexpr double kMinimumVariance =
+        1e-6;  // Make sure we covariance is not singular so that UT doesn't fail.
+    for (auto index = Eigen::Index{0}; index <= base_pose_covariance.cols(); ++index) {
+      base_pose_covariance.coeffRef(index, index) = std::max(base_pose_covariance(index, index), kMinimumVariance);
+    }
+    const auto& [base_pose_in_map_xyz_rpy, base_pose_covariance_xyz_rpy] =
+        beluga::unscented_transform(base_pose_in_map.log(), base_pose_covariance, kTangentialSpaceToXyzRpy);
```
hidmic, line 462: @serraramiro1 I haven't reviewed the math yet, but this looks like something that should live in `beluga_ros`.

beluga_amcl/test/test_ndt_amcl_3d_node.cpp
```diff
@@ -0,0 +1,564 @@
...
+  spin_for(10ms, ndt_amcl_node_);  // ensure orderly processing
+  tester_node_->publish_3d_laser_scan();
+  ASSERT_TRUE(wait_for_pose_estimate());
+  ASSERT_TRUE(wait_for_transform("map", "odom"));
+
+  {
+    const auto [pose, _] = ndt_amcl_node_->estimate();
+    ASSERT_NEAR(pose.translation().x(), 0.0, 0.01);
+    ASSERT_NEAR(pose.translation().y(), 0.0, 0.01);
+    // ASSERT_NEAR(pose.so2().log(), 0., 0.01);
```
hidmic, line 445: @serraramiro1 why is this commented out? Same below.

hidmic (commented):
First pass, great work.

beluga_amcl/src/ndt_amcl_node_3d.cpp
serraramiro1: Good point, although we first apply NDT to the clouds and then we transform, for every particle, so we do this (somewhat costly) transform only once. We could update the measurement type to take in the already NDT transformed type and maybe make this even cheaper

beluga_amcl/src/ndt_amcl_node_3d.cpp
```diff
@@ -0,0 +1,447 @@
...
+        params.alpha_fast = get_parameter("recovery_alpha_fast").as_double();
+        params.kld_epsilon = get_parameter("pf_err").as_double();
+        params.kld_z = get_parameter("pf_z").as_double();
+
+        auto hasher = beluga::spatial_hash<Sophus::SE3d>(
+            get_parameter("spatial_resolution_x").as_double(), get_parameter("spatial_resolution_theta").as_double());
+
+        // We don't support randomly sampling from NDT maps, so we enforce a non-adaptive filter.
+        assert(params.min_particles == params.max_particles);
+        RandomStateGenerator random_state_maker = []() { return Sophus::SE3d{}; };
```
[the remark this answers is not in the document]
hidmic: One thought about random sampling these maps. If sparse, sampling from a uniform over their convex hull (or some other user-defined boundary) or a few sigmas away from any given normally distributed "voxel" will often (if not almost always) yield free space. We just need an efficient way of checking that it is indeed free.

beluga_amcl/src/ndt_amcl_node_3d.cpp
```diff
@@ -0,0 +1,447 @@
...
+    RCLCPP_ERROR(get_logger(), "Couldn't find needed transform, : %s", error.what());
+    return;
+  }
+
+  const auto update_start_time = std::chrono::high_resolution_clock::now();
+
+  std::vector<Eigen::Vector3d> measurement;
+  measurement.reserve(laser_scan->height * laser_scan->width);
+
+  //  Accessing XYZ as suggested here:
```
hidmic, line 345: @serraramiro1 nit: 
```suggestion
  // Accessing XYZ as suggested here:
```

beluga_amcl/src/ndt_amcl_node_3d.cpp
```diff
@@ -0,0 +1,447 @@
...
+  // https://docs.ros.org/en/jade/api/sensor_msgs/html/classsensor__msgs_1_1PointCloud2Iterator.html
```
hidmic, line 346: @serraramiro1 meta: this makes extra assumptions about pointcloud field layout. IMO an iterator for each well-known named field does not make such assumptions.

hidmic (commented):
Second pass!

beluga_amcl/src/ndt_amcl_node_3d.cpp
serraramiro1: Addressed, cannot use fancy ranges things as these iterators are custom classes that don't adhere to iterator conventions.
(No `begin()`, for instance).

beluga_amcl/src/ndt_amcl_node_3d.cpp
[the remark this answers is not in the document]
serraramiro1: > @serraramiro1 meta: should we have a beluga::Mcl class for this instead of hacking beluga::Amcl? Also, why NdtAmcl if it's not really adaptive?


Probably, or at least an option to disable the adaptiveness.

There's a ticket for this:
https://github.com/Ekumen-OS/beluga/issues/424

> One thought about random sampling these maps. If sparse, sampling from a uniform over their convex hull (or some other user-defined boundary) or a few sigmas away from any given normally distributed "voxel" will often (if not almost always) yield free space. We just need an efficient way of checking that it is indeed free.

Randomly sampling 3D space seems like a hail mary, to me, how likely is it that we sample a decent state from a 3D map? It get's way more unlikely. Also, rotations come into play aswell. I personally don't expect that uniformly sampling SO3, + picking free 3D translations, regardless of the sampling approach will yield good returns for realistic use cases.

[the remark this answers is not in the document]
hidmic: Define decent state. Likelihood function variance plays a big role here. I wouldn't outright disregard the approach, at least not without some experimentation.

FWIW it's not that you don't have a point. You do. We know this will eventually break down -- not just random sampling, but the bootstrap particle filter itself. But it'll be easier to justify an optimal transport particle filter when we get there.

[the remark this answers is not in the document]
serraramiro1: Fair.
Anyways, this won't happen as part of this PR. How should we move forward with this?

beluga_amcl/src/ndt_amcl_node_3d.cpp
```diff
@@ -0,0 +1,450 @@
...
+        laser_pose_in_base);
+  } catch (const tf2::TransformException& error) {
+    RCLCPP_ERROR(get_logger(), "Couldn't find needed transform, : %s", error.what());
+    return;
+  }
+
+  const auto update_start_time = std::chrono::high_resolution_clock::now();
+
+  std::vector<Eigen::Vector3d> measurement;
+  measurement.reserve(laser_scan->height * laser_scan->width);
```
hidmic, line 344: Note for myself. We need a better abstraction for point clouds in Beluga.

beluga_ros/include/beluga_ros/tf2_sophus.hpp
```diff
@@ -245,6 +269,28 @@ inline Sophus::Matrix3<Scalar>& covarianceRowMajorToEigen(const Array<Scalar, 36
...
+
+  // ROS covariances use extrinsic RPY parametrization, we need to convert to it from our tangent space
+  // representation. We use an unscented transform to apply such transform as it's non linear.
+  // See https://www.ros.org/reps/rep-0103.html#covariance-representation .
+
+  const auto& [base_pose_in_map_xyz_rpy, base_pose_covariance_xyz_rpy] =
+      beluga::unscented_transform(in.log(), covariance, detail::TangentialSpaceToXyzRpy<Scalar>);
+  std::copy(
+      base_pose_covariance_xyz_rpy.data(), base_pose_covariance_xyz_rpy.data() + base_pose_covariance_xyz_rpy.size(),
+      out.covariance.data());
```
hidmic, line 290: @serraramiro1 meta: are we sure storage layout matches between Eigen and ROS messages? I vaguely recall Eigen using column major layouts by default, and ROS messages using row major layouts.

hidmic (approved):
How can I give this a shot locally @serraramiro1?

beluga_ros/include/beluga_ros/tf2_sophus.hpp
serraramiro1: You're totally right. I was 100% sure Eigen was row major too, turns out I was wrong.

Fixed!

serraramiro1:
> How can I give this a shot locally @serraramiro1?

Try with the dataset and map I shared locally, and use:

Replace arguments to match your local env
```
ros2 launch beluga_example ndt_replay_3d.launch.xml localization_ndt_map:=/ws/src/beluga/1m_map.hdf5 bag_path:=/ws/src/beluga/rosbag2_2024_08_01-12_53_07
```

hidmic (approved):
Proof of life working

Issue #55, opened 2023-01-09
Extend likelihood field to model unexplored space


nahueespinosa:
## Description

_Probabilistic Robotics, Chapter 6.4.2_ mentions a few disadvantages of the likelihood field sensor model:

- It does not explicitly model people and other dynamics that might cause short readings.
- It treats sensors as if they can “see through walls".
- It does not take map uncertainty into account.

To diminish the effect of these limitations, the following extension is proposed:

> [...] one might sort map occupancy values into three categories: occupied, free, and unknown, instead of just the first two. When a sensor measurement $z^k_t$ falls into the category unknown, its probability $p(z^k_t | x_t, m)$ is assumed to be the constant value $1 / z_{max}$. The resulting probabilistic model is that in the unexplored space, every sensor measurement is equally likely.

## Definition of Done

- [ ] Update our likelihood field model implementation to model unexplored space.

nahueespinosa:
Closed by #430.

PR #430, opened 2024-08-22
Extending likelihood field to model unexplored spaces

DPR00:
### Proposed changes

The following changes are proposed in order to enhance the Likelihood Field Model based on the suggestions described on #55:
- First, a new function `unknown_obstacle_data()` is created to ideally replace the `obstacle_data()` function in `beluga/sensor/data/OccupancyGrid.hpp`. This new function will return a `srd::tuple<bool, bool>` containing the values which are occupied and unknown. 
- Then, this new function will be passed to calculate the distance map using a new function `nearest_obstacle_unknown_distance_map`. This new function will ideally replace the `nearest_obstacle_distance_map` function.
- The value of the distance map for unknown values is assigned with a pre-computed value inside the `nearest_obstacle_unknown_distance_map` function. Based on this computed value, the likelihood will be $\frac{1}{z_{max}}$.

Some comments:
- It passed the compilation steps and the beluga example was tested. It would be nice to make some maps testing the draft. How could I perform that?

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/actions/overlay.hpp
+++ b/beluga/include/beluga/actions/overlay.hpp
@@ -1,0 +1,126 @@
+// Copyright 2024 Ekumen, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef BELUGA_ACTIONS_OVERLAY_HPP
+#define BELUGA_ACTIONS_OVERLAY_HPP
+
+#include <algorithm>
+#include <execution>
+
+#include <range/v3/action/action.hpp>
+#include <range/v3/view/common.hpp>
+#include <range/v3/view/transform.hpp>
+
+/**
+ * \file
+ * \brief Implementation of the overlay range adaptor object
+ */
+namespace beluga::actions {
+
+namespace detail {
+
+/// Implementation detail for an overlay range adaptor object.
+struct overlay_base_fn {
+  /// Overload that implements an overlay of a value in a range.
+  /**
+   * \tparam ExecutionPolicy An [execution policy](https://en.cppreference.com/w/cpp/algorithm/execution_policy_tag_t).
+   * \tparam Range An [input range](https://en.cppreference.com/w/cpp/ranges/input_range).
+   * \tparam MaskRange An [input range](https://en.cppreference.com/w/cpp/ranges/input_range).
+   * \param policy The execution policy to use.
+   * \param range An existing range to apply this action to.
+   * \param mask The mask where the values will be overlaid.
+   * \param mask_value The value to be overlaid.
+   */
+  template <
+      class ExecutionPolicy,
+      class Range,
+      class MaskRange,
+      class Mask = ranges::range_value_t<Range>,
+      std::enable_if_t<std::is_execution_policy_v<std::decay_t<ExecutionPolicy>>, int> = 0,
+      std::enable_if_t<ranges::range<Range>, int> = 0,
+      std::enable_if_t<ranges::range<MaskRange>, int> = 0>
+  constexpr auto operator()(ExecutionPolicy&& policy, Range& range, MaskRange& mask, Mask mask_value) const -> Range& {
+    auto map = range | ranges::views::common;
+
+    std::transform(
+        policy,            //
+        std::begin(map),   //
+        std::end(map),     //
+        std::begin(mask),  //
+        std::begin(map),   //
+        [mask_value](const auto& base_value, bool flag) { return flag ? mask_value : base_value; });
+
+    return range;
+  }
+
+  /// Overload that re-orders arguments from an action closure.
+  template <
+      class ExecutionPolicy,
+      class Range,
+      class MaskRange,
+      class Mask = ranges::range_value_t<Range>,
+      std::enable_if_t<std::is_execution_policy_v<ExecutionPolicy>, int> = 0,
+      std::enable_if_t<ranges::range<Range>, int> = 0,
+      std::enable_if_t<ranges::range<MaskRange>, int> = 0>
+  constexpr auto operator()(Range&& range, MaskRange& mask, Mask mask_value, ExecutionPolicy policy) const -> Range& {
+    return (*this)(std::move(policy), std::forward<Range>(range), mask, mask_value);
+  }
+
+  /// Overload that returns an action closure to compose with other actions.
+  template <
+      class ExecutionPolicy,
+      class MaskRange,
+      class Mask,
+      std::enable_if_t<std::is_execution_policy_v<ExecutionPolicy>, int> = 0,
+      std::enable_if_t<ranges::range<MaskRange>, int> = 0>
+  constexpr auto operator()(ExecutionPolicy policy, MaskRange&& mask, const Mask& mask_value) const {
+    return ranges::make_action_closure(
+        ranges::bind_back(overlay_base_fn{}, std::forward<MaskRange>(mask), mask_value, std::move(policy)));
+  }
+};
+
+/// Implementation detail for an overlay range adaptor object with a default execution policy.
+struct overlay_fn : public overlay_base_fn {
+  using overlay_base_fn::operator();
+
+  /// Overload that defines a default execution policy.
+  template <
+      class Range,
+      class MaskRange,
+      class Mask = ranges::range_value_t<Range>,
+      std::enable_if_t<ranges::range<Range>, int> = 0,
+      std::enable_if_t<ranges::range<MaskRange>, int> = 0>
+  constexpr auto operator()(Range&& range, MaskRange& mask, Mask mask_value) const -> Range& {
+    return (*this)(std::execution::seq, std::forward<Range>(range), mask, mask_value);
+  }
+
+  /// Overload that returns an action closure to compose with other actions.
+  template <class MaskRange, class Mask, std::enable_if_t<ranges::range<MaskRange>, int> = 0>
+  constexpr auto operator()(MaskRange&& mask, const Mask& mask_value) const {
+    return ranges::make_action_closure(ranges::bind_back(overlay_fn{}, std::forward<MaskRange>(mask), mask_value));
+  }
+};
+
+}  // namespace detail
+
+/// [Range adaptor object](https://en.cppreference.com/w/cpp/named_req/RangeAdaptorObject) that
+/// can overlay a range of values (or a range of particles).
+/**
+ * The `overlay` range adaptor allows to overlay the values of the range that match a mask.
+ * All the values are overlaid for a given value.
+ */
+inline constexpr ranges::actions::action_closure<detail::overlay_fn> overlay;
+}  // namespace beluga::actions
+
+#endif
--- a/beluga/include/beluga/actions/overlay.hpp
+++ b/beluga/include/beluga/actions/overlay.hpp
@@ -53,6 +53,7 @@ struct overlay_base_fn {
   constexpr auto operator()(ExecutionPolicy&& policy, Range& range, MaskRange&& mask, Mask&& mask_value) const
       -> Range& {
     auto map = range | ranges::views::common;
+    const auto converted_mask_value = static_cast<ranges::range_value_t<Range>>(mask_value);
 
     std::transform(
         policy,            //
@@ -60,7 +61,9 @@ struct overlay_base_fn {
         std::end(map),     //
         std::begin(mask),  //
         std::begin(map),   //
-        [&mask_value](const auto& base_value, bool flag) { return flag ? mask_value : base_value; });
+        [&converted_mask_value](const auto& base_value, bool flag) {
+          return flag ? converted_mask_value : base_value;
+        });
 
     return range;
   }
--- a/beluga/include/beluga/actions/overlay.hpp
+++ b/beluga/include/beluga/actions/overlay.hpp
@@ -50,7 +50,8 @@ struct overlay_base_fn {
       std::enable_if_t<std::is_execution_policy_v<std::decay_t<ExecutionPolicy>>, int> = 0,
       std::enable_if_t<ranges::range<Range>, int> = 0,
       std::enable_if_t<ranges::range<MaskRange>, int> = 0>
-  constexpr auto operator()(ExecutionPolicy&& policy, Range& range, MaskRange& mask, Mask mask_value) const -> Range& {
+  constexpr auto operator()(ExecutionPolicy&& policy, Range& range, const MaskRange& mask, const Mask& mask_value) const
+      -> Range& {
     auto map = range | ranges::views::common;
 
     std::transform(
@@ -73,8 +74,9 @@ struct overlay_base_fn {
       std::enable_if_t<std::is_execution_policy_v<ExecutionPolicy>, int> = 0,
       std::enable_if_t<ranges::range<Range>, int> = 0,
       std::enable_if_t<ranges::range<MaskRange>, int> = 0>
-  constexpr auto operator()(Range&& range, MaskRange& mask, Mask mask_value, ExecutionPolicy policy) const -> Range& {
-    return (*this)(std::move(policy), std::forward<Range>(range), mask, mask_value);
+  constexpr auto operator()(Range&& range, MaskRange&& mask, const Mask& mask_value, ExecutionPolicy policy) const
+      -> Range& {
+    return (*this)(std::move(policy), std::forward<Range>(range), std::forward<MaskRange>(mask), std::move(mask_value));
   }
 
   /// Overload that returns an action closure to compose with other actions.
@@ -86,7 +88,7 @@ struct overlay_base_fn {
       std::enable_if_t<ranges::range<MaskRange>, int> = 0>
   constexpr auto operator()(ExecutionPolicy policy, MaskRange&& mask, const Mask& mask_value) const {
     return ranges::make_action_closure(
-        ranges::bind_back(overlay_base_fn{}, std::forward<MaskRange>(mask), mask_value, std::move(policy)));
+        ranges::bind_back(overlay_base_fn{}, std::forward<MaskRange>(mask), std::move(mask_value), std::move(policy)));
   }
 };
 
@@ -101,14 +103,16 @@ struct overlay_fn : public overlay_base_fn {
       class Mask = ranges::range_value_t<Range>,
       std::enable_if_t<ranges::range<Range>, int> = 0,
       std::enable_if_t<ranges::range<MaskRange>, int> = 0>
-  constexpr auto operator()(Range&& range, MaskRange& mask, Mask mask_value) const -> Range& {
-    return (*this)(std::execution::seq, std::forward<Range>(range), mask, mask_value);
+  constexpr auto operator()(Range&& range, MaskRange&& mask, const Mask& mask_value) const -> Range& {
+    return (*this)(
+        std::execution::seq, std::forward<Range>(range), std::forward<MaskRange>(mask), std::move(mask_value));
   }
 
   /// Overload that returns an action closure to compose with other actions.
   template <class MaskRange, class Mask, std::enable_if_t<ranges::range<MaskRange>, int> = 0>
   constexpr auto operator()(MaskRange&& mask, const Mask& mask_value) const {
-    return ranges::make_action_closure(ranges::bind_back(overlay_fn{}, std::forward<MaskRange>(mask), mask_value));
+    return ranges::make_action_closure(
+        ranges::bind_back(overlay_fn{}, std::forward<MaskRange>(mask), std::move(mask_value)));
   }
 };
 
--- a/beluga/include/beluga/actions/overlay.hpp
+++ b/beluga/include/beluga/actions/overlay.hpp
@@ -50,7 +50,7 @@ struct overlay_base_fn {
       std::enable_if_t<std::is_execution_policy_v<std::decay_t<ExecutionPolicy>>, int> = 0,
       std::enable_if_t<ranges::range<Range>, int> = 0,
       std::enable_if_t<ranges::range<MaskRange>, int> = 0>
-  constexpr auto operator()(ExecutionPolicy&& policy, Range& range, const MaskRange& mask, const Mask& mask_value) const
+  constexpr auto operator()(ExecutionPolicy&& policy, Range& range, MaskRange&& mask, Mask&& mask_value) const
       -> Range& {
     auto map = range | ranges::views::common;
 
@@ -60,7 +60,7 @@ struct overlay_base_fn {
         std::end(map),     //
         std::begin(mask),  //
         std::begin(map),   //
-        [mask_value](const auto& base_value, bool flag) { return flag ? mask_value : base_value; });
+        [&mask_value](const auto& base_value, bool flag) { return flag ? mask_value : base_value; });
 
     return range;
   }
@@ -74,9 +74,10 @@ struct overlay_base_fn {
       std::enable_if_t<std::is_execution_policy_v<ExecutionPolicy>, int> = 0,
       std::enable_if_t<ranges::range<Range>, int> = 0,
       std::enable_if_t<ranges::range<MaskRange>, int> = 0>
-  constexpr auto operator()(Range&& range, MaskRange&& mask, const Mask& mask_value, ExecutionPolicy policy) const
+  constexpr auto operator()(Range&& range, MaskRange&& mask, Mask&& mask_value, ExecutionPolicy policy) const
       -> Range& {
-    return (*this)(std::move(policy), std::forward<Range>(range), std::forward<MaskRange>(mask), std::move(mask_value));
+    return (*this)(
+        std::move(policy), std::forward<Range>(range), std::forward<MaskRange>(mask), std::forward<Mask>(mask_value));
   }
 
   /// Overload that returns an action closure to compose with other actions.
@@ -86,9 +87,9 @@ struct overlay_base_fn {
       class Mask,
       std::enable_if_t<std::is_execution_policy_v<ExecutionPolicy>, int> = 0,
       std::enable_if_t<ranges::range<MaskRange>, int> = 0>
-  constexpr auto operator()(ExecutionPolicy policy, MaskRange&& mask, const Mask& mask_value) const {
-    return ranges::make_action_closure(
-        ranges::bind_back(overlay_base_fn{}, std::forward<MaskRange>(mask), std::move(mask_value), std::move(policy)));
+  constexpr auto operator()(ExecutionPolicy policy, MaskRange&& mask, Mask&& mask_value) const {
+    return ranges::make_action_closure(ranges::bind_back(
+        overlay_base_fn{}, std::forward<MaskRange>(mask), std::forward<Mask>(mask_value), std::move(policy)));
   }
 };
 
@@ -103,16 +104,16 @@ struct overlay_fn : public overlay_base_fn {
       class Mask = ranges::range_value_t<Range>,
       std::enable_if_t<ranges::range<Range>, int> = 0,
       std::enable_if_t<ranges::range<MaskRange>, int> = 0>
-  constexpr auto operator()(Range&& range, MaskRange&& mask, const Mask& mask_value) const -> Range& {
+  constexpr auto operator()(Range&& range, MaskRange&& mask, Mask&& mask_value) const -> Range& {
     return (*this)(
-        std::execution::seq, std::forward<Range>(range), std::forward<MaskRange>(mask), std::move(mask_value));
+        std::execution::seq, std::forward<Range>(range), std::forward<MaskRange>(mask), std::forward<Mask>(mask_value));
   }
 
   /// Overload that returns an action closure to compose with other actions.
   template <class MaskRange, class Mask, std::enable_if_t<ranges::range<MaskRange>, int> = 0>
-  constexpr auto operator()(MaskRange&& mask, const Mask& mask_value) const {
+  constexpr auto operator()(MaskRange&& mask, Mask&& mask_value) const {
     return ranges::make_action_closure(
-        ranges::bind_back(overlay_fn{}, std::forward<MaskRange>(mask), std::move(mask_value)));
+        ranges::bind_back(overlay_fn{}, std::forward<MaskRange>(mask), std::forward<Mask>(mask_value)));
   }
 };
 
--- a/beluga/include/beluga/actions/overlay.hpp
+++ b/beluga/include/beluga/actions/overlay.hpp
@@ -46,7 +46,7 @@ struct overlay_base_fn {
       class ExecutionPolicy,
       class Range,
       class MaskRange,
-      class Mask = ranges::range_value_t<Range>,
+      class Mask,
       std::enable_if_t<std::is_execution_policy_v<std::decay_t<ExecutionPolicy>>, int> = 0,
       std::enable_if_t<ranges::range<Range>, int> = 0,
       std::enable_if_t<ranges::range<MaskRange>, int> = 0>
@@ -70,7 +70,7 @@ struct overlay_base_fn {
       class ExecutionPolicy,
       class Range,
       class MaskRange,
-      class Mask = ranges::range_value_t<Range>,
+      class Mask,
       std::enable_if_t<std::is_execution_policy_v<ExecutionPolicy>, int> = 0,
       std::enable_if_t<ranges::range<Range>, int> = 0,
       std::enable_if_t<ranges::range<MaskRange>, int> = 0>
@@ -101,7 +101,7 @@ struct overlay_fn : public overlay_base_fn {
   template <
       class Range,
       class MaskRange,
-      class Mask = ranges::range_value_t<Range>,
+      class Mask,
       std::enable_if_t<ranges::range<Range>, int> = 0,
       std::enable_if_t<ranges::range<MaskRange>, int> = 0>
   constexpr auto operator()(Range&& range, MaskRange&& mask, Mask&& mask_value) const -> Range& {
--- a/beluga/include/beluga/algorithm/distance_map.hpp
+++ b/beluga/include/beluga/algorithm/distance_map.hpp
@@ -41,7 +41,7 @@ namespace beluga {
  *  (std::size_t) -> NeighborsT, where NeighborsT is a
  *    [Range](https://en.cppreference.com/w/cpp/ranges/range)
  *    with value type std::size_t.
- * \param obstacle_map A map that represents obstacles in an environment.
+ * \param obstacle_mask A mask that represents obstacles in an environment.
  *  If the value of a cell is True, the cell has an obstacle.
  * \param distance_function Given the indexes of two cells in the map i and j,
  *  obstacle_map(i, j) must return the distance between the two cells.
@@ -52,7 +52,7 @@ namespace beluga {
  */
 template <class Range, class DistanceFunction, class NeighborsFunction>
 auto nearest_obstacle_distance_map(
-    Range&& obstacle_map,
+    Range&& obstacle_mask,
     DistanceFunction&& distance_function,
     NeighborsFunction&& neighbors_function) {
   struct IndexPair {
@@ -61,15 +61,15 @@ auto nearest_obstacle_distance_map(
   };
 
   using DistanceType = std::invoke_result_t<DistanceFunction, std::size_t, std::size_t>;
-  auto distance_map = std::vector<DistanceType>(ranges::size(obstacle_map));
-  auto visited = std::vector<bool>(ranges::size(obstacle_map), false);
+  auto distance_map = std::vector<DistanceType>(ranges::size(obstacle_mask));
+  auto visited = std::vector<bool>(ranges::size(obstacle_mask), false);
 
   auto compare = [&distance_map](const IndexPair& first, const IndexPair& second) {
     return distance_map[first.index] > distance_map[second.index];
   };
   auto queue = std::priority_queue<IndexPair, std::vector<IndexPair>, decltype(compare)>{compare};
 
-  for (auto [index, is_obstacle] : ranges::views::enumerate(obstacle_map)) {
+  for (auto [index, is_obstacle] : ranges::views::enumerate(obstacle_mask)) {
     if (is_obstacle) {
       visited[index] = true;
       distance_map[index] = 0;
--- a/beluga/include/beluga/sensor/data/occupancy_grid.hpp
+++ b/beluga/include/beluga/sensor/data/occupancy_grid.hpp
@@ -68,7 +68,8 @@ namespace beluga {
  *   `g.coordinates_for(r, f)` returns a range of embedding space coordinates in the
  *   corresponding frame as `Eigen::Vector2d` values;
  * - `g.free_cells()` returns a range of `std::size_t` indices to free grid cells;
- * - `g.obstacle_data()` returns a range of `bool` values, representing grid cell occupancy;
+ * - `g.obstacle_mask()` returns a range of `bool` values, representing grid cell occupancy;
+ * - `g.unknown_mask()` returns a range of `bool` values, representing the unkwnown space of the grid cells;
  */
 
 /// Occupancy 2D grid base type.
@@ -174,11 +175,19 @@ class BaseOccupancyGrid2 : public BaseLinearGrid2<Derived> {
            ranges::views::transform([](const auto& tuple) { return std::get<0>(tuple); });
   }
 
-  /// Retrieves grid data using true booleans for obstacles.
-  [[nodiscard]] auto obstacle_data() const {
+  /// Retrieves a mask over occupied cells in the grid.
+  [[nodiscard]] auto obstacle_mask() const {
     return this->self().data() |
            ranges::views::transform([value_traits = this->self().value_traits()](const auto& value) {
              return value_traits.is_occupied(value);
+           });
+  }
+
+  /// Retrieves a mask over unknown cells in the grid.
+  [[nodiscard]] auto unknown_mask() const {
+    return this->self().data() |
+           ranges::views::transform([value_traits = this->self().value_traits()](const auto& value) {
+             return value_traits.is_unknown(value);
            });
   }
 };
--- a/beluga/include/beluga/sensor/likelihood_field_model.hpp
+++ b/beluga/include/beluga/sensor/likelihood_field_model.hpp
@@ -187,9 +187,8 @@ class LikelihoodFieldModel {
     if (params.model_unknown_space) {
       const double inverse_max_distance = 1 / params.max_laser_distance;
       const double background_distance = -two_squared_sigma * std::log((inverse_max_distance - offset) / amplitude);
-      const auto mask_value = static_cast<ranges::range_value_t<decltype(distance_map)>>(background_distance);
 
-      distance_map |= beluga::actions::overlay(grid.unknown_mask(), mask_value);
+      distance_map |= beluga::actions::overlay(grid.unknown_mask(), background_distance);
     }
 
     auto likelihood_values = std::move(distance_map) |                               //
--- a/beluga/include/beluga/sensor/likelihood_field_model.hpp
+++ b/beluga/include/beluga/sensor/likelihood_field_model.hpp
@@ -187,8 +187,9 @@ class LikelihoodFieldModel {
     if (params.model_unknown_space) {
       const double inverse_max_distance = 1 / params.max_laser_distance;
       const double background_distance = -two_squared_sigma * std::log((inverse_max_distance - offset) / amplitude);
+      const auto mask_value = static_cast<ranges::range_value_t<decltype(distance_map)>>(background_distance);
 
-      distance_map |= beluga::actions::overlay(grid.unknown_mask(), background_distance);
+      distance_map |= beluga::actions::overlay(grid.unknown_mask(), mask_value);
     }
 
     auto likelihood_values = std::move(distance_map) |                               //
--- a/beluga/include/beluga/sensor/likelihood_field_model.hpp
+++ b/beluga/include/beluga/sensor/likelihood_field_model.hpp
@@ -24,6 +24,7 @@
 #include <beluga/algorithm/distance_map.hpp>
 #include <beluga/sensor/data/occupancy_grid.hpp>
 #include <beluga/sensor/data/value_grid.hpp>
+#include <range/v3/action/transform.hpp>
 #include <range/v3/range/conversion.hpp>
 #include <range/v3/view/all.hpp>
 #include <range/v3/view/transform.hpp>
@@ -173,7 +174,7 @@ class LikelihoodFieldModel {
 
     const double offset = params.z_random / params.max_laser_distance;
 
-    const auto to_likelihood = [=](double squared_distance) {
+    const auto to_likelihood = [amplitude, two_squared_sigma, offset](double squared_distance) {
       return amplitude * std::exp(-squared_distance / two_squared_sigma) + offset;
     };
 
@@ -190,10 +191,11 @@ class LikelihoodFieldModel {
       distance_map |= beluga::actions::overlay(grid.unknown_mask(), background_distance);
     }
 
-    std::transform(distance_map.begin(), distance_map.end(), distance_map.begin(), truncate_to_max_distance);
-    std::transform(distance_map.begin(), distance_map.end(), distance_map.begin(), to_likelihood);
+    auto likelihood_values = std::move(distance_map) |                               //
+                             ranges::actions::transform(truncate_to_max_distance) |  //
+                             ranges::actions::transform(to_likelihood);
 
-    return ValueGrid2<float>{std::move(distance_map), grid.width(), grid.resolution()};
+    return ValueGrid2<float>{std::move(likelihood_values), grid.width(), grid.resolution()};
   }
 };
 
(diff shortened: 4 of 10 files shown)


beluga/include/beluga/sensor/data/occupancy_grid.hpp
```diff
@@ -181,6 +181,13 @@ class BaseOccupancyGrid2 : public BaseLinearGrid2<Derived> {
              return value_traits.is_occupied(value);
            });
   }
+
+  [[nodiscard]] auto unknown_obstacle_data() const {
+    return this->self().data() |
+           ranges::views::transform([value_traits = this->self().value_traits()](const auto& value) {
+             return std::make_tuple(value_traits.is_occupied(value), value_traits.is_unknown(value));
+           });
+  }
```
hidmic, line 190: @DPR00 I'm not a huge fan of using data types with logically invalid states (what does `(true, true)` mean?). I'd rather use an `enum` or a fully fledged [`tribool`](https://www.boost.org/doc/libs/1_61_0/doc/html/boost/logic/tribool.html) equivalent. Also, perhaps `ternary_data()` would be more accurate for a name.

beluga/include/beluga/sensor/likelihood_field_model.hpp
```diff
@@ -58,6 +58,10 @@ struct LikelihoodFieldModelParam {
    * Used to calculate the probability of the obstacle being hit.
    */
   double sigma_hit = 0.2;
+  /// Pre-computed distance map for unkwnown spaces
+  double unknown_distance_map =
+      -2 * sigma_hit * sigma_hit *
+      std::log(std::sqrt(2 * Sophus::Constants<double>::pi()) * (1 - z_random) / max_laser_distance / z_hit);
```
hidmic, line 64: @DPR00 consider using the same auxiliary variables as for the forward calculation for readability.

beluga/include/beluga/algorithm/distance_map.hpp
```diff
@@ -92,6 +92,53 @@ auto nearest_obstacle_distance_map(
...
+  };
+  auto queue = std::priority_queue<IndexPair, std::vector<IndexPair>, decltype(compare)>{compare};
+
+  for (auto [index, tuple_data] : ranges::views::enumerate(obstacle_map)) {
+    if (std::get<0>(tuple_data)) {
+      visited[index] = true;
+      distance_map[index] = 0;
+      queue.push(IndexPair{index, index});  // The nearest obstacle is itself
+    } else if (std::get<1>(tuple_data)) {
+      visited[index] = true;
```
hidmic, line 122: @DPR00 hmm, won't this partially (or totally) block the greedy expansion below? Thinking about this again, it may be safer to override the distance map at the indices we know are actually unknown after computing it in full. We are doing this once, so the performance penalty shouldn't be that bad.

hidmic (commented):
First pass!

> It would be nice to make some maps testing the draft. How could I perform that?

I recommend extending the likelihood field model unit tests for this next.

beluga/include/beluga/sensor/data/occupancy_grid.hpp
DPR00: That's true. I'm going for `enum class`, since it would be more readable.

DPR00:
Hi @hidmic! I'm trying to replicate locally the 'build_and_test.sh' file, since it is failing in the pipeline. I wanted to tested first before updating the MR. However, I'm getting the following error (similar to #422, but it actually build when using `-colcon build --packages-up-to beluga_example --cmake-args -DBUILD_TESTING=OFF` as indicated in Documentation):

```bash
-- Configuring incomplete, errors occurred!
---
--- stderr: beluga
CMake Error at test/CMakeLists.txt:30 (find_package):
  By not providing "Findbenchmark.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "benchmark", but CMake did not find one.

  Could not find a package configuration file provided by "benchmark" with
  any of the following names:

    benchmarkConfig.cmake
    benchmark-config.cmake

  Add the installation prefix of "benchmark" to CMAKE_PREFIX_PATH or set
  "benchmark_DIR" to a directory containing one of the above files.  If
  "benchmark" provides a separate development package or SDK, be sure it has
  been installed.
---
```

This is happening in `jazzy` distro, but it actually works on `humble`. I also tried running it from the main branch, but it keeps throwing that error. Should I arise an issue?

beluga/test/beluga/include/beluga/test/static_occupancy_grid.hpp
```diff
@@ -31,6 +31,15 @@ class StaticOccupancyGrid : public BaseOccupancyGrid2<StaticOccupancyGrid<Rows,
     [[nodiscard]] bool is_free(bool value) const { return !value; }
     [[nodiscard]] bool is_unknown(bool) const { return false; }
     [[nodiscard]] bool is_occupied(bool value) const { return value; }
+    [[nodiscard]] beluga::GridStatus get_status(bool value) const {
```
hidmic, line 34: @DPR00 consider moving this over to the `BaseOccupancyGrid` class. The fewer the requirements on the user the better.

beluga/include/beluga/sensor/likelihood_field_model.hpp
```diff
@@ -58,6 +58,14 @@ struct LikelihoodFieldModelParam {
...
    */
   double sigma_hit = 0.2;
+  /// Pre-computed two_squared_sigma (auxiliar variable)
+  double two_squared_sigma = 2 * sigma_hit * sigma_hit;
+  /// Pre-computed amplitude (auxiliar variable)
+  double amplitude = z_hit / (sigma_hit * std::sqrt(2 * Sophus::Constants<double>::pi()));
+  /// Pre-computed offset (auxiliar variable)
+  double offset = z_random / max_laser_distance;
+  /// Pre-computed distance map for unkwnown spaces
+  double unknown_distance_map = -two_squared_sigma * std::log((1 / max_laser_distance - offset) / z_hit / amplitude);
```
hidmic, line 68: @DPR00 hmm, why `z_hit` again? Also, consider giving `1 / max_laser_distance` a name.

beluga/include/beluga/sensor/likelihood_field_model.hpp
```diff
@@ -58,6 +58,14 @@ struct LikelihoodFieldModelParam {
    * Used to calculate the probability of the obstacle being hit.
    */
   double sigma_hit = 0.2;
+  /// Pre-computed two_squared_sigma (auxiliar variable)
+  double two_squared_sigma = 2 * sigma_hit * sigma_hit;
```
hidmic, line 62: @DPR00 this won't be recomputed if the user instantiates `LikelihoodFieldModelParam` and then sets parameters. Consider moving this over to the `LikelihoodFieldSensor` class instead, no need for this unknown distance gimmick to be exposed to the user.

beluga/include/beluga/sensor/data/occupancy_grid.hpp
```diff
@@ -181,6 +184,14 @@ class BaseOccupancyGrid2 : public BaseLinearGrid2<Derived> {
              return value_traits.is_occupied(value);
            });
   }
+
+  /// Retrieves grid data using GridStatus enum class (Free, Unknown or Occupied) for obstacles.
+  [[nodiscard]] auto unknown_obstacle_data() const {
```
hidmic, line 189: @DPR00 not sold on the the name yet. `unknown_obstacle_data` gives no hint as to what you are getting from it.

beluga/include/beluga/sensor/data/occupancy_grid.hpp
```diff
@@ -34,6 +34,9 @@
 
 namespace beluga {
 
+/// General value traits
+enum class GridStatus : std::uint8_t { kFree, kUnknown, kOccupied };
```
hidmic, line 38: @DPR00 do we need to pollute the `beluga` namespace with this, or can we nest it within the class?

beluga/include/beluga/algorithm/distance_map.hpp
```diff
@@ -92,6 +93,56 @@ auto nearest_obstacle_distance_map(
...
+    for (const std::size_t index : neighbors_function(parent.index)) {
+      if (!visited[index]) {
+        visited[index] = true;
+        distance_map[index] = distance_function(parent.nearest_obstacle_index, index);
+        queue.push(IndexPair{parent.nearest_obstacle_index, index});
+      }
+    }
+  }
+
+  for (auto [index, enum_data] : ranges::views::enumerate(obstacle_map)) {
```
hidmic, line 137: @DPR00 meta: can't we decouple this so that we reuse `nearest_obstacle_distance_map`? If distance map matches 1-to-1 with grid data, why not zip it with grid ternary value and transform it downstream?

hidmic (commented):
`codecov` isn't happy because we are lacking tests :)

beluga/include/beluga/sensor/likelihood_field_model.hpp
DPR00: I was thinking about moving it as protected inside the `LikelihoodFieldModel` class, WDYT?

DPR00: Moving it inside the `make_likelihood_field` function, since there are used just there.

beluga/include/beluga/sensor/data/occupancy_grid.hpp
DPR00: what about `data_status()`?

hidmic: It's really a ternary view, as opposed to the binary view that `obstacle_data` provides. Maybe we should move in that direction? `binary_data()` and `ternary_data()`? @nahueespinosa WDYT?

nahueespinosa: I don't have great ideas for this. I like `ternary_data()`, maybe that's all we need here?

Otherwise, which flavor of binary data we want to favor? Just one or all three of them?
 - `obstacle_data()` (only occupied data)
 - `free_space_data()` (only free data)
 - `unknown_data()` (only unknown data)

beluga/include/beluga/algorithm/distance_map.hpp
DPR00: This made me realize that we may not need `ternary_data()`. PR updated.

beluga/include/beluga/algorithm/distance_map.hpp
```diff
@@ -92,6 +92,31 @@ auto nearest_obstacle_distance_map(
...
+
+    for (auto [index, is_unknown] : ranges::views::enumerate(uknown_map)) {
+      if (is_unknown) {
+        distance_map[index] = unknown_distance_map;
+      }
+    }
+  }
+
+  return distance_map;
+}
```
hidmic, line 118: @DPR00 meta: I'm on the fence about having a separate function with a similar name but not quite, taking a grid instead of a range. Following the idea of the mask below, I feel that something like:

```c++
  auto distance_map = nearest_obstacle_distance_map(grid.obstacle_mask(), distance_function, neighbors_function);
  if (params_.model_unknown_space) {
      const double background_distance = /* compute it */ ;
      distance_map |= beluga::actions::overlay(grid.unknown_mask(), background_distance);
  }
```
would much more generally applicable. I can see the notion of a mask or stencil coming up again later on. How do you feel about implementing a new action? [`beluga::actions::normalize`](https://github.com/Ekumen-OS/beluga/blob/main/beluga/include/beluga/actions/normalize.hpp) is a good reference for this.

hidmic:
@DPR00 we should update the PR title.

beluga/include/beluga/sensor/likelihood_field_model.hpp
```diff
@@ -144,30 +146,42 @@ class LikelihoodFieldModel {
...
                                       std::size_t first, std::size_t second) {
       return std::min((grid.coordinates_at(first) - grid.coordinates_at(second)).squaredNorm(), squared_max_distance);
     };
 
+    /// Pre-computed variables
+    const double two_squared_sigma = 2 * params.sigma_hit * params.sigma_hit;
+    const double amplitude = params.z_hit / (params.sigma_hit * std::sqrt(2 * Sophus::Constants<double>::pi()));
+    const double offset = params.z_random / params.max_laser_distance;
+    const double inverse_max_distance = 1 / params.max_laser_distance;
+    const double unknown_distance_map = -two_squared_sigma * std::log((inverse_max_distance - offset) / amplitude);
```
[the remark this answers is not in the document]
DPR00: `background_distance` seems more appropriate.

beluga/include/beluga/sensor/data/occupancy_grid.hpp
```diff
@@ -181,6 +181,14 @@ class BaseOccupancyGrid2 : public BaseLinearGrid2<Derived> {
              return value_traits.is_occupied(value);
            });
   }
+
+  /// Retrieves grid data using true booleans for unknowns.
+  [[nodiscard]] auto unknown_data() const {
```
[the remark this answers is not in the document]
DPR00: It makes sense, since both return only boolean values and not actual data.

beluga/include/beluga/algorithm/distance_map.hpp
DPR00: I think I got the idea. It looks cleaner and much more generally applicable, so I will implement the action.

DPR00: PR updated :)

beluga/include/beluga/sensor/data/occupancy_grid.hpp
```diff
@@ -174,13 +175,21 @@ class BaseOccupancyGrid2 : public BaseLinearGrid2<Derived> {
            ranges::views::transform([](const auto& tuple) { return std::get<0>(tuple); });
   }
 
-  /// Retrieves grid data using true booleans for obstacles.
-  [[nodiscard]] auto obstacle_data() const {
+  /// Retrieves a mask of true booleans to focus on obstacles data in the grid.
```
hidmic, line 178: @DPR00 nit: 
```suggestion
  /// Retrieves a mask over occupied cells in the grid.
```

beluga/include/beluga/sensor/data/occupancy_grid.hpp
```diff
@@ -174,13 +175,21 @@ class BaseOccupancyGrid2 : public BaseLinearGrid2<Derived> {
...
+  [[nodiscard]] auto obstacle_mask() const {
     return this->self().data() |
            ranges::views::transform([value_traits = this->self().value_traits()](const auto& value) {
              return value_traits.is_occupied(value);
            });
   }
+
+  /// Retrieves a mask of true booleans to focus on unkwnown data in the grid.
```
hidmic, line 186: @DPR00 nit: 
```suggestion
  /// Retrieves a mask over unknown cells in the grid.
```

beluga/include/beluga/sensor/likelihood_field_model.hpp
```diff
@@ -58,6 +59,8 @@ struct LikelihoodFieldModelParam {
    * Used to calculate the probability of the obstacle being hit.
    */
   double sigma_hit = 0.2;
+  /// Bool parameter to decide if unknown spaces should be explored
```
hidmic, line 62: @DPR00 nit: 
```suggestion
  /// Whether to model unknown space or assume it free.
```

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,103 @@
...
+  }
+};
+
+/// Implementation detail for an overlay range adaptor object with a default execution policy.
+struct overlay_fn : public overlay_base_fn {
+  using overlay_base_fn::operator();
+
+  /// Overload that defines a default execution policy.
+  template <class Range, class MaskRange, std::enable_if_t<ranges::range<Range>, int> = 0>
+  constexpr auto operator()(Range& range, MaskRange& mask, double value) const -> Range& {
```
hidmic, line 81: @DPR00 let's make the `value` type a template parameter. It may not necessarily be a `double`.

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,103 @@
...
+  constexpr auto operator()(ExecutionPolicy&& policy, Range& range, MaskRange& mask, double value) const -> Range& {
+    auto map = range | ranges::views::common;
+
+    std::transform(
+        policy,            //
+        std::begin(map),   //
+        std::end(map),     //
+        std::begin(mask),  //
+        std::begin(map),   //
+        [value](const auto& m, bool flag) { return flag ? value : m; });
```
hidmic, line 60: @DPR00 nit:
```suggestion
        [value](const auto& base_value, bool flag) { return flag ? mask_value : base_value; });
```

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,103 @@
...
+   * \param mask The mask where the values will be overlaid
+   * \param value The value to be overlaid
+   */
+  template <
+      class ExecutionPolicy,
+      class Range,
+      class MaskRange,
+      std::enable_if_t<std::is_execution_policy_v<std::decay_t<ExecutionPolicy>>, int> = 0,
+      std::enable_if_t<ranges::range<Range>, int> = 0>
+  constexpr auto operator()(ExecutionPolicy&& policy, Range& range, MaskRange& mask, double value) const -> Range& {
```
hidmic, line 51: @DPR00 nit: 
```suggestion
  constexpr auto operator()(ExecutionPolicy&& policy, Range& range, MaskRange& mask, double mask_value) const -> Range& {
```

hidmic (commented):
This is amazing @DPR00, great job!

We should add tests now though, for the likelihood model and for the action we just added.

hidmic:
This is no draft anymore, it's solid PR material.

DPR00:
Awesome @hidmic, thanks! I will be addressing the comments during the following days and adding tests, as well.
I also want to test it in a simulated world. Any recommendations of a world different than `turtlebot_world`?

beluga/test/beluga/sensor/test_lfm_with_unknown_space.cpp
```diff
@@ -0,0 +1,271 @@
...
+    -1 , -1 , -1 , 100, 100,
+    -1 , 0  , 0  , 0  , 100,
+    -1 , 0  , 0  , 0  , 100,
+    100, 0  , 0  , 0  , 100,
+    100, 100, 100, 100, 100},
+    kResolution};
+  // clang-format on
+
+  const double param_mld = 20.0;
+  const double unknown_likelihood = 1 / param_mld;
```
hidmic, line 47: @DPR00 nit: consider making these `constexpr` and switching to Hungarion notation e.g. `kUnknownSpaceLikelihood`.

hidmic: Also, please don't use non-obvious abbreviations. I don't know what `param_mld` is. Characters are free. Spell it out in full.

beluga/test/beluga/sensor/test_lfm_with_unknown_space.cpp
```diff
@@ -0,0 +1,271 @@
...
+  const auto params = beluga::LikelihoodFieldModelParam{2.0, param_mld, 0.5, 0.5, 0.2, true};
```
hidmic, line 48: @DPR00 nit: here and elsewhere, use named constants for parameters. I don't know what the first `0.5` argument stands for on first look. If it was `constexpr auto kLaserHitProbability = 0.5;` and `beluga::LikelihoodFieldModelParam{2.0, param_mld, kLaserHitProbability, 0.5, 0.2, true};` , then I would.

beluga/test/beluga/sensor/test_lfm_with_unknown_space.cpp
```diff
@@ -0,0 +1,271 @@
...
+  // the likelihood field model includes an optimization that stores the likelihood elevated to the cube
+  auto expected_cubed_likelihood =
+      expected_likelihood_field | ranges::views::transform([](auto v) { return v * v * v; }) | ranges::to<std::vector>;
+
+  ASSERT_THAT(
+      sensor_model.likelihood_field().data(),
+      testing::Pointwise(testing::DoubleNear(0.003), expected_cubed_likelihood));
+}
+
+TEST(LikelihoodFieldModelUnknownSpace, ImportanceWeight) {
```
hidmic, line 106: @DPR00 it would be nice to have a test with a 1-point scan that hits an unknown cell, to show it in fact returns `kUnknownSpaceLikelihood`.

hidmic (commented):
Third pass I think. Really good work @DPR00.

hidmic:
This PR needs a rebase after #433.

beluga/test/beluga/sensor/test_lfm_with_unknown_space.cpp
DPR00: I add the following tests:
```cpp
  {
    auto state_weighting_function = sensor_model(std::vector<std::pair<double, double>>{{0, 0}});
    ASSERT_NEAR(
        1.000 + kUnknownSpaceLikelihood * kUnknownSpaceLikelihood * kUnknownSpaceLikelihood,
        state_weighting_function(grid.origin()), 0.003);
  }

  {
    auto state_weighting_function = sensor_model(std::vector<std::pair<double, double>>{{2.25, 2.25}});
    ASSERT_NEAR(
        1.000 + kUnknownSpaceLikelihood * kUnknownSpaceLikelihood * kUnknownSpaceLikelihood,
        state_weighting_function(grid.origin()), 0.003);
  }
```
Keep in mind that `kUnknownSpaceLikelihood * kUnknownSpaceLikelihood * kUnknownSpaceLikelihood` is neglictible compared to `1.000`, that is why I had written `1.000` previously.

DPR00:
I did the rebase and reduce all the changes to one commit. The latter was due there were a lot of commits where the changes are no longer valid.

beluga/test/beluga/sensor/data/test_occupancy_grid.cpp
```diff
@@ -79,15 +79,29 @@ TEST(OccupancyGrid2, AllFreeCells) {
...
-  ASSERT_THAT(grid.obstacle_data() | ranges::to<std::vector>, testing::Pointwise(testing::Eq(), grid.data()));
+  ASSERT_THAT(grid.obstacle_mask() | ranges::to<std::vector>, testing::Pointwise(testing::Eq(), grid.data()));
+#pragma GCC diagnostic pop
+}
+
+TEST(OccupancyGrid2, UnknownMask) {
+  const auto grid = StaticOccupancyGrid<5, 2, std::int8_t>{{-1, -1, -1, -1, 0, 0, 0, 0, 100, 100}};
+
+  const auto expected_unknown_mask =
+      std::vector<bool>{true, true, true, true, false, false, false, false, false, false};
```
hidmic, line 98: @DPR00 nit^2: you can drop the `<bool>` thingie, it will get deduced.

hidmic (approved):
I left a few more comments and CI is not happy, but this is a great piece of work @DPR00.

I'm happy with it once the missing bits are addressed and CI goes green (looks like `clang-tidy` is not happy).

beluga/include/beluga/sensor/likelihood_field_model.hpp
```diff
@@ -161,20 +164,32 @@ class LikelihoodFieldModel {
...
+    const auto to_likelihood = [=](double squared_distance) {
       return amplitude * std::exp(-squared_distance / two_squared_sigma) + offset;
     };
 
     const auto neighborhood = [&grid](std::size_t index) { return grid.neighborhood4(index); };
 
     // determine distances to obstacles and calculate likelihood values in-place
     // to minimize memory usage when dealing with large maps
-    auto likelihood_values = nearest_obstacle_distance_map(grid.obstacle_data(), squared_distance, neighborhood);
+    auto likelihood_values = nearest_obstacle_distance_map(grid.obstacle_mask(), squared_distance, neighborhood);
```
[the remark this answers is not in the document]
DPR00: It end up being likelihood values after:
```cpp
std::transform(likelihood_values.begin(), likelihood_values.end(), likelihood_values.begin(), to_likelihood);
```
If I change the value to `distance_map`, then should I create a new variable called `likelihood_values` after the previous transform?

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,126 @@
...
+  /// Overload that re-orders arguments from an action closure.
+  template <
+      class ExecutionPolicy,
+      class Range,
+      class MaskRange,
+      class Mask = ranges::range_value_t<Range>,
+      std::enable_if_t<std::is_execution_policy_v<ExecutionPolicy>, int> = 0,
+      std::enable_if_t<ranges::range<Range>, int> = 0,
+      std::enable_if_t<ranges::range<MaskRange>, int> = 0>
+  constexpr auto operator()(Range&& range, MaskRange& mask, Mask mask_value, ExecutionPolicy policy) const -> Range& {
```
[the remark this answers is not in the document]
DPR00: I am having issues with this :thinking:. Compiler throws errors when I forward `MaskRange`, e.g.:
```bash
error: cannot bind non-const lvalue reference of type ‘std::vector<bool, std::allocator<bool> >&’ to an rvalue of type ‘std::vector<bool, std::allocator<bool> >’
```

beluga/include/beluga/sensor/likelihood_field_model.hpp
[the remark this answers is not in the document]
nahueespinosa: I think that we could potentially write something like this:

```cpp
auto distance_map = nearest_obstacle_distance_map(grid.obstacle_mask(), squared_distance, neighborhood);

if (params.model_unknown_space) {
    const double inverse_max_distance = 1 / params.max_laser_distance;
    const double background_distance = -two_squared_sigma * std::log((inverse_max_distance - offset) / amplitude);

    distance_map |= beluga::actions::overlay(grid.unknown_mask(), background_distance);
}

ranges::transform(distance_map, distance_map.begin(), truncate_to_max_distance);

auto& likelihood_values = distance_map;  // re-use storage
ranges::transform(distance_map, likelihood_values.begin(), to_likelihood);

return ValueGrid2<float>{std::move(likelihood_values), grid.width(), grid.resolution()};
```

---

P.D. I'm also not a fan of defining a named lambda to use later in the code. But this might not be everyone's cup of tea.

```cpp
auto distance_map = nearest_obstacle_distance_map(grid.obstacle_mask(), squared_distance, neighborhood);

if (params.model_unknown_space) {
    const double inverse_max_distance = 1 / params.max_laser_distance;
    const double background_distance = -two_squared_sigma * std::log((inverse_max_distance - offset) / amplitude);

    distance_map |= beluga::actions::overlay(grid.unknown_mask(), background_distance);
}

ranges::transform(distance_map, distance_map.begin(),
  [squared_max_distance](double squared_distance) {
    return std::min(squared_distance, squared_max_distance);
  });

auto& likelihood_values = distance_map;  // re-use storage

ranges::transform(distance_map, likelihood_values.begin(),
  [amplitude, two_squared_sigma, offset](double squared_distance) {
    return amplitude * std::exp(-squared_distance / two_squared_sigma) + offset;
  });

return ValueGrid2<float>{std::move(likelihood_values), grid.width(), grid.resolution()};
```

beluga/include/beluga/sensor/likelihood_field_model.hpp
```diff
@@ -161,20 +164,32 @@ class LikelihoodFieldModel {
...
+    /// Pre-computed variables
+    const double two_squared_sigma = 2 * params.sigma_hit * params.sigma_hit;
+    assert(two_squared_sigma > 0.0);
+
+    const double amplitude = params.z_hit / (params.sigma_hit * std::sqrt(2 * Sophus::Constants<double>::pi()));
+    assert(amplitude > 0.0);
+
+    const double offset = params.z_random / params.max_laser_distance;
+
+    const auto to_likelihood = [=](double squared_distance) {
```
nahueespinosa, line 176: I'd suggest capturing only what you need here. It is better to express intent. Copying everything that's in scope is not great (it could potentially include `grid` although inlining probably solves all issues).

beluga/include/beluga/sensor/likelihood_field_model.hpp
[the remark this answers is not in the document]
nahueespinosa: After a good night of sleep, I realized that the move-transform-assign pattern can actually be written like this:

```cpp
auto distance_map = nearest_obstacle_distance_map(grid.obstacle_mask(), squared_distance, neighborhood);

if (params.model_unknown_space) {
    const double inverse_max_distance = 1 / params.max_laser_distance;
    const double background_distance = -two_squared_sigma * std::log((inverse_max_distance - offset) / amplitude);

    distance_map |= beluga::actions::overlay(grid.unknown_mask(), background_distance);
}

auto likelihood_values = std::move(distance_map) |
                         ranges::actions::transform(truncate_to_max_distance) |
                         ranges::actions::transform(to_likelihood);

return ValueGrid2<float>{std::move(likelihood_values), grid.width(), grid.resolution()};
```

This should re-use the storage.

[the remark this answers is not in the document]
DPR00: Compiler is throwing an error (in testing) when trying to use `std::move()` with `|` operator. The error is due, and cite `view.hpp`:
```
            // **************************************************************************
            // *    When piping a range into an adaptor, the range must satisfy the     *
            // *    "viewable_range" concept. A range is viewable when either or both   *
            // *    of these things are true:                                           *
            // *      - The range is an lvalue (not a temporary object), OR             *
            // *      - The range is a view (not a container).                          *
            // **************************************************************************
```
Doing the following works just fine:
```cpp
auto likelihood_values = distance_map | ranges::views::transform(truncate_to_max_distance) |
                     ranges::views::transform(to_likelihood) | ranges::to<std::vector<float>>;
```
However, this won't re-use the storage.

I tried to find a workaround using `ranges::view::all(std::move(distance_map))`, but it didn't work either.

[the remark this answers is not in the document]
nahueespinosa: Yeah, you should use `ranges::actions::transform` instead of `ranges::views::transform`.

You will also need to include `#include <range/v3/action/transform.hpp>`.

And finally, you can force the linter to keep the newline by adding an empty comment to the EOL in a range pipeline. Personally I think this is much better for readability.

```cpp
auto likelihood_values = distance_map |                                          //
                         ranges::actions::transform(truncate_to_max_distance) |  //
                         ranges::actions::transform(to_likelihood);
```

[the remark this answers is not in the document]
DPR00: > Yeah, you should use `ranges::actions::transform` instead of `ranges::views::transform`.

My bad, I didn't notice it. It's working now :tada:. I've just updated the PR.

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,130 @@
...
+  using overlay_base_fn::operator();
+
+  /// Overload that defines a default execution policy.
+  template <
+      class Range,
+      class MaskRange,
+      class Mask = ranges::range_value_t<Range>,
+      std::enable_if_t<ranges::range<Range>, int> = 0,
+      std::enable_if_t<ranges::range<MaskRange>, int> = 0>
+  constexpr auto operator()(Range&& range, MaskRange&& mask, const Mask& mask_value) const -> Range& {
```
hidmic, line 106: @DPR00 hmm shouldn't this be `Mask&& mask_value` and be `std::forward`ed instead? I actually wonder how the compiler doesn't complain about `std::move`ing a const ref :thinking:

hidmic (approved):
Plain amazing. I'm good with this PR as-is. @nahueespinosa ?

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,130 @@
...
+   */
+  template <
+      class ExecutionPolicy,
+      class Range,
+      class MaskRange,
+      class Mask = ranges::range_value_t<Range>,
+      std::enable_if_t<std::is_execution_policy_v<std::decay_t<ExecutionPolicy>>, int> = 0,
+      std::enable_if_t<ranges::range<Range>, int> = 0,
+      std::enable_if_t<ranges::range<MaskRange>, int> = 0>
+  constexpr auto operator()(ExecutionPolicy&& policy, Range& range, const MaskRange& mask, const Mask& mask_value) const
```
nahueespinosa, line 53:
```suggestion
  constexpr auto operator()(ExecutionPolicy&& policy, Range& range, MaskRange&& mask, Mask&& mask_value) const
```

beluga/include/beluga/actions/overlay.hpp
nahueespinosa: It compiles because all the overloads take `mask_value` by const ref, so we never actually move.

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,130 @@
...
+  constexpr auto operator()(Range&& range, MaskRange&& mask, const Mask& mask_value) const -> Range& {
+    return (*this)(
+        std::execution::seq, std::forward<Range>(range), std::forward<MaskRange>(mask), std::move(mask_value));
+  }
+
+  /// Overload that returns an action closure to compose with other actions.
+  template <class MaskRange, class Mask, std::enable_if_t<ranges::range<MaskRange>, int> = 0>
+  constexpr auto operator()(MaskRange&& mask, const Mask& mask_value) const {
+    return ranges::make_action_closure(
+        ranges::bind_back(overlay_fn{}, std::forward<MaskRange>(mask), std::move(mask_value)));
```
nahueespinosa, line 115:
```suggestion
        ranges::bind_back(overlay_fn{}, std::forward<MaskRange>(mask), std::forward<Mask>(mask_value)));
```

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,130 @@
...
+      std::enable_if_t<ranges::range<Range>, int> = 0,
+      std::enable_if_t<ranges::range<MaskRange>, int> = 0>
+  constexpr auto operator()(Range&& range, MaskRange&& mask, const Mask& mask_value) const -> Range& {
+    return (*this)(
+        std::execution::seq, std::forward<Range>(range), std::forward<MaskRange>(mask), std::move(mask_value));
+  }
+
+  /// Overload that returns an action closure to compose with other actions.
+  template <class MaskRange, class Mask, std::enable_if_t<ranges::range<MaskRange>, int> = 0>
+  constexpr auto operator()(MaskRange&& mask, const Mask& mask_value) const {
```
nahueespinosa, line 113:
```suggestion
  constexpr auto operator()(MaskRange&& mask, Mask&& mask_value) const {
```

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,130 @@
...
+  /// Overload that defines a default execution policy.
+  template <
+      class Range,
+      class MaskRange,
+      class Mask = ranges::range_value_t<Range>,
+      std::enable_if_t<ranges::range<Range>, int> = 0,
+      std::enable_if_t<ranges::range<MaskRange>, int> = 0>
+  constexpr auto operator()(Range&& range, MaskRange&& mask, const Mask& mask_value) const -> Range& {
+    return (*this)(
+        std::execution::seq, std::forward<Range>(range), std::forward<MaskRange>(mask), std::move(mask_value));
```
nahueespinosa, line 108:
```suggestion
        std::execution::seq, std::forward<Range>(range), std::forward<MaskRange>(mask), std::forward<Mask>(mask_value));
```

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+/// Implementation detail for an overlay range adaptor object with a default execution policy.
+struct overlay_fn : public overlay_base_fn {
+  using overlay_base_fn::operator();
+
+  /// Overload that defines a default execution policy.
+  template <
+      class Range,
+      class MaskRange,
+      class Mask = ranges::range_value_t<Range>,
```
nahueespinosa, line 103: We need to let the compiler deduce this for perfect forwarding.
```suggestion
      class Mask,
```

beluga/include/beluga/actions/overlay.hpp
nahueespinosa, line 106:
```suggestion
  constexpr auto operator()(Range&& range, MaskRange&& mask, Mask&& mask_value) const -> Range& {
```

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  /// Overload that returns an action closure to compose with other actions.
+  template <
+      class ExecutionPolicy,
+      class MaskRange,
+      class Mask,
+      std::enable_if_t<std::is_execution_policy_v<ExecutionPolicy>, int> = 0,
+      std::enable_if_t<ranges::range<MaskRange>, int> = 0>
+  constexpr auto operator()(ExecutionPolicy policy, MaskRange&& mask, const Mask& mask_value) const {
```
nahueespinosa, line 89:
```suggestion
  constexpr auto operator()(ExecutionPolicy policy, MaskRange&& mask, Mask&& mask_value) const {
```

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,130 @@
...
+    return ranges::make_action_closure(
+        ranges::bind_back(overlay_base_fn{}, std::forward<MaskRange>(mask), std::move(mask_value), std::move(policy)));
```
nahueespinosa, line 91:
```suggestion
        ranges::bind_back(overlay_base_fn{}, std::forward<MaskRange>(mask), std::forward<Mask>(mask_value), std::move(policy)));
```

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,130 @@
...
+      class ExecutionPolicy,
+      class Range,
+      class MaskRange,
+      class Mask = ranges::range_value_t<Range>,
+      std::enable_if_t<std::is_execution_policy_v<ExecutionPolicy>, int> = 0,
+      std::enable_if_t<ranges::range<Range>, int> = 0,
+      std::enable_if_t<ranges::range<MaskRange>, int> = 0>
+  constexpr auto operator()(Range&& range, MaskRange&& mask, const Mask& mask_value, ExecutionPolicy policy) const
+      -> Range& {
+    return (*this)(std::move(policy), std::forward<Range>(range), std::forward<MaskRange>(mask), std::move(mask_value));
```
nahueespinosa, line 79:
```suggestion
    return (*this)(std::move(policy), std::forward<Range>(range), std::forward<MaskRange>(mask), std::forward<Mask>(mask_value));
```

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+    return range;
+  }
+
+  /// Overload that re-orders arguments from an action closure.
+  template <
+      class ExecutionPolicy,
+      class Range,
+      class MaskRange,
+      class Mask = ranges::range_value_t<Range>,
```
nahueespinosa, line 73:
```suggestion
      class Mask,
```

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,130 @@
...
+      std::enable_if_t<std::is_execution_policy_v<ExecutionPolicy>, int> = 0,
+      std::enable_if_t<ranges::range<Range>, int> = 0,
+      std::enable_if_t<ranges::range<MaskRange>, int> = 0>
+  constexpr auto operator()(Range&& range, MaskRange&& mask, const Mask& mask_value, ExecutionPolicy policy) const
```
nahueespinosa, line 77:
```suggestion
  constexpr auto operator()(Range&& range, MaskRange&& mask, Mask&& mask_value, ExecutionPolicy policy) const
```

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,130 @@
...
+   * \param policy The execution policy to use.
+   * \param range An existing range to apply this action to.
+   * \param mask The mask where the values will be overlaid.
+   * \param mask_value The value to be overlaid.
+   */
+  template <
+      class ExecutionPolicy,
+      class Range,
+      class MaskRange,
+      class Mask = ranges::range_value_t<Range>,
```
nahueespinosa, line 49:
```suggestion
      class Mask,
```

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,130 @@
...
+      -> Range& {
+    auto map = range | ranges::views::common;
+
+    std::transform(
+        policy,            //
+        std::begin(map),   //
+        std::end(map),     //
+        std::begin(mask),  //
+        std::begin(map),   //
+        [mask_value](const auto& base_value, bool flag) { return flag ? mask_value : base_value; });
```
nahueespinosa, line 63:
```suggestion
        [&mask_value](const auto& base_value, bool flag) { return flag ? mask_value : base_value; });
```

beluga/test/beluga/actions/test_overlay.cpp
```diff
@@ -0,0 +1,135 @@
+// Copyright 2024 Ekumen, Inc.
```
nahueespinosa, line 1: It would be good to have a test that creates an action with a temporary mask, and uses it in a separate line:

```cpp
TEST(OverlayAction, Temporary) {
  auto input = std::vector{...};

  auto action = beluga::actions::overlay(std::vector{...}, true);

  input |= action;
  ASSERT_THAT(input, ::testing::Pointwise(::testing::Eq(), mask));
}
```

nahueespinosa (approved):
LGTM! After fixing forwarding references.

beluga/include/beluga/actions/overlay.hpp
DPR00: If I remove this, how could I make `Mask` to be derived from `Range`? Discussed [here](https://github.com/Ekumen-OS/beluga/pull/430#discussion_r1770736903).

nahueespinosa: So, if you don't remove this, the compiler won't deduce the type and `std::forward` won't work as expected.
If you want the conversion to only happen once, you can do this just before calling `std::transform`:

```cpp
const auto converted_mask_value = static_cast<ranges::range_value_t<Range>>(mask_value);
```

DPR00: Interesting. I haven't realized about all the details behind forwarding. Thanks @nahueespinosa, TIL!

beluga/include/beluga/actions/overlay.hpp
```diff
@@ -0,0 +1,131 @@
...
+      class Range,
+      class MaskRange,
+      class Mask,
+      std::enable_if_t<std::is_execution_policy_v<std::decay_t<ExecutionPolicy>>, int> = 0,
+      std::enable_if_t<ranges::range<Range>, int> = 0,
+      std::enable_if_t<ranges::range<MaskRange>, int> = 0>
+  constexpr auto operator()(ExecutionPolicy&& policy, Range& range, MaskRange&& mask, Mask&& mask_value) const
+      -> Range& {
+    auto map = range | ranges::views::common;
+
```
nahueespinosa, line 56: This is a better place to add the conversion:
```cpp
const auto converted_mask_value = static_cast<ranges::range_value_t<Range>>(mask_value);
```

beluga/include/beluga/sensor/likelihood_field_model.hpp
```diff
@@ -161,23 +165,36 @@ class LikelihoodFieldModel {
...
-    auto likelihood_values = nearest_obstacle_distance_map(grid.obstacle_data(), squared_distance, neighborhood);
-    std::transform(
-        likelihood_values.begin(), likelihood_values.end(), likelihood_values.begin(), truncate_to_max_distance);
-    std::transform(likelihood_values.begin(), likelihood_values.end(), likelihood_values.begin(), to_likelihood);
+    auto distance_map = nearest_obstacle_distance_map(grid.obstacle_mask(), squared_distance, neighborhood);
+
+    if (params.model_unknown_space) {
+      const double inverse_max_distance = 1 / params.max_laser_distance;
+      const double background_distance = -two_squared_sigma * std::log((inverse_max_distance - offset) / amplitude);
+      const auto mask_value = static_cast<ranges::range_value_t<decltype(distance_map)>>(background_distance);
```
nahueespinosa, line 190: Let's move this to the overlay implementation and merge!
See https://github.com/Ekumen-OS/beluga/pull/430/files#r1788603115.

nahueespinosa (approved):
LGTM!

hidmic (approved):
@DPR00 go go go :rocket:

DPR00:
I am not able to merge it because I need to be an authorized user. Could you merge it @hidmic, pls? :rocket:

beluga/include/beluga/sensor/likelihood_field_model.hpp
```diff
@@ -161,23 +165,35 @@ class LikelihoodFieldModel {
...
     // determine distances to obstacles and calculate likelihood values in-place
     // to minimize memory usage when dealing with large maps
-    auto likelihood_values = nearest_obstacle_distance_map(grid.obstacle_data(), squared_distance, neighborhood);
-    std::transform(
-        likelihood_values.begin(), likelihood_values.end(), likelihood_values.begin(), truncate_to_max_distance);
-    std::transform(likelihood_values.begin(), likelihood_values.end(), likelihood_values.begin(), to_likelihood);
+    auto distance_map = nearest_obstacle_distance_map(grid.obstacle_mask(), squared_distance, neighborhood);
+
+    if (params.model_unknown_space) {
+      const double inverse_max_distance = 1 / params.max_laser_distance;
```
glpuga-eka, line 188: @DPR00 @hidmic I'm going through this code to help rodrigo with https://github.com/Ekumen-OS/beluga/pull/577, but I can't fully make sense of this.

After a while of thinking this go a long way around assigning `1 / params.max_laser_distance` likelihood to any unknown cell. The latest version of this blurs this even more because that value may be overriden [here](https://github.com/Ekumen-OS/beluga/blob/main/beluga/include/beluga/sensor/likelihood_field_model_base.hpp#L178), which may be right or may be wrong depending what we model here.

So my question is: why  `1 / params.max_laser_distance`? That models a uniform distribution over the lenght of the lidar assuming with 100% certainty that this is a random measurement in that range. Why is this not `offset = params.z_random / params.max_laser_distance` which is the overall probability of a random measurement falling here?

DPR00: IIRC this was took from this part of Probabilistic Robotics book (Section 6.4.2, page 173):
image

hidmic: That. IIUC the idea is that the probability of a hit for any given measurement is uniform: $z \in [0, z_{max}]$, then $p(z) = \frac{1}{z_{max}}$). Mixing parameters are not applicable because those presume knowledge we don't have (well, kinda, we could still model failure and unexpected objects, I think, but they didn't and we don't).

glpuga-eka: I don't think that's right, and I explain why.

The problem is that that causes the likelihood map to implement different likelihood functions depending on the lidar happens to hit, causing non-uniform weight scales. Depending on the lidar range the likelihood might be higher through an open window than 20cm away from an actually mapped obstacle.

> Mixing parameters are not applicable because those presume knowledge we don't have

I think that _not_ mixing assumes a knowledge we don't have. It asummes we know with 100% certainty that the measurement IS a random hit, which we can't know. Where a hit falls does not depend as much on the measurement as much as on the hyphotesis/particle you are tracing it from, so a hit on unknown area behind a wall surface or through a window most likely just means a hyphotesis which is 5 cm closer to the wall than the actual pose of the lidar. That hypothesis is not _wrong_, it's just part of the internal representation of the distribution that the filter uses for the pose.

If we assign `1 / params.max_laser_distance`, we are overweighing particles with beams that hit within unknown areas, because `z_random < 1.0`,

If we assign a weight `params.z_random / params.max_laser_distance` to that case, that's the weight we assign to random impacts everywhere, and **indeed that's how we model "random obstacles that are not mapped"**, which makes sense for unknown areas. 

Notice that while I think that using `params.z_random / params.max_laser_distance` is at least in line with the larger model and does not distort the likelihood distribution value scale, **I still don't think it is right**. 

That's because the likelihood model only really models known obstacles; more than a probabilistic model, LF is a "scan correlator". In fact, the `z_random` term is not really there to decide on the "goodness" of an impact over another, since it's the same everywhere; it's just there to avoid the likelihood product going to zero too fast if we hit unmapped obstacles.

In ignoring the `z_hit` term behind a wall we are biasing the estimation solution towards the external side of the wall by truncating the natural symmetry of the distance-likelihood distribution on both sides of the surface. In closed quarters this may not make much of a difference, but if you only see a straight wall on your left, this alone may add you a few centimeters of error due to that bias this causes in the particle distribution.

hidmic: > If we assign a weight params.z_random / params.max_laser_distance to that case, that's the weight we assign to random impacts everywhere, and indeed that's how we model "random obstacles that are not mapped", which makes sense for unknown areas.

Hmm, but random hit != unknown hit. Random hits have lower probability because it is implied they are transient. Thrun mentions cross-talk and echoes as potential sources. 

That said...

> In ignoring the z_hit term behind a wall we are biasing the estimation solution towards the external side of the wall by truncating the natural symmetry of the distance-likelihood distribution on both sides of the surface.

Yeah, I see the problem now :thinking:  

(stream of consciousness ahead) 

IIUC we are folding both unobservable and unmapped as unknown, and we are overpowering the base likelihood model when we shouldn't. Assigning a higher likelihood to hits within internal walls or past external walls just gives us bias. Assigning a higher likelihood to hits within an unmapped but otherwise visible region does prevent us from excessively penalizing hypothesis close to such regions when those regions happen to be occupied. If the unmapped space is within a room that had its door closed during mapping, we do want hits within that unexplored space not be ruled out as highly unlikely.

Hmm, what if we let the user configure a distance threshold? Unexplored space that is within that distance from occupied cells decays normally, while unexplored space past that distance gets the truncated likelihood to signal we don't really know if a hit in that region is likely or not. We cover for the variance in the estimate while not (excessively) penalizing hitting a car parked out the window.

glpuga: I created https://github.com/Ekumen-OS/beluga/issues/582 to further discuss this.

PR #433, opened 2024-09-23
Fix memory usage problem in 2D likelihood sensor model

glpuga:
### Proposed changes

Initially, this changes likelihood storage from double to float, inspired by https://github.com/ros-navigation/navigation2/pull/4426 . 

More importantly, this fixes a bug in the `make_likelihood_field()` function which causes memory footprint to go through the roof for large maps.

Basically, because we applied the `max_obstacle_distance` threshold **before** calculating the distance map,  for distances larger than  `max_obstacle_distance`  the priority queue in `nearest_obstacle_distance_map()` was processing  the map in a random fashion instead of the "advancing wavefront" that the algorithm is meant to do. This caused the priority queue length to be at least as large as the map itself  and even larger for some maps.

This change does not affect either the accuracy or the average CPU usage.

##### Before 

6000x6000 pixels map @ 0.05 cm/px (300m x 300m)

Screenshot from 2024-09-23 09-45-21

##### After

Same 6000x6000 pixels map @ 0.05 cm/px (300m x 300m)

Screenshot from 2024-09-23 09-45-28

#### Type of change

### Checklist

### Additional comments

### DIFF
--- a/beluga/include/beluga/sensor/likelihood_field_model.hpp
+++ b/beluga/include/beluga/sensor/likelihood_field_model.hpp
@@ -120,19 +120,17 @@ class LikelihoodFieldModel {
       const auto unknown_space_occupancy_prob = 1. / params_.max_laser_distance;
       // TODO(glpuga): Investigate why AMCL and QuickMCL both use this formula for the weight.
       // See https://github.com/Ekumen-OS/beluga/issues/153
-      const auto unknown_space_occupancy_likelihood_cubed =
-          unknown_space_occupancy_prob * unknown_space_occupancy_prob * unknown_space_occupancy_prob;
       return std::transform_reduce(
           points.cbegin(), points.cend(), 1.0, std::plus{},
-          [this, x_offset, y_offset, cos_theta, sin_theta,
-           unknown_space_occupancy_likelihood_cubed](const auto& point) {
+          [this, x_offset, y_offset, cos_theta, sin_theta, unknown_space_occupancy_prob](const auto& point) {
             // Transform the end point of the laser to the grid local coordinate system.
             // Not using Eigen/Sophus because they make the routine x10 slower.
             // See `benchmark_likelihood_field_model.cpp` for reference.
             const auto x = point.first * cos_theta - point.second * sin_theta + x_offset;
             const auto y = point.first * sin_theta + point.second * cos_theta + y_offset;
-            // for performance, we store the likelihood already elevated to the cube
-            return likelihood_field_.data_near(x, y).value_or(unknown_space_occupancy_likelihood_cubed);
+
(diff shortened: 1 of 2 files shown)


beluga/include/beluga/sensor/likelihood_field_model.hpp
```diff
@@ -120,19 +120,17 @@ class LikelihoodFieldModel {
...
             // Transform the end point of the laser to the grid local coordinate system.
             // Not using Eigen/Sophus because they make the routine x10 slower.
             // See `benchmark_likelihood_field_model.cpp` for reference.
             const auto x = point.first * cos_theta - point.second * sin_theta + x_offset;
             const auto y = point.first * sin_theta + point.second * cos_theta + y_offset;
-            // for performance, we store the likelihood already elevated to the cube
-            return likelihood_field_.data_near(x, y).value_or(unknown_space_occupancy_likelihood_cubed);
+
+            const auto cubed = [](auto pz) { return pz * pz * pz; };
+            return cubed(likelihood_field_.data_near(x, y).value_or(unknown_space_occupancy_prob));
```
hidmic, line 133: @glpuga hmm, if we are moving the power law to runtime to avoid the loss of the precision, we should make sure `pz` is is treated as a `double`. IIUC [`std::optional::value_or`](https://en.cppreference.com/w/cpp/utility/optional/value_or) will narrow cast `unknown_space_occupancy_prob`, not the other way around. 

So how about having `cubed` take a `double` explicitly? That should do.

hidmic (approved):
Great find @glpuga !

beluga/include/beluga/sensor/likelihood_field_model.hpp
glpuga: There are no accuracy or performance implications in this change. I moved it here because the place where it was before was purely premature optimization, since it makes no mensurable runtime difference, and yet the likelihood map is not a likelihood map if the value is cubed (against least surprise).

> IIUC [std::optional::value_or](https://en.cppreference.com/w/cpp/utility/optional/value_or) will narrow cast unknown_space_occupancy_prob, not the other way around.

`unknown_space_occupancy_prob` will be narrow cast, and then the result of data_near will be broadened again to double. I can move the narrowcast out of the loop, but doing it once when `unknown_space_occupancy_prob` is calculated.

I'll also rephrase this bit so that its more similar to the code in the beam model.

Check d9212cd2 .

hidmic (approved):
LGTM

beluga/include/beluga/sensor/likelihood_field_model.hpp
hidmic: > and yet the likelihood map is not a likelihood map if the value is cubed (against least surprise).

Fair point.

PR #441, opened 2024-10-05
Rewrite normalize action

nahueespinosa:
### Proposed changes

Trying a new way of implementing actions that gets rid of the argument re-ordering madness and is hopefully easier to maintain.

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/actions/normalize.hpp
+++ b/beluga/include/beluga/actions/normalize.hpp
@@ -51,7 +51,7 @@ struct normalize_closure {
   constexpr normalize_closure(ExecutionPolicy policy, double factor) : policy_{std::move(policy)}, factor_{factor} {}
 
   template <class Range>
-  constexpr auto operator()(Range& range) const {
+  constexpr auto operator()(Range& range) const -> Range& {
     static_assert(ranges::forward_range<Range>);
 
     auto weights = std::invoke([&range]() {
@@ -146,7 +146,7 @@ struct normalize_fn {
  * If none is specified, the default normalization factor corresponds to the total sum of weights
  * in the given range.
  */
-inline constexpr detail::normalize_fn normalize;
+inline constexpr ranges::actions::action_closure<detail::normalize_fn> normalize;
 
 }  // namespace beluga::actions
 
--- a/beluga/include/beluga/actions/normalize.hpp
+++ b/beluga/include/beluga/actions/normalize.hpp
@@ -17,6 +17,7 @@
 
 #include <algorithm>
 #include <execution>
+#include <optional>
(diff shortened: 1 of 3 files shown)


hidmic (commented):
Overall LGTM. The one thing I don't love is the API break. I'd rather define and immediately deprecate pipe operators to keep the no argument variant operational instead of dropping.

nahueespinosa:
> The one thing I don't love is the API break

Fixed in https://github.com/Ekumen-OS/beluga/pull/441/commits/185f744f89d142cfe1c13450b63647bb537d1fbb.

beluga/include/beluga/actions/normalize.hpp
```diff
@@ -34,120 +35,106 @@ namespace beluga::actions {
...
-    }();
+    });
+
+    const double factor = std::invoke([this, weights]() {
+      if (factor_.has_value()) {
+        return factor_.value();
+      }
+
+      return ranges::accumulate(weights, 0.0);  // The default normalization factor is the total sum of weights.
+    });
```
DPR00, line 71: Could this be replaced by:
```cpp
const double factor = factor_.value_or(ranges::accumulate(weights, 0.0));
```
?

beluga/include/beluga/actions/normalize.hpp
```diff
@@ -34,120 +35,106 @@ namespace beluga::actions {
...
     std::transform(
-        policy,               //
-        std::begin(weights),  //
-        std::end(weights),    //
-        std::begin(weights),  //
+        policy_,          //
+        weights.begin(),  //
+        weights.end(),    //
+        weights.begin(),  //
         [factor](const auto w) { return w / factor; });
```
DPR00, line 82: meta: Should we add some kind of `assert` to ensure that factor won't take a zero value? Probably when weights is empty.

DPR00 (commented):
Awesome! All the actions should take this structure, right? If so, I can help with the overlay action.

beluga/include/beluga/actions/normalize.hpp
nahueespinosa: That would accumulate the weights every time, and then decide whether to use a user-provided factor or the sum. This version only accumulates weights if there is no user-provided factor.

I'd like to use C++20's monadic operations, but this is C++17. See [`or_else`](https://en.cppreference.com/w/cpp/utility/optional/or_else).

hidmic (approved):

nahueespinosa:
> All the actions should take this structure, right? If so, I can help with the overlay action.

Yeah, that would be great!

PR #442, opened 2024-10-07
Rewrite take_evenly view

nahueespinosa:
### Proposed changes

Avoids using a range pipeline to implement the take_evenly view which is not great for compile times.
This implementation also makes it a random access view when the underlying range is random access.

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/views/take_evenly.hpp
+++ b/beluga/include/beluga/views/take_evenly.hpp
@@ -30,63 +30,143 @@ namespace beluga::views {
 
 namespace detail {
 
-/// Implementation detail for a take_evenly range adaptor object.
-struct take_evenly_fn {  // NOLINT(readability-identifier-naming)
-  /// Overload that implements the take_evenly algorithm.
-  /**
-   * \tparam Range A [sized range](https://en.cppreference.com/w/cpp/ranges/sized_range).
-   * \param range Source range from where to take elements.
-   * \param count Number of elements to take.
-   *
-   * If `count` or the range size are zero, it returns an empty range.
-   * If `count` is greater than the range size, it returns all the elements.
-   * The first and last elements of the range are always included.
-   */
-  template <class Range>
-  constexpr auto operator()(Range&& range, std::size_t count) const {
-    const std::size_t size = ranges::size(range);
+/// \cond detail
 
-    const auto filter_function = [size, count](const auto& pair) {
-      if ((size == 0UL) || (count == 0UL)) {
-        return false;
-      }
+template <class Range>
+struct take_evenly_view : public ranges::view_facade<take_evenly_view<Range>, ranges::finite> {
+ public:
+  static_assert(ranges::view_<Range>);
+  static_assert(ranges::forward_range<Range>);
+  static_assert(ranges::sized_range<Range>);
 
-      if (count > size) {
-        return true;
-      }
+  constexpr take_evenly_view() = default;
 
-      const auto [index, _] = pair;
-      if (count == 1UL) {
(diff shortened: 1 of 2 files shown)


beluga/include/beluga/views/take_evenly.hpp
```diff
@@ -30,64 +30,145 @@ namespace beluga::views {
...
-      const std::size_t m0 = (index - 1UL) * (count - 1UL) / (size - 1UL);
-      const std::size_t m1 = index * (count - 1UL) / (size - 1UL);
-      return m0 != m1;
-    };
-
-    // `cache1` ensures that views prior to `filter` in the pipeline are iterated exactly once.
-    // This is needed because filter needs to dereference the input iterator twice.
-    return ranges::views::enumerate(range) | ranges::views::cache1 | ranges::views::filter(filter_function) |
-           beluga::views::elements<1>;
+    return take_evenly_view{std::forward<Range>(range), count};
```
nahueespinosa, line 162: Slightly helps with #385 by reducing the max RSS when compiling using a single thread.

Before: Maximum resident set size (kbytes): 2703864
After: Maximum resident set size (kbytes): 2640764

beluga/test/beluga/views/test_take_evenly.cpp
```diff
@@ -15,84 +15,160 @@
...
+  static_assert(ranges::bidirectional_range<decltype(output)>);
+
+  static_assert(!ranges::random_access_range<decltype(input)>);
+  static_assert(!ranges::random_access_range<decltype(output)>);
+}
+
 TEST(TakeEvenlyView, NoElementsTakeZero) {
   const auto input = std::vector<int>{};
-  const auto output = input | beluga::views::take_evenly(0) | ranges::to<std::vector>;
+  auto output = input | beluga::views::take_evenly(0);
```
hidmic, line 75: @nahueespinosa why not `const` anymore? Same elsewhere, except perhaps for random access tests.

beluga/include/beluga/views/take_evenly.hpp
```diff
@@ -30,64 +30,144 @@ namespace beluga::views {
...
+struct take_evenly_view : public ranges::view_facade<take_evenly_view<Range>, ranges::finite> {
+ public:
+  static_assert(ranges::view_<Range>);
+  static_assert(ranges::forward_range<Range>);
+  static_assert(ranges::sized_range<Range>);
+
+  constexpr take_evenly_view() = default;
+
+  constexpr take_evenly_view(Range range, std::size_t count)
+      : range_{std::move(range)}, count_{count}, size_{ranges::size(range_)} {}
```
hidmic, line 45: @nahueespinosa hmm, range copies? Are we positively sure this will always be cheap?

nahueespinosa: `Range` is constrained to be a `view` at this point. Views are supposed to be cheaply copyable or move-only.

beluga/test/beluga/views/test_take_evenly.cpp
nahueespinosa: It comes from a time when the view cached something. I now used const when possible. There are only two exceptions:

1. `iota` can't be const as it is a range factory with mutable state.
2. `reverse` can't be const because it caches.

beluga/include/beluga/views/take_evenly.hpp
hidmic: Ah, missed the `static_assert` on first read. Cool.

hidmic (approved):
LGTM

PR #448, opened 2024-10-22
Expose model_unknown_space parameter to ROS2

DPR00:
### Proposed changes

Expose `model_unknown_space` parameter of the Likelihood Field sensor model to Beluga AMCL (ROS 2). 

#### Type of change

### Checklist

### Additional comments

### DIFF
--- a/beluga_amcl/src/amcl_node.cpp
+++ b/beluga_amcl/src/amcl_node.cpp
@@ -187,17 +187,23 @@ AmclNode::AmclNode(const rclcpp::NodeOptions& options) : BaseAMCLNode{"amcl", ""
 
 AmclNode::~AmclNode() {
   RCLCPP_INFO(get_logger(), "Destroying");
-  // In case this lifecycle node wasn't properly shut down, do it here
+  // In case this lifecycle node wasn't properly shut down, do it here.
   on_shutdown(get_current_state());
 }
 
 void AmclNode::do_activate(const rclcpp_lifecycle::State&) {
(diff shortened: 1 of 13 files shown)


beluga_amcl/src/amcl_node.cpp
```diff
@@ -120,6 +120,12 @@ AmclNode::AmclNode(const rclcpp::NodeOptions& options) : BaseAMCLNode{"amcl", ""
     declare_parameter("z_rand", rclcpp::ParameterValue(0.5), descriptor);
   }
 
+  {
+    auto descriptor = rcl_interfaces::msg::ParameterDescriptor();
+    descriptor.description = "If false, Likelihood Field sensor model won't model unknown space.";
```
hidmic, line 125: @DPR00 consider having matching descriptions in ROS 1 and ROS 2. I like the ROS 1 one better.

beluga_example/params/default.ros2.yaml
```diff
@@ -61,6 +61,8 @@ amcl:
     z_hit: 0.5
     # Weight used to combine the probability of random noise in perception.
     z_rand: 0.5
+    # Whether to model unknown space or assume it free.
+    model_unknown_space : false
```
hidmic, line 65: @DPR00 nit: 
```suggestion
    model_unknown_space: false
```

hidmic (commented):
Looks about correct. Have we tested it in simulation or against some dataset already?

PR #451, opened 2024-10-27
Clean up NDT implementation and examples

### LINKED ISSUE #424: Add option to remove the adaptiveness to the core filter pipeline
### Feature description

Some map representations don't allow (at least not in a straightforward, practical way) sampling randomly from the map.
This is the case, just to name one, of sparse voxel grids, where there're no notion of free space, only occupied space.

Given this difficulty, the adaptive part of the AMCL pipeline has limited effect, because of not being able to generate random states from the map.

For these cases, an option where adaptiveness is turned off could be of use, and remove the need of forcefully provide a random state generator.
https://github.com/Ekumen-OS/beluga/blob/7a01f39bc78a6124fe0609637b657da0ded9bc6c/beluga/include/beluga/algorithm/amcl_core.hpp#L108

## Definition of done

Add the feature described above, and make sure this is the option we use for all the nodes where randomly sampling from the map doens't make sense.

As of today, `ndt` and `ndt_3d` variants.

nahueespinosa:
### Proposed changes

Simplifies the NDT localization nodes by reducing the number of supported features, specifically the available motion models and the recovery strategy configuration. This change eliminates the need for static dispatch and `std::visit`, easing some pressure on the compiler. Additionally, it cleans up header file inclusions.

Closes #424, since it removes the recovery strategy from the core AMCL pipeline.
Related to #385, since it reduces memory usage during compilation.

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/algorithm/amcl_core.hpp
+++ b/beluga/include/beluga/algorithm/amcl_core.hpp
@@ -21,12 +21,25 @@
 #include <utility>
 #include <vector>
 
-#include <beluga/beluga.hpp>
-
-#include <range/v3/range/concepts.hpp>
-#include <range/v3/view/any_view.hpp>
+#include <range/v3/range/conversion.hpp>
 #include <range/v3/view/take_exactly.hpp>
-#include "beluga/policies/on_motion.hpp"
+
+#include <beluga/actions/assign.hpp>
+#include <beluga/actions/normalize.hpp>
+#include <beluga/actions/propagate.hpp>
+#include <beluga/actions/reweight.hpp>
+#include <beluga/algorithm/estimation.hpp>
+#include <beluga/algorithm/spatial_hash.hpp>
+#include <beluga/containers/circular_array.hpp>
+#include <beluga/containers/tuple_vector.hpp>
+#include <beluga/policies/every_n.hpp>
+#include <beluga/policies/on_effective_size_drop.hpp>
+#include <beluga/policies/on_motion.hpp>
+#include <beluga/primitives.hpp>
+#include <beluga/random/multivariate_normal_distribution.hpp>
+#include <beluga/views/random_intersperse.hpp>
+#include <beluga/views/sample.hpp>
+#include <beluga/views/take_while_kld.hpp>
 
 namespace beluga {
 
@@ -44,10 +57,6 @@ struct AmclParams {
   std::size_t min_particles = 500UL;
   /// Maximum number of particles in the filter at any point in time.
   std::size_t max_particles = 2000UL;
-  /// Used as part of the recovery mechanism when considering how many random particles to insert.
-  double alpha_slow = 0.001;
-  /// Used as part of the recovery mechanism when considering how many random particles to insert.
-  double alpha_fast = 0.1;
   /// Used as part of the kld sampling mechanism.
   double kld_epsilon = 0.05;
   /// Used as part of the kld sampling mechanism.
@@ -58,37 +67,19 @@ struct AmclParams {
 /**
  * \tparam MotionModel Class representing a motion model. Must satisfy \ref MotionModelPage.
  * \tparam SensorModel Class representing a sensor model. Must satisfy \ref SensorModelPage.
- * \tparam RandomStateGenerator A callable able to produce random states, optionally based on the current particles
- * state.
- * Class 'T' is a valid RandomStateGenerator if given 't' a possibly const instance of 'T' any of the following
- * conditions are true:
- * - t can be called without arguments returning an instance of 'SensorModel::state_type'
- *   representing a random state.
- * - t can be called with `const beluga::TupleVector<ParticleType>&> returning a callable
- *   that can be called without arguments and return an instance of 'SensorModel::state_type'.
- * \tparam WeightT Type to represent a weight of a particle.
  * \tparam ParticleType Full particle type, containing state, weight and possibly
  * other information .
- * \tparam ExecutionPolicy Execution policy for particles processing.
  */
 template <
     class MotionModel,
     class SensorModel,
-    class RandomStateGenerator,
-    typename WeightT = beluga::Weight,
-    class ParticleType = std::tuple<typename SensorModel::state_type, WeightT>,
-    class ExecutionPolicy = std::execution::sequenced_policy>
+    class Particle = std::tuple<typename SensorModel::state_type, beluga::Weight>>
 class Amcl {
-  static_assert(
-      std::is_same_v<ExecutionPolicy, std::execution::parallel_policy> or
-      std::is_same_v<ExecutionPolicy, std::execution::sequenced_policy>);
-
-  using particle_type = ParticleType;
+  using particle_type = Particle;
   using measurement_type = typename SensorModel::measurement_type;
   using state_type = typename SensorModel::state_type;
   using map_type = typename SensorModel::map_type;
   using spatial_hasher_type = spatial_hash<state_type>;
-  using random_state_generator_type = RandomStateGenerator;
   using estimation_type = std::invoke_result_t<beluga::detail::estimate_fn, std::vector<state_type>>;
 
  public:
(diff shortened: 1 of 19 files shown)


nahueespinosa:
Code coverage reports show -0.01%, but that seems like a numerical bug:
image

beluga/include/beluga/sensor/data/ndt_cell.hpp
```diff
@@ -33,7 +33,7 @@ namespace beluga {
 
 /// Representation for a cell of a N dimensional NDT cell.
 template <int NDim, typename Scalar = double>
-struct NDTCell {
+struct NdtCell {
```
hidmic, line 36: @nahueespinosa meta: why the name change? NDT is an acronym. If we were to spell it out in full it would read `NormalDistributionTransformCell`, emphasis on capital letters. Same elsewhere.

beluga_amcl/src/ndt_amcl_node.cpp
```diff
@@ -178,81 +132,52 @@ auto NdtAmclNode::get_initial_estimate() const -> std::optional<std::pair<Sophus
   return std::make_pair(pose, covariance);
 }
 
-auto NdtAmclNode::get_motion_model() const -> MotionModelVariant {
+auto NdtAmclNode::get_motion_model() const -> beluga::DifferentialDriveModel2d {
   const auto name = get_parameter("robot_model_type").as_string();
-  if (name == kDifferentialModelName) {
+  if (name == kDifferentialModelName || name == kNav2DifferentialModelName) {
```
hidmic, line 137: @nahueespinosa meta: why adding `kNav2DifferentialModelName` here? The AMCL node needs it to ensure parity with Nav2 AMCL, but the NDT node does not have such constraint.

beluga/include/beluga/algorithm/amcl_core.hpp
```diff
@@ -58,66 +67,40 @@ struct AmclParams {
...
  */
 template <
     class MotionModel,
     class SensorModel,
-    class RandomStateGenerator,
-    typename WeightT = beluga::Weight,
-    class ParticleType = std::tuple<typename SensorModel::state_type, WeightT>,
-    class ExecutionPolicy = std::execution::sequenced_policy>
+    class Particle = std::tuple<typename SensorModel::state_type, beluga::Weight>>
 class Amcl {
```
hidmic, line 77: @nahueespinosa meta: there is redundancy between `beluga::Amcl` and `beluga_ros::Amcl`. I recall we did not get as far as to deduplicate them, but that'd be the logical path. In which case, `beluga_ros::Amcl` is doing a lot more than it should, and with this change, `beluga::Amcl` would be doing less. Why remove the execution policy? Why remove (as opposed to generalize, if applicable) state jitter to avoid distribution collapse?

hidmic (commented):
First pass

beluga/include/beluga/sensor/data/ndt_cell.hpp
nahueespinosa: We say `NDTSensorModel` and `NDTCell`, but we also say `NdtAmclNode`. What should we do with the latter? `NDTAMCLNode`?

beluga_amcl/src/ndt_amcl_node.cpp
nahueespinosa: We use `nav2_amcl::DifferentialMotionModel` in the `default_ndt.ros2.yaml` file. Let me change that.

beluga/include/beluga/algorithm/amcl_core.hpp
nahueespinosa: I don't think that's the kind of generalization that leads to maintainable code.

State jitter does not apply to NDT models. IIUC, the 2D case was sampling from the particle set, which has no effect; and the 3D case was adding zero-initialized states, which is simply wrong. We never created a map distribution compatible with the NDT map representation.

In any case, I think we should treat the NDT examples as a nice proof of concept, and avoid adding unnecessary features. If we do want those features to be easily applicable (e.g. configurable execution policies), we should do more thinking before generalizing things.

nahueespinosa: Also, having this amount of `std::visit` calls and templates of templates, makes it really hard to play with explicit template instantiation. I really don't think that we should keep things as they are today.

nahueespinosa: The range pipeline refactor was meant to provide a way to leave behind the class templates and the mixin pattern. We are kind of going back to that if we want to provide an `Amcl` class that's general enough for any use case.

There are other approaches to composition that I think are worth exploring. I reckon I haven't had enough time to do that lately.

beluga/include/beluga/sensor/data/ndt_cell.hpp
hidmic: Hmm, mixed case acronyms don't stand out as much when parsing code. Alas, `NDTAMCLNode` _is_ almost excessive :thinking:

I like neither. Short term, let's just be consistent. Too many acronyms for a single class may suggest the class is bundling too much, but that's out of scope.

beluga/include/beluga/algorithm/amcl_core.hpp
hidmic: There's a lot to unpack there.

> State jitter does not apply to NDT models.

But it does for AMCL, at least for this instantiation of it. Whether NDT filters should use this AMCL variant or a different MCL variant is a completely valid yet separate discussion IMO.

> having this amount of std::visit calls and templates of templates makes it really hard to play with explicit template instantiation

That is a very good point. We ended up using `std::visit` to enable runtime configurations and to keep execution paths transparent to the compiler. We can push it to the edges, but I don't think we'll get to remove it. Can we?

> If we do want those features (e.g. configurable execution policies), we should do more thinking before generalizing things.

> I really don't think that we should keep things as they are today.

> We are kind of going back to that if we try to provide an Amcl class that's general enough for any use case.

> There are other approaches to composition that I think are worth exploring.

Hmm, all of that is valid, but that alternative design that is better than this doesn't exist yet. I give you that these AMCL classes are far from complete solutions, but walking back on them with nothing to show in exchange seems like a regression to me.

nahueespinosa: > But it does for AMCL, at least for this instantiation of it. Whether NDT filters should use this AMCL variant or a different MCL variant is a completely valid yet separate discussion IMO.

On second read, I didn't remove state jitter to avoid distribution collapse. That's taken care of by the motion model. I removed the recovery strategy based on injecting random particles from the map (which none of the NDT variants were actually doing).

"Adaptive" in AMCL originally stands for the fact that the sample size can change based on the KLD metric. The recovery strategy was not mentioned in the [original paper](https://www.robots.ox.ac.uk/~cvrg/hilary2005/adaptive.pdf), it comes from Prob Rob.

> Can we?

Yes, with carefully implemented type erasure and benchmarking. I don't know how much benefit we get from inlining the sensor and motion models, we should measure that. I've also seen really interesting techniques to make type-erasure really efficient. The only drawback is not being able to inline, which might be a big deal in some scenarios. We should investigate.

Range pipelines on the other hand will suffer greatly without inlining. But that's not what we are trying to customize at runtime.

> Hmm, all of that is valid, but that alternative design that is better than this doesn't exist yet. 

Fair point.

> I give you that these AMCL classes are far from complete solutions, but walking back on them with nothing to show for in exchange seems like a regression to me.

I understand your point, but it seems to me that we took a lot of shortcuts to get to this point and this is only one small step back. I'm not removing vital features, the two classes still implement (a version of) AMCL, and this unblocks experimentation to reduce memory usage at build time.

hidmic: > I removed the recovery strategy based on injecting random particles from the map

FWIW that still counts as roughening to cope with (particle depletion and) distribution collapse. Artificially increasing the variance of other distributions helps too but it is a different approach.

> I've also seen really interesting techniques to make type-erasure really efficient. The only drawback is not being able to inline, which might be a big deal in some scenarios.

Hmm, the only path I see that avoids explicit combinatorial explosion (as `std::visit` already explodes but we don't see it) is type erasing within loops. Performance will take a hit, though I guess we could argue that you can't have it both ways: it's either performance or flexibility. I'm OK with that so long as top performance is still achievable.

> it seems to me that we took a lot of shortcuts to get to this point

That's true. 

> this unblocks experimentation to reduce memory usage at build time.

Hmm, what's preventing explicit instantiation of the `beluga::Amcl` class in its current form? :thinking:

PR #453, opened 2024-12-03
Ndt map visualization

Fernando-Sanz:
### Proposed changes

NDT map visualization publisher. It uses ellipsoids to represent obstacles placed using the mean and the covariance matrix of each cell.

ndt_map_visualization_enlarged
>Obstacle representation made with green ellipsoids (scale factor is used).

#### Type of change

### Checklist


beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,74 @@
...
+    marker.action = beluga_ros::msg::Marker::ADD;
+
+    // Set the position
+    // TODO: check the mean's data type and how to access to its members
+    marker.pose.position.x = cell.mean[0];
+    marker.pose.position.y = cell.mean[1];
+    marker.pose.position.z = cell.mean[2];
+
+    // Set the scale based on the covariance
+    // TODO: check what values of the covariance should be used and if they need be changed before
```
brt-ivanpauno, line 55: you need to set rotation and scale based on the covariance matrix.
For that you need to:

* Get eigenvectors of the matrix. You can use eigen for that, I cannot access the docs page for some reason, but I will copy links when it's back online.
* Make a matrix with each eigenvector as a column (maybe eigen provides that already).
  Order the columns so it's a rotation matrix (det == 1 and not det == -1).
* That's your rotation matrix, you can convert that to a quaternion for the ros message.
* The eigenvalues should be the x, y, z scale.

@hidmic is that reasonable to you?

brt-ivanpauno: one thing, the eigenvectors solver may fail to converge to a solution, in that case we should have an alternative representation.
Probably, we should make it a cube of cell size, so the difference is obvious vs the ellipsoids.

But I don't think that should happen much in practice (or likely not at all) ....

beluga_amcl/include/beluga_amcl/ndt_amcl_node_3d.hpp
```diff
@@ -145,6 +145,10 @@ class NdtAmclNode3D : public BaseAMCLNode {
   std::optional<Sophus::SE3d> last_known_odom_transform_in_map_;
   /// Whether to broadcast transforms or not.
   bool enable_tf_broadcast_{false};
+
+  /// Map representation publisher.
+  //rclcpp_lifecycle::LifecyclePublisher<visualization_msgs::msg::MarkerArray>::SharedPtr map_visualization_pub_;
```
ivanpauno, line 150: remove commented out line

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * \file
+ * \brief Utilities for NDT I/O over ROS interfaces.
+ */
+
+namespace beluga_ros {
+
+/// 
+template <typename MapType, int NDim>
+//beluga_ros::msg::MarkerArray& assign_obstacle_map(beluga::SparseValueGrid<MapType, NDim> grid){
```
ivanpauno, line 30: remove commented out line

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,74 @@
...
+  {
+    beluga_ros::msg::Marker marker;
+    marker.header.frame_id = "map";
+    marker.id = idCount++;  // TODO: Compile and check if this is the way to set id
+    marker.ns = "obstacles";
+    marker.type = beluga_ros::msg::Marker::SPHERE;
+    marker.action = beluga_ros::msg::Marker::ADD;
+
+    // Set the position
+    // TODO: check the mean's data type and how to access to its members
```
ivanpauno, line 49: this is fine

```suggestion
```

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,74 @@
...
+  auto& map = grid.data();
+
+  beluga_ros::msg::MarkerArray message{};
+
+  int idCount = 0;
+  for (const auto& [key, cell] : map)
+  {
+    beluga_ros::msg::Marker marker;
+    marker.header.frame_id = "map";
+    marker.id = idCount++;  // TODO: Compile and check if this is the way to set id
```
ivanpauno, line 43: this is fine

```suggestion
```

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
ivanpauno, line 55: you need to set rotation and scale based on the covariance matrix.
For that you need to:

* Get eigenvectors of the matrix. You can use eigen for that, I cannot access the docs page for some reason, but I will copy links when it's back online.
* Make a matrix with each eigenvector as a column (maybe eigen provides that already).
  Order the columns so it's a rotation matrix (det == 1 and not det == -1).
* That's your rotation matrix, you can convert that to a quaternion for the ros message.
* The eigenvalues should be the x, y, z scale.

@hidmic is that reasonable to you?

ivanpauno: one thing, the eigenvectors solver may fail to converge to a solution, in that case we should have an alternative representation.
Probably, we should make it a cube of cell size, so the difference is obvious vs the ellipsoids.

But I don't think that should happen much in practice (or likely not at all) ....

hidmic: Sounds good to me @ivanpauno

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+/// 
+template <typename MapType, int NDim>
+//beluga_ros::msg::MarkerArray& assign_obstacle_map(beluga::SparseValueGrid<MapType, NDim> grid){
+beluga_ros::msg::MarkerArray assign_obstacle_map(const beluga::SparseValueGrid<MapType, NDim>& grid){
+
+  // Get data from the grid
+  auto& map = grid.data();
+
+  beluga_ros::msg::MarkerArray message{};
```
ivanpauno, line 36: to remove the previous map you can use:

```suggestion
  beluga_ros::msg::MarkerArray message{};
  {
    beluga_ros::msg::Marker marker;
    marker.ns = "obstacles";
    marker.action = beluga_ros::msg::Marker::REMOVE_ALL;
  }
```

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,76 @@
...
+  {
+    beluga_ros::msg::Marker marker;
+    marker.ns = "obstacles";
+    marker.action = beluga_ros::msg::Marker::DELETE;
```
ivanpauno, line 39:
```suggestion
    marker.action = beluga_ros::msg::Marker::DELETEALL;
```

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,76 @@
...
+  }
```
ivanpauno, line 40: you're missing adding this marker

```suggestion
    message.markers.push_back(marker);
  }
```

ivanpauno:
@Fernando-Sanz Linters are failing, check https://github.com/Ekumen-OS/beluga/blob/main/CONTRIBUTING.md#how-do-i-submit-a-good-code-contribution to install pre-commit.
After those steps, linters will run automatically each time you run `git commit`.
Run `pre-commit run -a` to fix your previous commits.

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,106 @@
...
+    const auto eigenvectors = eigenSolver.eigenvectors().real();
+    const auto eigenvalues = eigenSolver.eigenvalues();
+
+    // TODO: Discuss a better way to do this (threshold and permutation)
+    Eigen::Matrix3d rotationMatrix;
+    rotationMatrix << eigenvectors.col(0), eigenvectors.col(1), eigenvectors.col(2);
+    const auto scalevector = Eigen::Vector3d{eigenvalues.x(), eigenvalues.y(), eigenvalues.z()};
+
+    // Permutation
+    if (std::abs(rotationMatrix.determinant() - 1.0) > 1e-6) {
```
ivanpauno, line 68: maybe create an:

```cpp
template <typename RealScalar>
inline bool isApprox(RealScalar x, RealScalar y, Scalar prec = eigen::NumTraits<RealScalar>::dummy_precision()) {
  return std::abs(x -y) < prec;
}

File: beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp [hunk 10]
```diff
@@ -0,0 +1,106 @@
...
+      rotationMatrix << eigenvectors.col(1), eigenvectors.col(0), eigenvectors.col(2);
+      scalevector = Eigen::Vector3d{eigenvalues.y(), eigenvalues.x(), eigenvalues.z()};
+
+      if (std::abs(rotationMatrix.determinant() - 1.0) > 1e-6) {
+        rotationMatrix << eigenvectors.col(0), eigenvectors.col(2), eigenvectors.col(1);
+        scalevector = Eigen::Vector3d{eigenvalues.x(), eigenvalues.z(), eigenvalues.y()};
+      }
```
ivanpauno, line 75: you don't need to check again, one permutation is enough to change the sign

ivanpauno (commented):
Looks quite solid!

Could you upload a video or image of how the map is looking?

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * \file
+ * \brief Utilities for NDT I/O over ROS interfaces.
+ */
+
+namespace beluga_ros {
+
+///
+template <typename MapType, int NDim>
+beluga_ros::msg::MarkerArray assign_obstacle_map(const beluga::SparseValueGrid<MapType, NDim>& grid) {
```
ivanpauno, line 29: add docstrings, see for example https://github.com/Ekumen-OS/beluga/blob/08e682fe0795952ad01fc991979fe3056a27c29a/beluga_ros/include/beluga_ros/particle_cloud.hpp#L96-L102

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,132 @@
...
+#include <beluga_ros/messages.hpp>
+
+/**
+ * \file
+ * \brief Utilities for NDT I/O over ROS interfaces.
+ */
+
+namespace beluga_ros {
+
+static void use_mean_covariance(const auto& mean, const auto& eigenSolver, auto& marker){
```
ivanpauno, line 27:
```suggestion
inline void use_mean_covariance(const auto& mean, const auto& eigenSolver, auto& marker){
```

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,132 @@
...
+  marker.pose.orientation.y = rotation.y();
+  marker.pose.orientation.z = rotation.z();
+  marker.pose.orientation.w = rotation.w();
+
+  marker.scale.x = scalevector.x();
+  marker.scale.y = scalevector.y();
+  marker.scale.z = scalevector.z();
+}
+
+static void use_cell_size(const auto& position, const auto& size, auto& marker){
```
ivanpauno, line 65:
```suggestion
inline void use_cell_size(const auto& position, const auto& size, auto& marker){
```

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,132 @@
...
+  beluga_ros::msg::MarkerArray message{};
+  {
+    beluga_ros::msg::Marker marker;
+    marker.ns = "obstacles";
+    marker.action = beluga_ros::msg::Marker::DELETEALL;
+    message.markers.push_back(marker);
+  }
+
+  // Add the markers
+  int idCount = 0;
```
ivanpauno, line 102: you can use https://github.com/Ekumen-OS/beluga/blob/08e682fe0795952ad01fc991979fe3056a27c29a/beluga/include/beluga/algorithm/distance_map.hpp#L72-L78

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
ivanpauno: it may be worth to move the definitions to a .cpp file though

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,132 @@
...
+  marker.pose.position.z = mean[2];
+
+  marker.pose.orientation.x = rotation.x();
+  marker.pose.orientation.y = rotation.y();
+  marker.pose.orientation.z = rotation.z();
+  marker.pose.orientation.w = rotation.w();
+
+  marker.scale.x = scalevector.x();
+  marker.scale.y = scalevector.y();
+  marker.scale.z = scalevector.z();
```
ivanpauno, line 62: @hidmic we were thinking to use *3 or *5 here, so it visualizes better

beluga_amcl/src/ndt_amcl_node_3d.cpp
```diff
@@ -245,8 +255,14 @@ beluga::NDTSensorModel<NDTMapRepresentation> NdtAmclNode3D::get_sensor_model() c
...
-  return beluga::NDTSensorModel<NDTMapRepresentation>{
-      params, beluga::io::load_from_hdf5<NDTMapRepresentation>(get_parameter("map_path").as_string())};
+  // Load the map from hdf5 file
+  const auto map = beluga::io::load_from_hdf5<NDTMapRepresentation>(get_parameter("map_path").as_string());
+
+  // Publish markers for map visualization
+  beluga_ros::msg::MarkerArray obstacle_markers = beluga_ros::assign_obstacle_map(map);
+  map_visualization_pub_->publish(obstacle_markers);
+
+  return beluga::NDTSensorModel<NDTMapRepresentation>{params, map};
```
hidmic, line 265: @Fernando-Sanz meta: I understand why this was added here but it is not a suitable place for publishing messages. Consider decoupling sensor model instantiation from map loading so you can publish these markers on node activate.

beluga_amcl/include/beluga_amcl/ndt_amcl_node_3d.hpp
```diff
@@ -145,6 +145,9 @@ class NdtAmclNode3D : public BaseAMCLNode {
   std::optional<Sophus::SE3d> last_known_odom_transform_in_map_;
   /// Whether to broadcast transforms or not.
   bool enable_tf_broadcast_{false};
+
+  /// Map representation publisher.
+  rclcpp::Publisher<visualization_msgs::msg::MarkerArray>::SharedPtr map_visualization_pub_;
```
hidmic, line 150: @Fernando-Sanz meta: we should be using `rclcpp::LifecyclePublisher`s for this

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
hidmic: Also, `assign` should take an output argument like other other `assign` functions.

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
hidmic: To be clear, @ivanpauno are you proposing we use `ranges::views::enumerate`? If so, I agree.

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+  // Add the markers
+  int idCount = 0;
+  for (const auto& [key, cell] : map) {
+    beluga_ros::msg::Marker marker;
+    marker.header.frame_id = "map";
+    marker.id = idCount++;
+    marker.ns = "obstacles";
+
+    const auto eigenSolver = Eigen::SelfAdjointEigenSolver<Eigen::Matrix<double, 3, 3>>{cell.covariance};
```
hidmic, line 109: @Fernando-Sanz consider instantiating the solver once, using `Eigen::SelfAdjointEigenSolver::compute` on each covariance matrix.

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,132 @@
...
+  marker.action = beluga_ros::msg::Marker::ADD;
+  
+  // Compute the rotation based on the covariance matrix
+  const auto eigenvectors = eigenSolver.eigenvectors().real();
+  const auto eigenvalues = eigenSolver.eigenvalues();
+
+  Eigen::Matrix3d rotationMatrix;
+  Eigen::Vector3d scalevector;
+  rotationMatrix << eigenvectors.col(0), eigenvectors.col(1), eigenvectors.col(2);
+  scalevector = Eigen::Vector3d{eigenvalues.x(), eigenvalues.y(), eigenvalues.z()};
```
hidmic, line 38: @Fernando-Sanz eigenvalues of a covariance matrix are variances, we are missing `std::sqrt` to get 1 std ellipsoids.

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+  // Permutation
+  if (std::abs(rotationMatrix.determinant() - 1.0) > Eigen::NumTraits<double>::dummy_precision()) {
+    rotationMatrix << eigenvectors.col(1), eigenvectors.col(0), eigenvectors.col(2);
+    scalevector = Eigen::Vector3d{eigenvalues.y(), eigenvalues.x(), eigenvalues.z()};
+  }
```
hidmic, line 44: @Fernando-Sanz why do we need to change the determinant sign by permuting the matrix?

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+  Eigen::Quaterniond rotation{1.0f, 0.0f, 0.0f, 0.0f};
+  if (std::abs(rotationMatrix.determinant() - 1.0) < Eigen::NumTraits<double>::dummy_precision()) {
+    rotation = Eigen::Quaterniond{rotationMatrix};
+  }
```
hidmic, line 49: @Fernando-Sanz meta: are there any instances of non-orthogonal eigenvectors? They should be for a symmetric covariance matrix. If it's a numerical issue, I'd suggest we orthogonalize rather than ignore.

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
hidmic: That makes sense.

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
Fernando-Sanz: Because we want only a rotation as transformation. I noticed that this way it produces more reasonable results (e.g. some ellipsoids get aligned with other in walls or the floor). Maybe that was a wrong behaviour, is it a bad idea?
ndt_map_visualization_no_permutation
> Obstacle representation with no permutation. Note that some of the ellipsoids in the floor and walls are not aligned in contrast with the same ellipsoids in the main description picture.

beluga_amcl/src/ndt_amcl_node_3d.cpp
ivanpauno: this is called from `make_particle_filter()`, which is called on `on_activate()`.
We  could first save the loaded sparse grid and then separately publish on `on_activate()` but it's not going to make any real difference

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
ivanpauno: @Fernando-Sanz you will need to modify the cmake in that case https://github.com/Ekumen-OS/beluga/blob/08e682fe0795952ad01fc991979fe3056a27c29a/beluga_ros/cmake/ament.cmake#L26-L27

beluga_amcl/include/beluga_amcl/ndt_amcl_node_3d.hpp
ivanpauno: IMHO it does not make any real difference.
A lifecycle publisher only provides an on/off, that will discard the message if you publish before activating.
It's not really useful.

As long as you create the publisher on configure and start publishing after activation, then it's fine.

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
ivanpauno: Yes, I missed that.
@Fernando-Sanz feel free to modify the multiplier to make the visualization okay, but you need to calculate the square root as Mich says.

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
ivanpauno: I think we can probably just ignore this case (?).
Or just warn if it happens, it shouldn't happen in practice.

Or do the same as when the eigenvector solver does not converge and draw a cell-size square

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -99,11 +100,11 @@ beluga_ros::msg::MarkerArray assign_obstacle_map(const beluga::SparseValueGrid<M
   }
 
   // Add the markers
-  int idCount = 0;
-  for (const auto& [key, cell] : map) {
+  for (auto [index, entry] : ranges::views::enumerate(map)) {
+    const auto [cellCenter, cell] = entry;
```
ivanpauno, line 106:
```suggestion
    const auto& [cellCenter, cell] = entry;
```

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
ivanpauno: @Fernando-Sanz this one is still likely a good idea, it will make the code easier to read

beluga_amcl/include/beluga_amcl/ndt_amcl_node_3d.hpp
hidmic: Still, consistency with the rest of the codebase is good.

beluga_amcl/src/ndt_amcl_node_3d.cpp
hidmic: +1 to separate publishing

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,146 @@
...
+#include <beluga_ros/messages.hpp>
+
+/**
+ * \file
+ * \brief Utilities for NDT I/O over ROS interfaces.
+ */
+
+namespace beluga_ros {
+
+template <typename RealScalar>
```
hidmic, line 27: @Fernando-Sanz let's put these in an anonymous namespace.

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+  // Permutation
+  if(!isApprox(rotationMatrix.determinant(), 1.0)) {
+    rotationMatrix << eigenvectors.col(1), eigenvectors.col(0), eigenvectors.col(2);
+    scalevector = Eigen::Vector3d{eigenvalues.y(), eigenvalues.x(), eigenvalues.z()};
+  }
+
+  if(!isApprox(rotationMatrix.determinant(), 1.0)) {
+    return false;
+  }
```
hidmic, line 55: @Fernando-Sanz nit: consider flipping the order and computing fewer times:
```suggestion
  if(!isApprox(std::abs(rotationMatrix.determinant()), 1.0)) {
    return false;
  }
  
  // Permute columns to ensure the rotation is in fact a rotation
  if (rotationMatrix.determinant() < 0) {
    rotationMatrix << eigenvectors.col(1), eigenvectors.col(0), eigenvectors.col(2);
    scalevector = Eigen::Vector3d{eigenvalues.y(), eigenvalues.x(), eigenvalues.z()};
  }
```

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+  float scaleConstant = 2.0f;
+  marker.scale.x = std::sqrt(scalevector.x()) * scaleConstant;
+  marker.scale.y = std::sqrt(scalevector.y()) * scaleConstant;
+  marker.scale.z = std::sqrt(scalevector.z()) * scaleConstant;
+
+  return true;
+}
+
+inline void use_cell_size(const auto& position, const auto& size, auto& marker){
```
hidmic, line 79: @Fernando-Sanz I sense this needs some formatting, are we using [`pre-commit` hooks](https://github.com/Ekumen-OS/beluga/blob/main/CONTRIBUTING.md#how-do-i-submit-a-good-code-contribution)?
```suggestion
inline void use_cell_size(const auto& position, const auto& size, auto& marker) {
```

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,146 @@
...
+    if(!use_mean_covariance(eigenSolver, cell, marker)){
+      // Create a cube based on the resolution of the grid
+      use_cell_size(cellCenter, grid.resolution(), marker);
+    }
+
+    float color[] = {0.0f, 1.0f, 0.0f, 1.0f};
+    marker.color.r = color[0];
+    marker.color.g = color[1];
+    marker.color.b = color[2];
+    marker.color.a = color[3];
```
hidmic, line 135: @Fernando-Sanz why not simply assign literals?

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,146 @@
...
+  marker.pose.position.x = cell.mean[0];
+  marker.pose.position.y = cell.mean[1];
+  marker.pose.position.z = cell.mean[2];
+
+  marker.pose.orientation.x = rotation.x();
+  marker.pose.orientation.y = rotation.y();
+  marker.pose.orientation.z = rotation.z();
+  marker.pose.orientation.w = rotation.w();
+
+  float scaleConstant = 2.0f;
```
hidmic, line 71: @Fernando-Sanz can be `constexpr`, also consider using hungarian notation e.g.:
```suggestion
  constexpr float kScaleConstant = 2.0f;
```

hidmic (commented):
Second pass

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
Fernando-Sanz: That way appear some cubes. A non-rotation matrix could have the determinant != -1 that change to 1 with the permutation, couldn't it?

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
Fernando-Sanz: It was a try to see the color code better. I'll change it to the previous way.

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
Fernando-Sanz: Yes(?). Now I'm not sure, I installed it and it was working with previous commits. I'm checking it.

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
ivanpauno: I think @hidmic intended the following:

```suggestion  
  // Permute columns to ensure the rotation is in fact a rotation
  if (rotationMatrix.determinant() < 0) {
    rotationMatrix << eigenvectors.col(1), eigenvectors.col(0), eigenvectors.col(2);
    scalevector = Eigen::Vector3d{eigenvalues.y(), eigenvalues.x(), eigenvalues.z()};
  }
  if(!isApprox(std::abs(rotationMatrix.determinant()), 1.0)) {
    return false;
  }
```

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
ivanpauno: Linter are happy now though https://github.com/Ekumen-OS/beluga/actions/runs/12319843458/job/34482145756?pr=453.

So I guess this isn't needed anymore

beluga_amcl/src/ndt_amcl_node_3d.cpp
```diff
@@ -152,6 +154,18 @@ NdtAmclNode3D::NdtAmclNode3D(const rclcpp::NodeOptions& options) : BaseAMCLNode{
 
 void NdtAmclNode3D::do_activate(const rclcpp_lifecycle::State&) {
   RCLCPP_INFO(get_logger(), "Making particle filter");
+
+  map_visualization_pub_->on_activate();
+
+  // Publish markers for map visualization
+  beluga_ros::msg::MarkerArray obstacle_markers{};
+  bool flag;
```
nahueespinosa, line 162: Please, use a less-generic name for this variable.

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+void use_cell_size(const Eigen::Vector<int, 3>& position, double size, beluga_ros::msg::Marker& marker);
+
+/// Assign an ellipsoid to each cell of a SparseValueGrid. A cube is used instead if the distribution of the
+/// cell is not suitable for the rotation matrix creation.
+/**
+ * \param grid A SparseValueGrid that contains cells representing obstacles.
+ * \return A message with the ellipsoids or cubes.
+ * \param[out] message Markers message to be assigned.
+ * \param[out] cubesGenerated Is set to true if there were problems with covariance matrices from cells.
```
nahueespinosa, line 48: style: Use snake_case for variables here and everywhere else.
```suggestion
 * \param[out] cubes_generated Is set to true if there were problems with covariance matrices from cells.
```

beluga_amcl/src/ndt_amcl_node_3d.cpp
```diff
@@ -152,6 +154,18 @@ NdtAmclNode3D::NdtAmclNode3D(const rclcpp::NodeOptions& options) : BaseAMCLNode{
...
+  beluga_ros::assign_obstacle_map(map_, obstacle_markers, flag);
+  if (flag) {
+    RCLCPP_WARN(get_logger(), "Covariances from map representation cells seem to be non-diagonalizable");
```
nahueespinosa, line 165: What happens in this case, are some of the markers not generated? Please explain that in the warning message.

beluga_ros/src/ndt_ellipsoid.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+#include <Eigen/Core>
+#include <beluga/eigen_compatibility.hpp>
+
+namespace beluga_ros {
+
+namespace {
+
+template <typename RealScalar>
+inline bool isApprox(RealScalar x, RealScalar y, RealScalar prec = Eigen::NumTraits<RealScalar>::dummy_precision()) {
```
nahueespinosa, line 25: style: Use snake_case for free functions here and everywhere else.

```suggestion
inline bool is_approx(RealScalar x, RealScalar y, RealScalar prec = Eigen::NumTraits<RealScalar>::dummy_precision()) {
```

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,93 @@
+// Copyright 2024 Ekumen, Inc.
```
nahueespinosa, line 1:
```suggestion
// Copyright 2025 Ekumen, Inc.
```

beluga_ros/src/ndt_ellipsoid.cpp
```diff
@@ -0,0 +1,105 @@
+// Copyright 2024 Ekumen, Inc.
```
nahueespinosa, line 1:
```suggestion
// Copyright 2025 Ekumen, Inc.
```

nahueespinosa (commented):
This is great! @Fernando-Sanz left some minor comments.

beluga_amcl/src/ndt_amcl_node_3d.cpp
ivanpauno: in that case, we create some cell-size cubes instead of generate an ellipsoid based on the covariance.
the color is different to indicate the error, and in should not happen in a well generated map (it was not a problem in the demo we have)

ivanpauno: +1 to the name of the variable.
It may be better to return a `std::pair<MarkerArray, bool>` as well.

nahueespinosa: Thanks for the context! Sounds reasonable.

Suggestion:

```cpp
RCLCPP_WARN(get_logger(), "Some cell covariances appear to be non-diagonalizable. A cube will be generated instead of an ellipsoid for those.");
```

beluga_ros/include/beluga_ros/ndt_ellipsoid.hpp
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+namespace beluga_ros {
+
+bool use_mean_covariance(
+    Eigen::SelfAdjointEigenSolver<Eigen::Matrix<double, 3, 3>>& eigen_solver,
+    beluga::NDTCell<3> cell,
+    beluga_ros::msg::Marker& marker);
+
+void use_cell_size(const Eigen::Vector<int, 3>& position, double size, beluga_ros::msg::Marker& marker);
```
nahueespinosa, line 40: This needs a docstring.

ivanpauno (approved):

ivanpauno:
@Fernando-Sanz you have linter failing

ivanpauno (approved):

ivanpauno:
We need to fix [this CI](https://github.com/Ekumen-OS/beluga/actions/runs/12929860016/job/36070351074?pr=453) job, independently of this PR

PR #458, opened 2025-01-23
Bump upload and download artifact actions to v4

nahueespinosa:
### Proposed changes

Version v3 has been deprecated.

Reference:
https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

#### Type of change

- [x] 🐛 Bugfix (change which fixes an issue)
- [ ] 🚀 Feature (change which adds functionality)
- [ ] 📚 Documentation (change which fixes or extends documentation)

### Checklist

- [x] Lint and unit tests (if any) pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] All commits have been signed for [DCO](https://developercertificate.org/)

PR #460, opened 2025-01-23
Fix artifact version

marcoshuck:
### Proposed changes

The CI job for uploading the docs is failing due to an outdated action version. This PR updates the version to the latest one, following the [migration guide](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md).

#### Type of change

### Checklist


.github/workflows/ci_pipeline.yml
```diff
@@ -169,9 +170,10 @@ jobs:
           git rm -rf .
 
       - name: Download documentation artifacts
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
         with:
           name: docs
+          merge-multiple: true
```
marcoshuck, line 176: I believe this is not actually needed.

.github/workflows/ci_pipeline.yml
```diff
@@ -142,10 +142,11 @@ jobs:
         run: make -C docs html
 
       - name: Upload documentation artifacts
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: docs
           path: ./docs/_build/html/
+          overwrite: true
```
marcoshuck, line 149: I believe this is not actually needed.

glpuga: From total ignorance, putting [this phrase](https://github.com/actions/upload-artifact?tab=readme-ov-file#not-uploading-to-the-same-artifact):

> Previously the behavior allowed for the artifact names to be the same which resulted in unexpected mutations and accidental corruption. Artifacts created by upload-artifact@v4 are immutable.

and [this one](https://github.com/actions/upload-artifact?tab=readme-ov-file#overwriting-an-artifact):

> Although it's not possible to mutate an Artifact, can completely overwrite one. But do note that this will give the Artifact a new ID, the previous one will no longer exist

sounds like it might be.

Or do you think that because the job is the same, overwrite is not needed?

.github/workflows/ci_pipeline.yml
glpuga: maybe the ide here is to avoid an extra dir in the intermediate path?

glpuga:
Just noticed this duplicates the almost simultaneous https://github.com/Ekumen-OS/beluga/pull/458 , I suggest we join discussions in a single PR @marcoshuck @nahueespinosa

.github/workflows/ci_pipeline.yml
marcoshuck: As far as I understood, we're generating a unique artifact called `docs` which might be overwritten by new PRs, however it's only intended to be overwritten in the `main` branch.

.github/workflows/ci_pipeline.yml
marcoshuck: That makes sense if you have multiple jobs generating an artifact, but since we have a single one, it's probably not required. See https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#multiple-uploads-to-the-same-named-artifact

.github/workflows/ci_pipeline.yml
glpuga: > As far as I understood, we're generating a unique artifact called `docs` which might be overwritten by new PRs

Indeed, I think the same, except for the part "might be overwritten by new PRs" since the v4 documentation says artifacts with the same id are now inmutable unless "overwrite=true".

My question here is whether "inmutable" applies only when different jobs create the same artifact (therefore a second job being unable to update the artifact, but the original job being able to update the artifact multiple times), or any job (therefore no job being able to update the artifact once created, unless overwrite is true).

.github/workflows/ci_pipeline.yml
glpuga: Indeed, I agree it may not be necessary, just what whoever wrote this preferred.

Since this might change paths and with a small probability have unintended effects in the deploy process later (haven't checked), but has no other advantages to fix otherwise, I vote for keeping as is (it-aint-broken policy).

.github/workflows/ci_pipeline.yml
marcoshuck: > Indeed, I think the same, except for the part "might be overwritten by new PRs" since the v4 documentation says artifacts with the same id are now inmutable unless "overwrite=true".

Exactly, I didn't specify it but I was referring to the fact of keeping the `overwrite = true`.

> My question here is whether "inmutable" applies only when different jobs create the same artifact (therefore a second job being unable to update the artifact, but the original job being able to update the artifact multiple times), or any job (therefore no job being able to update the artifact once created, unless overwrite is true).

I think we can check this by deploying this PR and fixing if it's not the case in a follow up, or someone could take a dive into the documentation to double check if it's the case. I'm a bit packed right now to do so, but I can try next week.

I'd suggest removing `overwrite` for now.

.github/workflows/ci_pipeline.yml
hidmic: +1 to keeping it. Eventually, we'll need to version documentation. Sooner or later we'll wind up having to collate generated documentation artifacts.

.github/workflows/ci_pipeline.yml
hidmic: Looking at https://github.com/actions/toolkit/blob/1f7c2c79e034fe8a0d28006f52fc5b70f6dbb750/packages/artifact/src/internal/upload/upload-artifact.ts#L24, my understanding is that, without `overwrite: true`, we won't be able to upload that `docs` artifact more than once. Artifacts IDs seem orthogonal to branches (and git refs in general) unless you couple them through the name.

hidmic: :thinking: we could, but we don't really have a use for older artifacts, and while we will eventually need versioned documentation it seems unlikely artifact names will play any role there.

marcoshuck: Correct me if I'm wrong: The problem that I see is that we're running this workflow for every PR, overwriting the docs on the `main` branch with a PR that hasn't been merged is not a ideal.

hidmic: Ah, good point. Docs will not get deployed for PRs, so that's not a risk, but we potentially lose the ability to inspect the docs generated for a PR if two or more are running CI concurrently.

Nevermind what I said, +1 to adding a git ref to the name.

hidmic (approved):
LGTM with green CI

hidmic:
`codecov` went nuts.

marcoshuck:
> `codecov` went nuts.

what do we do?

hidmic:
> what do we do?

We ignore it :smiley:

Issue #468, opened 2025-02-04
When using large maps, the likelihood map constructions takes very long time


glpuga:
### Bug description

Upon map reception, building the likelihood map when a new map is received takes a long time if the map is large (ej. 200mx600m). This is because of the construction of the distance map and then likelihood map in [this function](https://github.com/Ekumen-OS/beluga/blob/b6d5ee387cd196aa3c11617c993aa6300c7a9484/beluga/include/beluga/sensor/likelihood_field_model.hpp#L158).

On a Intel Core i9-9900 CPU this is around 22 seconds for a 200mx600m at 0.05 m/pixel resolution.

![Image](https://github.com/user-attachments/assets/5dce7c6b-1a65-4a53-b61c-6f57e8c191c0)

What's worse, during this time the node is unresponsive. 
There's a related failure mode that is triggered by this, see https://github.com/Ekumen-OS/beluga/issues/469 .

**Platform (please complete the following information):**
 - OS: Any.
 - `Beluga` version: 2.0.2

### How to reproduce

Run any system using beluga and a large map. 200mx600m at 0.05 m/pixel resolution is large enough.

**Expected behavior**

The node should not go unresponsive for so long.

**Actual behavior**

Node freezes, lifecycle node complains and actually fails.

glpuga:
Tried different variations of map arrangement to see how they impact execution time.

- Map concentrated in a single place, unknown space elsewhere (the original map I used for a quick test, but not a realistic scenario for actual users).
- Same as before, but masking all unused space with an "obstacle" mask, so that the fill algorithm only steps there once.
- Map tiled, something I've seen when multiple indoor spaces get crammed into a single ROS map to avoid switching


![Image](https://github.com/user-attachments/assets/07b23416-c1ef-4583-809d-6c2af3ba665f)
![Image](https://github.com/user-attachments/assets/afa1eafe-8cc9-4d9f-b521-094d80ebb98f)
![Image](https://github.com/user-attachments/assets/d7b165e0-9d7f-454f-8242-6300c2e0bfcc)

The original case is still the fastest at 22 seconds; the other two have even longer processing times. This was a bit surprising for the tiled map at first, but after thinking about it, this kind of map many more elements in the priority queue at any given time, so it makes sense.

Without totally changing the algorithm, the best fix to this is not to calculate the distance map over "unknown" regions, which we currently do (and if "model unknown spaces" is enabled, we later discard and overwrite). I'll try that next.

PR #471, opened 2025-02-11
Fix to likelihood map pre-calculation performance for large maps

### LINKED ISSUE #472: When there are no obstacles in the map, make_likelihood_field() turns the map into one large block of obstacles
### Bug description

If there are no obstacles in the map, the `make_likelihood_field()` function in the Likelihood turns the likelihood map into a solid block of obstacles where all pixels are at distance 0 from an obstacle.

Instead of that, all cells should be assigned the minimum possible likelihood value, corresponding to the `laser_likelihood_max_dist` parameter.

We did not really pay attention to this limit case, and cristalized this bug into [a unit test that checks just this](https://github.com/Ekumen-OS/beluga/blob/main/beluga/test/beluga/algorithm/test_distance_map.cpp#L51-L55).

Later it also became part of [this other unit test](https://github.com/Ekumen-OS/beluga/blob/main/beluga/test/beluga/sensor/test_lfm_with_unknown_space.cpp#L304-L307), which also checks a totally unexpected value given that there no obstacles in the map.

**Platform (please complete the following information):**
 - OS: [e.g. Ubuntu Focal]
 - `Beluga` version: [e.g. tag, commit sha]

### How to reproduce

Check that the unit tests listed above currently pass in the code, despite the values being checked being incorrect.

**Expected behavior**

If there are no obstacles in the map, all cells should be assigned the minimum possible likelihood value, corresponding to the `laser_likelihood_max_dist` parameter.

**Actual behavior**

They are not.

### Additional context

Discovered while fixing #469

glpuga:
### Proposed changes

Addresses #468 
Also fixes #472 

#### Type of change

### Checklist

### Additional comments

### DIFF
--- a/beluga/include/beluga/algorithm/distance_map.hpp
+++ b/beluga/include/beluga/algorithm/distance_map.hpp
@@ -48,20 +48,22 @@ namespace beluga {
  * \param neighbors_function Given the index i of one cell in the map,
  *  neighbors_function(i) returns the cell indexes of neighbor cells in the
  *  obstacle map.
+ * \param max_distance_value The maximum distance anywhere in the map. Values over this will be truncated to this value.
  * \return A map where each cell value is the distance to the nearest object.
  */
 template <class Range, class DistanceFunction, class NeighborsFunction>
 auto nearest_obstacle_distance_map(
     Range&& obstacle_mask,
     DistanceFunction&& distance_function,
-    NeighborsFunction&& neighbors_function) {
+    NeighborsFunction&& neighbors_function,
+    std::invoke_result_t<DistanceFunction, std::size_t, std::size_t> max_distance_value) {
   struct IndexPair {
     std::size_t nearest_obstacle_index;
     std::size_t index;
   };
 
   using DistanceType = std::invoke_result_t<DistanceFunction, std::size_t, std::size_t>;
-  auto distance_map = std::vector<DistanceType>(ranges::size(obstacle_mask));
+  auto distance_map = std::vector<DistanceType>(ranges::size(obstacle_mask), max_distance_value);
   auto visited = std::vector<bool>(ranges::size(obstacle_mask), false);
 
   auto compare = [&distance_map](const IndexPair& first, const IndexPair& second) {
@@ -83,8 +85,11 @@ auto nearest_obstacle_distance_map(
     for (const std::size_t index : neighbors_function(parent.index)) {
       if (!visited[index]) {
         visited[index] = true;
-        distance_map[index] = distance_function(parent.nearest_obstacle_index, index);
-        queue.push(IndexPair{parent.nearest_obstacle_index, index});
+        const auto distance = distance_function(parent.nearest_obstacle_index, index);
+        if (distance < max_distance_value) {
+          distance_map[index] = distance;
+          queue.push(IndexPair{parent.nearest_obstacle_index, index});
+        }
       }
     }
   }
--- a/beluga/include/beluga/sensor/likelihood_field_model.hpp
+++ b/beluga/include/beluga/sensor/likelihood_field_model.hpp
@@ -160,11 +160,6 @@ class LikelihoodFieldModel {
       return static_cast<float>((grid.coordinates_at(first) - grid.coordinates_at(second)).squaredNorm());
     };
 
-    const auto squared_max_distance = static_cast<float>(params.max_obstacle_distance * params.max_obstacle_distance);
-    const auto truncate_to_max_distance = [squared_max_distance](auto squared_distance) {
-      return std::min(squared_distance, squared_max_distance);
-    };
-
     /// Pre-computed variables
     const double two_squared_sigma = 2 * params.sigma_hit * params.sigma_hit;
     assert(two_squared_sigma > 0.0);
@@ -180,19 +175,23 @@ class LikelihoodFieldModel {
 
     const auto neighborhood = [&grid](std::size_t index) { return grid.neighborhood4(index); };
 
+    const auto squared_max_distance = static_cast<float>(params.max_obstacle_distance * params.max_obstacle_distance);
+
     // determine distances to obstacles and calculate likelihood values in-place
     // to minimize memory usage when dealing with large maps
-    auto distance_map = nearest_obstacle_distance_map(grid.obstacle_mask(), squared_distance, neighborhood);
+    auto distance_map =
+        nearest_obstacle_distance_map(grid.obstacle_mask(), squared_distance, neighborhood, squared_max_distance);
 
     if (params.model_unknown_space) {
-      const double inverse_max_distance = 1 / params.max_laser_distance;
-      const double background_distance = -two_squared_sigma * std::log((inverse_max_distance - offset) / amplitude);
+      const auto inverse_max_distance = 1 / params.max_laser_distance;
+      const auto squared_background_distance =
+          -two_squared_sigma * std::log((inverse_max_distance - offset) / amplitude);
 
-      distance_map |= beluga::actions::overlay(grid.unknown_mask(), background_distance);
+      distance_map |= beluga::actions::overlay(
+          grid.unknown_mask(), std::min(squared_max_distance, static_cast<float>(squared_background_distance)));
     }
 
-    auto likelihood_values = std::move(distance_map) |                               //
-                             ranges::actions::transform(truncate_to_max_distance) |  //
+    auto likelihood_values = std::move(distance_map) |  //
                              ranges::actions::transform(to_likelihood);
 
     return ValueGrid2<float>{std::move(likelihood_values), grid.width(), grid.resolution()};
--- a/beluga/test/beluga/algorithm/test_distance_map.cpp
+++ b/beluga/test/beluga/algorithm/test_distance_map.cpp
@@ -44,38 +44,44 @@ auto make_neighbors_function(std::size_t size) {
 
 TEST(DistanceMap, None) {
   auto map = std::vector<bool>{};
-  auto distance_map = beluga::nearest_obstacle_distance_map(map, array_distance, make_neighbors_function(6));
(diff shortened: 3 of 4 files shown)


beluga/test/beluga/sensor/test_lfm_with_unknown_space.cpp
```diff
@@ -276,7 +276,7 @@ TEST(LikelihoodFieldModelUnknownSpace, GridWithRotationAndOffset) {
   }
 }
 
-TEST(LikelihoodFieldModelUnknownSpace, GridUpdates) {
+TEST(LikelihoodFieldModelUnknownSpace, DISABLED_GridUpdates) {
```
glpuga, line 279: I still need to figure why this test fails now.

beluga/include/beluga/algorithm/distance_map.hpp
```diff
@@ -54,19 +55,22 @@ template <class Range, class DistanceFunction, class NeighborsFunction>
...
+    std::invoke_result_t<DistanceFunction, std::size_t, std::size_t> max_distance_value) {
   struct IndexPair {
     std::size_t nearest_obstacle_index;
     std::size_t index;
   };
 
   using DistanceType = std::invoke_result_t<DistanceFunction, std::size_t, std::size_t>;
-  auto distance_map = std::vector<DistanceType>(ranges::size(obstacle_mask));
+
+  auto distance_map = std::vector<DistanceType>(ranges::size(obstacle_mask), max_distance_value);
```
glpuga, line 67:
```suggestion
  auto distance_map = std::vector<DistanceType>(ranges::size(obstacle_mask), max_distance_value);
```

beluga/include/beluga/algorithm/distance_map.hpp
```diff
@@ -54,19 +55,21 @@ template <class Range, class DistanceFunction, class NeighborsFunction>
...
 
   using DistanceType = std::invoke_result_t<DistanceFunction, std::size_t, std::size_t>;
-  auto distance_map = std::vector<DistanceType>(ranges::size(obstacle_mask));
+  auto distance_map = std::vector<DistanceType>(ranges::size(obstacle_mask), max_distance_value);
   auto visited = std::vector<bool>(ranges::size(obstacle_mask), false);
 
   auto compare = [&distance_map](const IndexPair& first, const IndexPair& second) {
     return distance_map[first.index] > distance_map[second.index];
   };
+
```
glpuga, line 72:
```suggestion
```

beluga/include/beluga/algorithm/distance_map.hpp
```diff
@@ -15,6 +15,7 @@
 #ifndef BELUGA_ALGORITHM_DISTANCE_MAP_HPP
 #define BELUGA_ALGORITHM_DISTANCE_MAP_HPP
 
+#include <algorithm>
```
glpuga, line 18:
```suggestion
```

beluga/test/beluga/sensor/test_lfm_with_unknown_space.cpp
glpuga: Fixed. The reason the test began to fail was because the behavior tested was actaully wrong. See #472 .

beluga/include/beluga/sensor/data/value_grid.hpp
```diff
@@ -15,6 +15,7 @@
 #ifndef BELUGA_SENSOR_DATA_VALUE_GRID_HPP
 #define BELUGA_SENSOR_DATA_VALUE_GRID_HPP
 
+#include <cmath>
```
glpuga, line 18:
```suggestion
```

beluga/include/beluga/sensor/likelihood_field_model.hpp
```diff
@@ -180,19 +175,22 @@ class LikelihoodFieldModel {
...
     // to minimize memory usage when dealing with large maps
-    auto distance_map = nearest_obstacle_distance_map(grid.obstacle_mask(), squared_distance, neighborhood);
+    auto distance_map = nearest_obstacle_distance_map(
+        grid.obstacle_mask(), squared_distance, neighborhood, static_cast<float>(squared_max_distance));
 
     if (params.model_unknown_space) {
-      const double inverse_max_distance = 1 / params.max_laser_distance;
-      const double background_distance = -two_squared_sigma * std::log((inverse_max_distance - offset) / amplitude);
+      const auto inverse_max_distance = 1 / params.max_laser_distance;
+      const auto background_distance = -two_squared_sigma * std::log((inverse_max_distance - offset) / amplitude);
```
glpuga, line 187: @hidmic @DPR00 this should be a squared distance, is it?

DPR00: I don't think so. The background distance is that function according to the formula. However, I think in the next line should be like this:
```
distance_map |= beluga::actions::overlay(
          grid.unknown_mask(), std::min(params.max_obstacle_distance, static_cast<float>(background_distance)));
```
The overlay action will write the distance value into the distance_map, so none of the values shouldn't be squared. Sorry, I missed this.

glpuga: Nop. The somewhat misnamed distance_map does not store linear distances, stored distances squared to avoid calculating the square root only to raise to the square again. The to_likehood() function expects distances to be actually squared distances, that's why this `background_distance` should actually be a distance squared too.

DPR00: Then, you're right it should be squared.

glpuga: Now, I see that the dimensionality in `background_distance` is given by `two_squared_sigma`, which by the name seems to be a distance squared, so maybe only the name `background_distance` is wrong, and not the value.

My question is, according to the model you guys followed here, is the value calculated there a squared distance?

DPR00: So, the background distance is obtained by clearing `dist` from the equation on line 8:
image
Therefore, the implemented formula here indeed corresponds to a squared distance. Likewise, the name `background_distance` is wrong. It should be something like `squared_background_distance`.

glpuga: Awesome, I can correct that. Thanks @DPR00 !

DPR00: thanks for noticing that!

beluga/include/beluga/algorithm/distance_map.hpp
```diff
@@ -48,20 +48,22 @@ namespace beluga {
...
+ * \param max_distance_value The maximum distance anywhere in the map. Values over this will be truncated to this value.
  * \return A map where each cell value is the distance to the nearest object.
  */
 template <class Range, class DistanceFunction, class NeighborsFunction>
 auto nearest_obstacle_distance_map(
     Range&& obstacle_mask,
     DistanceFunction&& distance_function,
-    NeighborsFunction&& neighbors_function) {
+    NeighborsFunction&& neighbors_function,
+    std::invoke_result_t<DistanceFunction, std::size_t, std::size_t> max_distance_value) {
```
hidmic, line 59: @glpuga nit^2: consider making `DistanceType` below a template parameter with default, then use it here.

beluga/include/beluga/sensor/likelihood_field_model.hpp
```diff
@@ -180,19 +175,23 @@ class LikelihoodFieldModel {
 
     const auto neighborhood = [&grid](std::size_t index) { return grid.neighborhood4(index); };
 
+    const auto squared_max_distance = static_cast<float>(params.max_obstacle_distance * params.max_obstacle_distance);
+
     // determine distances to obstacles and calculate likelihood values in-place
     // to minimize memory usage when dealing with large maps
-    auto distance_map = nearest_obstacle_distance_map(grid.obstacle_mask(), squared_distance, neighborhood);
+    auto distance_map = nearest_obstacle_distance_map(
+        grid.obstacle_mask(), squared_distance, neighborhood, static_cast<float>(squared_max_distance));
```
hidmic, line 183: @glpuga I think we can spare this `static_cast`, we already cast it to `float` above.

beluga/test/beluga/algorithm/test_distance_map.cpp
```diff
@@ -44,38 +44,44 @@ auto make_neighbors_function(std::size_t size) {
...
 }
 
 TEST(DistanceMap, Case3) {
   auto map = std::array<bool, 6>{false, false, false, false, false, true};
-  auto distance_map = beluga::nearest_obstacle_distance_map(map, array_distance, make_neighbors_function(6));
+  auto distance_map = beluga::nearest_obstacle_distance_map(map, array_distance, make_neighbors_function(6), 10);
   ASSERT_THAT(distance_map, testing::ElementsAre(5, 4, 3, 2, 1, 0));
 }
 
+TEST(DistanceMap, MinDistaceSet) {
```
hidmic, line 81: @glpuga nit: 
```suggestion
TEST(DistanceMap, MinDistanceSet) {
```

hidmic (approved):
Awesome!

beluga/test/beluga/algorithm/test_distance_map.cpp
glpuga: Fixed.

beluga/include/beluga/sensor/likelihood_field_model.hpp
glpuga: True, fixed.

beluga/include/beluga/algorithm/distance_map.hpp
glpuga: I could not get it to build. If you can provide a version of this that builds, I'll happily take it.

hidmic: Hmm, it's inferring the template parameter from a literal and breaking down. Nevermind.

Issue #469, opened 2025-02-04
When using large maps, the lifecycle node kills amcl_node because of missing heartbeat


glpuga:
### Bug description

Related to #468 , which discusses the _beluga_amcl_ node becoming unresponsive for long periods of time when receiving large maps.

This issue is about a complementary issue that seems to be triggered by that behavior, which is that because of the node becoming unresponsive, the Nav2 Lifecycle node loses track of the connection to the node and kills the node after `bond_timeout` seconds (parameter to the lifecycle node), and proceeds to restart the beluga amcl node.

Notice that this happens regardless of the value of `bond_timeout`. Setting this to a very large value (e.g. 100 seconds) and providing the node with a 600mx200m@0.05p/m  map (which takes 22 seconds to initialize), the following sequence will happen.

- The lifecycle node and the amcl node get initialized.
- the lifecycle node and the amcl node connect to eacht other with bond.
- amcl receives the map, and becomes unresponsive for approx. 22 seconds.
- amcl completes the initialization after processing the likelihood map and becomes **fully operational**.
- after `bond_timeout` seconds after the creation of the bond, lifecycle node complains about 
"[lifecycle_manager]: CRITICAL FAILURE: SERVER amcl IS DOWN after not receiving a heartbeat for NNNN ms"
- **The lifecycle node restarts the amcl node.**

![Image](https://github.com/user-attachments/assets/3767d358-fd6c-4e9c-8483-6c389046a4e7)

The fact that the lifecycle node does not detect the amcl node returning back to operational status may not be an issue within _beluga_amcl_, and might be in the lifecycle node itself, but _beluga_amcl_ certainly triggers this failure. The same failure mode does not happen if everything else is the same but _nav2_amcl_ is used instead of _beluga_amcl_.

**Platform (please complete the following information):**
 - OS: Seen in ROS Humble.
 - `Beluga` version: 2.0.2

### How to reproduce

Run any system using beluga and a large map. 200mx600m at 0.05 m/pixel resolution is large enough.

**Expected behavior**

The node should not go unresponsive for so long.

**Actual behavior**

Node freezes, lifecycle node complains and actually fails.

glpuga:
I'm still digging into this issue.

glpuga:
The issue is in Beluga, not in the [Lifecle manager](https://github.com/ros-navigation/navigation2/tree/main/nav2_lifecycle_manager) nor the [bond_core](https://github.com/ros/bond_core) library.

[...]

PR #473, opened 2025-02-13
Add ros parameter for bond timeout

glpuga:
### Proposed changes

Addresses #469 

#### Type of change

### Checklist

### Additional comments

### DIFF
--- a/beluga_amcl/src/ros2_common.cpp
+++ b/beluga_amcl/src/ros2_common.cpp
@@ -385,6 +385,12 @@ BaseAMCLNode::BaseAMCLNode(
     this->declare_parameter("autostart_delay", 0.0, descriptor);
   }
 
+  {
+    auto descriptor = rcl_interfaces::msg::ParameterDescriptor();
+    descriptor.description = "Heartbeat timeout value for the bond connection.";
+    this->declare_parameter("bond_timeout", 4.0, descriptor);
+  }
+
   common_callback_group_ = create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive);
   common_subscription_options_ = rclcpp::SubscriptionOptions{};
   common_subscription_options_.callback_group = common_callback_group_;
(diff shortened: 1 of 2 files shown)


beluga_amcl/src/ros2_common.cpp
```diff
@@ -451,25 +457,45 @@ rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn BaseAM
...
+    };
+    auto on_bond_broken_callback = [this]() {
+      RCLCPP_ERROR(get_logger(), "The bond connection to the lifecycle manager has been broken");
+    };
+    bond_ = std::make_unique<bond::Bond>(
+        "bond", get_name(), shared_from_this(), on_bond_broken_callback, on_bond_formed_callback);
     bond_->setHeartbeatPeriod(0.10);
-    bond_->setHeartbeatTimeout(4.0);
+    const auto heartbeat_timeout_value = get_parameter("bond_timeout").as_double();
+    // we don't want to shorten the default connection timeout, but we want to be able to
```
hidmic, line 485: @glpuga nit^2: 
```suggestion
    // We don't want to shorten the default connection timeout, but we want to be able to
```

hidmic (approved):
LGTM, though I think we should update https://github.com/Ekumen-OS/beluga/blob/main/beluga_amcl/docs/ros2-reference.md

hidmic:
I wonder though, did https://github.com/Ekumen-OS/beluga/pull/471 make this patch somewhat irrelevant to https://github.com/Ekumen-OS/beluga/issues/469? Still a nice feature to have.

glpuga:
> I wonder though, did #471 make this patch somewhat irrelevant to #469? Still a nice feature to have.

It did not, maps with a large obstacle-surface-to-map-area ratio still take a long time to build and might trip the heartbeat. In fact, this would do well to be fixed in nav2 too, because nav2 suffers from issues there too.

See https://github.com/Ekumen-OS/beluga/issues/469#issuecomment-2657711635

PR #476, opened 2025-02-26
Add likelihood prob sensor model and enable Beluga's flavour of clustering

fbattocchia:
### Proposed changes

The `LikelihoodFieldProb` sensor model is incorporated, based on the implementation presented in the book Probabilistic Robotics. To achieve this, the `LikelihoodFieldModelCommon` class is created, which implements the Likelihood Field shared by both the `LikelihoodFieldProb` and `LikelihoodField` models.

The `LikelihoodFieldProb` model, in addition to following the book's implementation, includes the summation of the logarithm of probabilities to prevent underflow when multiplying numerous small probabilities. Additionally, the Cluster-Based Estimation algorithm is integrated, enhancing the model's robustness against unmapped objects.

Furthermore, unit tests are implemented for the `LikelihoodFieldProb` sensor model.

#### Type of change

### Checklist

### DIFF
--- a/beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
+++ b/beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef BELUGA_SENSOR_LIKELIHOOD_FIELD_MODEL_HPP
-#define BELUGA_SENSOR_LIKELIHOOD_FIELD_MODEL_HPP
+#ifndef BELUGA_SENSOR_LIKELIHOOD_FIELD_PROB_MODEL_HPP
+#define BELUGA_SENSOR_LIKELIHOOD_FIELD_PROB_MODEL_HPP
 
 #include <algorithm>
 #include <beluga/sensor/likelihood_field_model_base.hpp>
@@ -23,23 +23,23 @@
 
 /**
  * \file
- * \brief Implementation of a likelihood field sensor model for range finders.
+ * \brief Implementation of a likelihood field prob sensor model for range finders.
  */
 
 namespace beluga {
 
-/// Parameters used to construct a LikelihoodFieldModel instance.
+/// Parameters used to construct a LikelihoodFieldProbModelParam instance.
 /**
  * See Probabilistic Robotics \cite thrun2005probabilistic Chapter 6.4, particularly Table 6.3.
  */
-using LikelihoodFieldModelParam = LikelihoodFieldModelBaseParam;
+using LikelihoodFieldProbModelParam = LikelihoodFieldModelBaseParam;
 
-/// Likelihood field sensor model for range finders.
+/// Likelihood field prob sensor model for range finders.
 /**
  * @copydoc LikelihoodFieldModelBase
  */
 template <class OccupancyGrid>
-class LikelihoodFieldModel : public LikelihoodFieldModelBase<OccupancyGrid> {
+class LikelihoodFieldProbModel : public LikelihoodFieldModelBase<OccupancyGrid> {
  public:
   /// State type of a particle.
   using state_type = Sophus::SE2d;
@@ -50,13 +50,13 @@ class LikelihoodFieldModel : public LikelihoodFieldModelBase<OccupancyGrid> {
   /// Map representation type.
   using map_type = OccupancyGrid;
   /// Parameter type that the constructor uses to configure the likelihood field model.
-  using param_type = LikelihoodFieldModelParam;
+  using param_type = LikelihoodFieldProbModelParam;
 
-  /// Constructs a LikelihoodFieldModel instance.
+  /// Constructs a LikelihoodFieldProbModel instance.
   /**
    * @copydoc LikelihoodFieldModelBase::LikelihoodFieldModelBase
    */
-  explicit LikelihoodFieldModel(const param_type& params, const map_type& grid)
+  explicit LikelihoodFieldProbModel(const param_type& params, const map_type& grid)
       : LikelihoodFieldModelBase<OccupancyGrid>(params, grid) {}
 
   /// Returns a state weighting function conditioned on 2D lidar hits.
@@ -73,8 +73,9 @@ class LikelihoodFieldModel : public LikelihoodFieldModelBase<OccupancyGrid> {
       const auto cos_theta = transform.so2().unit_complex().x();
       const auto sin_theta = transform.so2().unit_complex().y();
       const auto unknown_space_occupancy_prob = static_cast<float>(1. / this->params_.max_laser_distance);
-      return std::transform_reduce(
-          points.cbegin(), points.cend(), 1.0, std::plus{},
+
+      return std::exp(std::transform_reduce(
+          points.cbegin(), points.cend(), 0.0, std::plus{},
           [this, x_offset, y_offset, cos_theta, sin_theta, unknown_space_occupancy_prob](const auto& point) {
             // Transform the end point of the laser to the grid local coordinate system.
             // Not using Eigen/Sophus because they make the routine x10 slower.
@@ -83,10 +84,8 @@ class LikelihoodFieldModel : public LikelihoodFieldModelBase<OccupancyGrid> {
             const auto y = point.first * sin_theta + point.second * cos_theta + y_offset;
             const auto pz =
                 static_cast<double>(this->likelihood_field_.data_near(x, y).value_or(unknown_space_occupancy_prob));
-            // TODO(glpuga): Investigate why AMCL and QuickMCL both use this formula for the weight.
-            // See https://github.com/Ekumen-OS/beluga/issues/153
-            return pz * pz * pz;
-          });
+            return std::log(pz);
+          }));
     };
   }
 };
--- a/beluga_ros/src/amcl.cpp
+++ b/beluga_ros/src/amcl.cpp
@@ -18,9 +18,10 @@
 #include <beluga/actions/normalize.hpp>
 #include <beluga/actions/propagate.hpp>
 #include <beluga/actions/reweight.hpp>
-#include <beluga/algorithm/estimation.hpp>
+#include <beluga/algorithm/cluster_based_estimation.hpp>
 #include <beluga/views/random_intersperse.hpp>
 #include <beluga/views/take_while_kld.hpp>
+#include <cmath>
 
 namespace beluga_ros {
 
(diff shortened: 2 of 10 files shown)


glpuga:
This is currently being evaluate by running https://github.com/Ekumen-OS/lambkin/pull/123 on the whole set of datasets.

glpuga:
After a lot of cajoling with storage and process issues, I managed to get a sensible set of benchmarks using Lambkin. 

I'm glad to report the new model comes without surprises and performs as expected:
- Equal or better error performance than the non-prob model on average.
- Increased sensibility shown in datases including unmapped obstacles, or open-ended maps (equivalent to unmapped obstacles in practice).

A few bagfiles seem to have had issues, but looksl like one-off cases where multiple sensor model experiments failed at the same time.

Don't take my word for it, look for yourselves: [report.pdf](https://github.com/user-attachments/files/19375155/report.pdf)

@fbattocchia if you current schedule allows it, please join the RnD sync to discuss with @hidmic and @nahueespinosa how to best move forward with this PR.

hidmic:
> A few bagfiles seem to have had issues, but looks like one-off cases where multiple sensor models failed at the same time.

Specifically, results for sections 4.20, 4.43, 4.54, 8.3, 9.1, 9.8, 9.19 show this variant diverging. In all these cases, Beluga LF stood its ground. In some cases, Beluga LFP was the only one that went rogue. 

I still think this constitutes an improvement to the project, but I'm not entirely convinced we should make it the default.

beluga/include/beluga/sensor/likelihood_field_model_common.hpp
```diff
@@ -0,0 +1,168 @@
...
+ * Probabilistic Robotics \cite thrun2005probabilistic, Chapter 6.4,
+ * for further reference.
+ *
+ * \note This class satisfies \ref SensorModelPage.
+ *
+ * \tparam OccupancyGrid Type representing an occupancy grid.
+ *  It must satisfy \ref OccupancyGrid2Page.
+ */
+template <class OccupancyGrid>
+class LikelihoodFieldModelCommon {
```
hidmic, line 77: @fbattocchia nit^N!: consider `LikelihoodFieldModelBase` for a name.

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
```diff
@@ -0,0 +1,109 @@
...
+          points.cbegin(), points.cend(), 0.0, std::plus{},
+          [this, x_offset, y_offset, cos_theta, sin_theta, unknown_space_occupancy_prob](const auto& point) {
+            // Transform the end point of the laser to the grid local coordinate system.
+            // Not using Eigen/Sophus because they make the routine x10 slower.
+            // See `benchmark_likelihood_field_model.cpp` for reference.
+            const auto x = point.first * cos_theta - point.second * sin_theta + x_offset;
+            const auto y = point.first * sin_theta + point.second * cos_theta + y_offset;
+            const auto pz =
+                static_cast<double>(this->likelihood_field_.data_near(x, y).value_or(unknown_space_occupancy_prob));
+            return std::log(pz);
```
hidmic, line 101: @fbattocchia meta: I really wonder what happens with the result of `std::log(pz)`, numerically speaking. Does it shoot to ~0?

beluga/include/beluga/sensor/likelihood_field_model_common.hpp
```diff
@@ -0,0 +1,168 @@
...
+   * Used to calculate the probability of the obstacle being hit.
+   */
+  double sigma_hit = 0.2;
+  /// Whether to model unknown space or assume it free.
+  bool model_unknown_space = false;
+};
+
+/// Likelihood field common sensor model for range finders.
+/**
+ * This model relies on a pre-computed likelihood map of the environment.
```
hidmic, line 64: @fbattocchia if we move `LikelihoodFieldModel` documentation here, we should deduplicate it there. See the `@copydoc`  command in Doxygen.

hidmic (commented):
First pass. Great work @fbattocchia !

glpuga:
> Specifically, results for sections 4.20, 4.43, 4.54, 8.3, 9.1, 9.8, 9.19 show this variant diverging. In all these cases, Beluga LF stood its ground. In some cases, Beluga LFP was the only one that went rogue.

To better qualify this:

- 4.20 Almost all MCL variants drop sideways, this is probably a problem during the execution of the benchmark.
- 4.43 Nav2 LFP, implementing the same algorithm, fails too.
- 4.54 All MCLs fail, probably a problem running the benchmark.
- 8.3 This is the OpenLoris dataset, which has characteristics known to make the LFP variants unstable (heavy in visible unmapped regions with lidar hits).

Regarding 9.1, 9.8 and 9.19, Beluga LFP is indeed the only one failing, but this is still within the expected distribution for that dataset. Keep in mind that this is the hardest dataset for the LFP variants without beam skipping (both Beluga and Nav2), because the environment is heavily populated by unmapped objects in close proximity to the robot.

Indeed, in the same dataset, Nav2 LFP (without beam skipping) fails quite often too:

- 9.16 Nav2 LFP fails alone
- 9.17 Nav2 LFP fails alone
- 9.18 Both Beluga LFP and Nav2 LFP fail
- 9.28 Nav2 LFP fails alone
- 9.34 Nav2 LFP fails alone
- 9.41 Nav2 LFP fails alone

glpuga:
> Specifically, results for sections 4.20, 4.43, 4.54, 8.3, 9.1, 9.8, 9.19 show this variant diverging. In all these cases, Beluga LF stood its ground. In some cases, Beluga LFP was the only one that went rogue.

Notice also that the benchmark for performance for Beluga LFP is not Beluga LF. Experimentally we know the LFP without beam skipping variants to be less robust (but more accurate as well) than the equivalent LF variants in presence of unmapped obstacles.

Beluga LFP aims to match Nav2 LFP without beam skipping, so thats the benchmark for the model being added here.

Later work will enhance this by adding beam skipping or some other other feature to make it more robust in the presence of unmapped obstacles and hits in unmapped areas.

hidmic:
> Notice also that the benchmark for performance for Beluga LFP is not Beluga LF. Experimentally we know the LFP without beam skipping variants to be less robust (but more accurate as well) than the equivalent LF variants in presence of unmapped obstacles.
> 
> Beluga LFP aims to match Nav2 LFP without beam skipping, so thats the benchmark for the model being added here.
> 
> Later work will enhance this by adding beam skipping or some other other feature to make it more robust in the presence of unmapped obstacles and hits in unmapped areas.

And I agree we should make it the default when that work lands and it shows it's better than Beluga LF all around. I just wouldn't do that just yet. It can be an optional for now.

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
fbattocchia: we tested and the results do not change by adding `std::log(pz)`, we left it as a precaution and because it seems to be a common practice

hidmic: Yeah, that I understand. I'm just wondering about how it affects the numerical properties of the algorithm, nevermind results.

fbattocchia:
@hidmic I have already implemented the changes, it is ready for review.

beluga_ros/src/amcl.cpp
```diff
@@ -97,7 +98,7 @@ auto Amcl::update(Sophus::SE2d base_pose_in_odom, beluga_ros::LaserScan laser_sc
   }
 
   force_update_ = false;
-  return beluga::estimate(beluga::views::states(particles_), beluga::views::weights(particles_));
+  return beluga::cluster_based_estimate(beluga::views::states(particles_), beluga::views::weights(particles_));
```
hidmic, line 101: I'm OK with everything else but I still think this shouldn't be made default. We can make it optional or keep `beluga::estimate` until we have proof this is all around the better choice.

fbattocchia: In `LikelihoodField`, there's no difference in the results if you use `cluster_based_estimate`, but in `LikelihoodFieldProb`, there is a difference in the results; they weren't as good without `cluster_based_estimate`. Could clustering be used only for the `LikelihoodFieldProb` model? If it's not the default option, should the report be regenerated?...Anyway, I think it would be better to investigate this part. At the time, I created an [issue](https://github.com/Ekumen-OS/beluga/issues/477) with this idea.

hidmic: > In `LikelihoodField`, there's no difference in the results if you use cluster_based_estimate

:eyes: that wasn't obvious from the benchmark reports. Comparing these side by side with older reports, it seems to be true. If so, then I'm OK with making cluster based estimate the default.

hidmic (approved):
LGTM with green CI

glpuga:
@fbattocchia CI is failing because a part of the new code has no test coverage. We are missing an integration test to ensure this code fragments is tested (you can see this information if you follow the [Codecov link](https://app.codecov.io/gh/Ekumen-OS/beluga/pull/476?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=Ekumen-OS) in the CI error information).

Screenshot from 2025-04-22 14-12-54

I think all we need to do is to add the new model to this pre-existing integration test: https://github.com/Ekumen-OS/beluga/blob/main/beluga_amcl/test/test_amcl_node.cpp#L209

fbattocchia:
Yes, I was planning to review the tests to improve coverage. I added this line `std::make_tuple("differential_drive", "likelihood_field_prob")`. Is that correct, or do you mean adding `std::make_tuple("differential_drive", "likelihood_field", "likelihood_field_prob")`?

glpuga:
> I added this line `std::make_tuple("differential_drive", "likelihood_field_prob")`. Is that correct, or do you mean adding `std::make_tuple("differential_drive", "likelihood_field", "likelihood_field_prob")`?

What you added is correct. Each tuple represents a combination of a motion model and a sensor model. We only need that at least one combination includes the prob sensor model.

glpuga:
> I added this line `std::make_tuple("differential_drive", "likelihood_field_prob")`. Is that correct, or do you mean adding `std::make_tuple("differential_drive", "likelihood_field", "likelihood_field_prob")`?

What you added is correct. Each tuple represents a combination of a motion model and a sensor model. We only need that at least one combination includes the prob sensor model.

fbattocchia:
@glpuga The coverage is OK. Is it possible to merge? Who is responsible for doing the merge? It appears blocked to me.

glpuga:
> @glpuga The coverage is OK. Is it possible to merge? Who is responsible for doing the merge? It appears blocked to me.

It's still failing the DCO check , maybe because of the merge-from-main commits?

Please squash all commits into a single one, then sign and rebase that one onto the current head of main and push.

fbattocchia:
@glpuga @hidmic all check have passed

glpuga (approved):

glpuga:
> @glpuga @hidmic all check have passed

Great work @fbattocchia !

PR #487, opened 2025-07-11
Likelihood field map representation

LaBruma:
### Proposed changes

- Apply correct transformation to likelihood_field map to correctly superpose with particle filter map
- Change discrete representation of likelihood_field map to continuous representation in the range [0, 100]
- Avoid publishing on topic likelihood_field if the sensor used doesn't allow this kind of representation  

#### Type of change

### Checklist

### DIFF
--- a/beluga_amcl/src/amcl_node.cpp
+++ b/beluga_amcl/src/amcl_node.cpp
@@ -186,14 +186,12 @@ AmclNode::AmclNode(const rclcpp::NodeOptions& options) : BaseAMCLNode{"amcl", ""
 
 AmclNode::~AmclNode() {
   RCLCPP_INFO(get_logger(), "Destroying");
-  // In case this lifecycle node wasn't properly shut down, do it here
+  // In case this lifecycle node wasn't properly shut down, do it here.
   on_shutdown(get_current_state());
 }
 
 void AmclNode::do_activate(const rclcpp_lifecycle::State&) {
-  // Creation of the likelihood field publisher is attached to the creation of the particle filter
-  // (and knowledge of sensor model used). Hence publication of the likelihood field and activation control
-  // is defined at this stage.
+  // Ensure likelihood field publisher is (re)activated early enough.
   if (likelihood_field_pub_) {
     likelihood_field_pub_->on_activate();
   }
@@ -254,7 +252,6 @@ void AmclNode::do_deactivate(const rclcpp_lifecycle::State&) {
   laser_scan_filter_.reset();
   laser_scan_sub_.reset();
   global_localization_server_.reset();
-  // Adding deactivation control on likelihood fields
   if (likelihood_field_pub_) {
     likelihood_field_pub_->on_deactivate();
   }
@@ -404,14 +401,10 @@ void AmclNode::map_callback(nav_msgs::msg::OccupancyGrid::SharedPtr map) {
       RCLCPP_ERROR(get_logger(), "Could not initialize particle filter: %s", error.what());
       return;
     }
-
-    // Conditionally creating likelihood_field publisher (if sensor model supports it)
-    if (particle_filter_->supports_likelihood_field()) {
-      // subscribe to likelihood_field (and couple to particle_filter)
+    if (particle_filter_->has_likelihood_field()) {
       likelihood_field_pub_ =
           create_publisher<nav_msgs::msg::OccupancyGrid>("likelihood_field", rclcpp::SystemDefaultsQoS());
-
-      // activate publisher for the very first time
+      // Activate publisher immediately, we are likely past the activation phase.
       likelihood_field_pub_->on_activate();
     }
 
--- a/beluga_ros/include/beluga_ros/amcl.hpp
+++ b/beluga_ros/include/beluga_ros/amcl.hpp
@@ -32,7 +32,7 @@
 #include <beluga/random.hpp>
 #include <beluga/sensor.hpp>
 #include <beluga/sensor/data/value_grid.hpp>
-#include <beluga/sensor/sensor_primitives.hpp>
+#include <beluga/sensor/primitives.hpp>
 #include <beluga/views/sample.hpp>
 
 #include <beluga_ros/laser_scan.hpp>
@@ -176,8 +176,8 @@ class Amcl {
     return *result;  // return by value
   }
 
-  /// Indicates if the sensor model supports likelihood fields.
-  [[nodiscard]] bool supports_likelihood_field() const {
+  /// Check if the sensor model bears a likelihood field.
+  [[nodiscard]] bool has_likelihood_field() const {
     return std::visit(
         [](const auto& sensor_model) {
           using T = std::decay_t<decltype(sensor_model)>;
--- a/beluga_ros/include/beluga_ros/likelihood_field.hpp
+++ b/beluga_ros/include/beluga_ros/likelihood_field.hpp
@@ -50,17 +50,18 @@ inline void assign_likelihood_field(
   const float max_val = *max_it;
   const float range = max_val - min_val;
 
-  // Handle degenerate case (Flat grid). That case fill the grid with zeros
+  // Handle degenerate case (flat grid) by filling the occupancy grid with zeros
   if (range <= std::numeric_limits<float>::epsilon()) {
     // All values are the same; treat as unknown or flat
     std::fill(message.data.begin(), message.data.end(), 0);
     return;
   }
 
-  // Normalizing each cell to [0, 100]
+  // Normalizing each cell to [0, 100] - To be consistent with nav2:
+  // navigation2/nav2_costmap_2d/src/costmap_2d_publisher.cpp
   for (std::size_t i = 0; i < grid_data.size(); ++i) {
     const float normalized = (grid_data[i] - min_val) / range;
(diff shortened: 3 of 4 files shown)


beluga_amcl/config/Amcl.cfg
```diff
@@ -263,6 +263,12 @@ gen.add(
     default=0.05, min=0., max=1.
 )
 
+gen.add(
+    "model_unknown_space", bool_t, 0,
+    "Whether to model unknown space or assume it free.",
+    default=False,
+)
```
hidmic, line 270: @LaBruma this is stacked on top of #448, isn't it? If so, we should target that branch instead of `main`.

hidmic (commented):
@LaBruma I see this is mostly a work in progress so I'll defer further review until next week. In the meantime, mind to update the PR name to be more accurate?

beluga_amcl/src/amcl_node.cpp
```diff
@@ -191,12 +191,18 @@ AmclNode::~AmclNode() {
 }
 
 void AmclNode::do_activate(const rclcpp_lifecycle::State&) {
+  // Adding activation control on likelihood fields
```
hidmic, line 194: @LaBruma worth adding a comment as to why this is up here.

beluga_ros/include/beluga_ros/amcl.hpp
```diff
@@ -132,11 +132,37 @@ class Amcl {
       const AmclParams& params,
       execution_policy_variant execution_policy);
 
-  /// Returns a reference to the current set of particles.
+  /// Returns a reference to the current set of particles
```
hidmic, line 135: @LaBruma nit: why remove the trailing dot?

beluga_ros/include/beluga_ros/amcl.hpp
```diff
@@ -132,11 +132,37 @@ class Amcl {
...
   [[nodiscard]] const auto& particles() const { return particles_; }
 
   /// Returns a reference to the current likelihood field.
-  [[nodiscard]] const auto& get_likelihood_field() const { return likelihood_field_; }
+  [[nodiscard]] const auto& get_likelihood_field() const {
+    if (std::holds_alternative<beluga::LikelihoodFieldModel<beluga_ros::OccupancyGrid>>(sensor_model_)) {
```
hidmic, line 140: @LaBruma this is good, but just as an FYI, you could have a `has_likelihood_field_v` trait doing something along the lines of https://github.com/Ekumen-OS/beluga/blob/main/beluga/include/beluga/primitives.hpp:

```c++

template <class T, class = void>
struct has_likelihood_field : std::false_type {};

template <class T>
struct has_likelihood_field<T, std::void_t<decltype(std::declval<T>().likelihood_field())>> : std::true_type {};

template <class T>
inline constexpr bool has_likelihood_field_v = has_likelihood_field<T>::value;
```

Then, you could:

```c++
return std::visit([](const auto& sensor_model) {
    if (!has_likelihood_field_v<std::decay_t<decltype(sensor_model)>>) {
        throw std::runtime_error();  
    }
    return sensor_model.likelihood_field();
}, sensor_model_);
```
And so and so.

beluga_ros/include/beluga_ros/amcl.hpp
```diff
@@ -132,11 +132,37 @@ class Amcl {
...
+      return std::get<beluga::LikelihoodFieldModel<beluga_ros::OccupancyGrid>>(sensor_model_).likelihood_field();
+    }
+    if (std::holds_alternative<beluga::LikelihoodFieldProbModel<beluga_ros::OccupancyGrid>>(sensor_model_)) {
+      return std::get<beluga::LikelihoodFieldProbModel<beluga_ros::OccupancyGrid>>(sensor_model_).likelihood_field();
+    }
+    throw std::invalid_argument(std::string("Invalid sensor argument: Likelihood_field"));
```
hidmic, line 146: @LaBruma consider using `std::runtime_error` instead, this getter takes no arguments.

beluga_ros/include/beluga_ros/amcl.hpp
```diff
@@ -132,11 +132,37 @@ class Amcl {
       const AmclParams& params,
       execution_policy_variant execution_policy);
 
-  /// Returns a reference to the current set of particles.
+  /// Returns a reference to the current set of particles
   [[nodiscard]] const auto& particles() const { return particles_; }
 
   /// Returns a reference to the current likelihood field.
-  [[nodiscard]] const auto& get_likelihood_field() const { return likelihood_field_; }
+  [[nodiscard]] const auto& get_likelihood_field() const {
```
hidmic, line 139: @LaBruma nit: throughout the codebase, we usually use the noun without a leading `get_`. See `particles()` for instance. Consider this just being `likelihood_field()`. Same for likelihood_field_origin()`.

beluga_ros/include/beluga_ros/likelihood_field.hpp
```diff
@@ -30,26 +31,37 @@ namespace beluga_ros {
 /// Convert from likelihood data to ROS2 message.
 inline void assign_likelihood_field(
     const beluga::ValueGrid2<float>& likelihood_field,
```
hidmic, line 33: @LaBruma consider making this a template function on the type of `ValueGrid2`.

beluga_ros/include/beluga_ros/likelihood_field.hpp
```diff
@@ -30,26 +31,37 @@ namespace beluga_ros {
...
+    const Sophus::SE2d& origin,
     nav_msgs::msg::OccupancyGrid& message,
     double free_threshold = 0.2,
     double occupied_threshold = 0.25) {
```
hidmic, line 37: @LaBruma I think we no longer need these thresholds.

beluga_ros/include/beluga_ros/likelihood_field.hpp
```diff
@@ -30,26 +31,37 @@ namespace beluga_ros {
...
-      message.data[i] = 0;  // Free
-    } else if (grid_data[i] >= occupied_threshold) {
-      message.data[i] = 100;  // Occupied
-    } else {
-      message.data[i] = -1;  // Unknown
-    }
+  // Normalizing each cell to [0, 100]
+  for (std::size_t i = 0; i < grid_data.size(); ++i) {
+    const float normalized = (grid_data[i] - min_val) / range;
+    message.data[i] = static_cast<int8_t>(normalized * 100.0f);  // Scale to [0, 100]
```
hidmic, line 64: @LaBruma out of curiosity, why fit to the [0, 100] interval? Why not [0, 255]?

beluga_ros/include/beluga_ros/amcl.hpp
LaBruma: Like it. It is slightly more complex but it avoids modifying the code if a future sensor class (supporting likelihood fields) is defined. I have no idea how probable is that to happen but if could be a more "dry" implementation. Would you like me to change it?

beluga_ros/include/beluga_ros/amcl.hpp
LaBruma: Got it. Changed to:

throw std::runtime_error("The current sensor model does not support likelihood field");

beluga_ros/include/beluga_ros/likelihood_field.hpp
LaBruma: I found (like [here](https://wiki.ros.org/rviz/DisplayTypes/Map)) in some places that map representation of gray scale maps was in that range. But I also saw that setting to [0,255] also produced a change in the visualization. I'll change it to that full 8-bit representation.

beluga_ros/include/beluga_ros/likelihood_field.hpp
LaBruma: Done

LaBruma:
Hey @hidmic! I’ve addressed all the changes you requested. PTAL.

hidmic (approved):
Final pass. Sorry for backtracking in a couple places. It only became apparent now.

Feel free to address / resolve and merge.

beluga_ros/include/beluga_ros/amcl.hpp
```diff
@@ -136,7 +137,54 @@ class Amcl {
...
+
+    if (!result) {
+      throw std::runtime_error("The current sensor model does not support likelihood field");
+    }
+
+    return *result;  // return by value
+  }
+
+  /// Indicates if the sensor model supports likelihood fields.
+  [[nodiscard]] bool supports_likelihood_field() const {
```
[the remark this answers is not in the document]
LaBruma: You're right. Changed.

beluga_amcl/src/amcl_node.cpp
```diff
@@ -391,6 +404,17 @@ void AmclNode::map_callback(nav_msgs::msg::OccupancyGrid::SharedPtr map) {
       RCLCPP_ERROR(get_logger(), "Could not initialize particle filter: %s", error.what());
       return;
     }
+
+    // Conditionally creating likelihood_field publisher (if sensor model supports it)
```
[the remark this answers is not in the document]
LaBruma: Removed.

PR #492, opened 2025-08-05
Add timer-based propagation functionality to AMCL

fbattocchia:
### Proposed changes

-  Add 'increase_propagation' parameter (0-100 Hz) for timer-based propagation
-  Implement propagation_timer_callback() for forced propagation updates
-  Add update_propagation() method to Amcl class for motion-only updates
-  Add comprehensive tests for propagation timer functionality
-  Use MutuallyExclusive callback group for thread safety
-  Timer disabled by default (frequency = 0.0)

This feature allows for increased propagation rates independent of sensor updates. 
This improvement allows the implementation of the speed-based motion model

#### Type of change

### Checklist


beluga_amcl/src/amcl_node.cpp
```diff
@@ -466,26 +488,51 @@ void AmclNode::do_periodic_timer_callback() {
...
+
+void AmclNode::laser_callback(sensor_msgs::msg::LaserScan::ConstSharedPtr laser_scan) {
+  if (!particle_filter_) {
+    RCLCPP_WARN_THROTTLE(
+        get_logger(), *get_clock(), 2000, "Ignoring laser data because the particle filter has not been initialized");
+    return;
+  }
+
+  // Get base pose in odom frame at laser scan timestamp
+  auto base_pose_in_odom = get_base_pose_in_odom(tf2_ros::fromMsg(laser_scan->header.stamp));
```
glpuga, line 534: I wonder if it can happen that we get a lidar message with a timestamp that is previous to a get_base_pose_in_odom() update.

Say, we execute get_base_pose_in_odom() at 10, 20, 30, 40, 50 and then we get a lidar message with timestamp 45. We might be adding a little bit more noise then.

Maybe when we update the motion model from a separate thread then we should either 
1. Only update if the time stamp is later (no noise added then). If we don't, particle state might be up to one timer event behind sensor data.
2. Repeat the value of the latest get_base_pose_in_odom() odom value we updated (so that no extra noise is ever added. We will **always** be up to one timer event behind sensor timestamp.
3. Always update. If we get "late" lidar messages, we might "roll back" motion from the particle state in between timer activations and in the end, end up adding more noise.

I tend to like more 1 or 2. 3 might add a lot more noise if the late arrivers are recurrent.
Say we update motion at 40hz, get lidar messages at 10hz, and lidar messages always end up arriving after the timer for a later timestamp already updated the filter. 1 in four message will add false motion and increase noise.

Thoughts @fbattocchia @hidmic ?

beluga_ros/src/amcl.cpp
```diff
@@ -50,6 +50,20 @@ void Amcl::update_map(beluga_ros::OccupancyGrid map) {
...
+void Amcl::update_propagation(Sophus::SE2d base_pose_in_odom) {
+  if (particles_.empty()) {
+    return;
+  }
+
+  // Force propagation without checking update_policy_
+  std::visit(
+      [&, this](auto& policy, auto& motion_model) {
+        particles_ |= beluga::actions::propagate(policy, motion_model(control_action_window_ << base_pose_in_odom)) |
+                      beluga::actions::normalize(policy);
```
glpuga, line 62: We don't need to normalize, since we only affected distribution density, but we left the weights unchanged.

beluga_ros/test/test_amcl.cpp
```diff
@@ -149,4 +149,10 @@ TEST(TestAmcl, UpdateWithParticlesForced) {
   ASSERT_TRUE(estimate.has_value());
 }
 
+TEST(TestAmcl, UpdatePropagationWithNoParticles) {
+  auto amcl = make_amcl();
+  amcl.update_propagation(Sophus::SE2d{});
+  ASSERT_EQ(amcl.particles().size(), 0);
+}
+
```
glpuga, line 157: I get that this manages to make the coverage checker think that the function was tested, but it does not look like its functionally testing anything. Let's do a propagation step with a few particles in the set.

glpuga (changes requested):
Great work @fbattocchia ! left a few comments.

beluga_amcl/src/amcl_node.cpp
glpuga: > I wonder if it can happen that we get a lidar message with a timestamp that is previous to a get_base_pose_in_odom() update.
> 
> Say, we execute get_base_pose_in_odom() at 10, 20, 30, 40, 50 and then we get a lidar message with timestamp 45. We might be adding a little bit more noise then because we'll falsely "roll back" motion and then apply it again when we execute the timer at 60 later.
> 
> Maybe when we update the motion model from a separate thread then we should either
> 
> 1. Only update if the time stamp is later (no noise added then). If we don't, particle state might be up to one timer event behind sensor data.
> 2. Repeat the value of the latest get_base_pose_in_odom() odom value we updated (so that no extra noise is ever added. We will **always** be up to one timer event behind sensor timestamp.
> 3. Always update. If we get "late" lidar messages, we might "roll back" motion from the particle state in between timer activations and in the end, end up adding more noise.
> 
> I tend to like more 1 or 2. 3 might add a lot more noise if the late arrivers are recurrent. Say we update motion at 40hz, get lidar messages at 10hz, and lidar messages always end up arriving after the timer for a later timestamp already updated the filter. 1 in four message will add false motion and increase noise.
> 
> Thoughts @fbattocchia @hidmic ?

glpuga: > Say, we execute get_base_pose_in_odom() at 10, 20, 30, 40, 50 and then we get a lidar message with timestamp 45. We might be adding a little bit more noise then.

I mean, we would be "rolling back" motion from 50 to 45, then forwarding again from 45 to 60 in the next timer event, thus adding extra "motion" and thus noise (proportional to motion).

beluga_amcl/src/amcl_node.cpp
```diff
@@ -466,26 +488,51 @@ void AmclNode::do_periodic_timer_callback() {
...
+  // Get current base pose in odom frame (latest available)
+  auto base_pose_in_odom = get_base_pose_in_odom(tf2::TimePointZero);
+  if (!base_pose_in_odom.has_value()) {
+    return;
+  }
+
+  // Force a propagation-only update (without sensor data)
+  particle_filter_->update_propagation(base_pose_in_odom.value());
+
+  RCLCPP_INFO(get_logger(), "Forced propagation update executed");
```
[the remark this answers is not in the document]
hidmic: +1. Consider demoting it to DEBUG.

beluga_ros/include/beluga_ros/amcl.hpp
```diff
@@ -210,6 +210,16 @@ class Amcl {
...
 
+  /// Update particles based on motion only (propagation only).
+  /**
+   * This method only performs the propagation step of the particle filter update,
+   * applying the motion model without any sensor correction. Useful for forced
+   * propagation at regular intervals without waiting for sensor data.
+   *
+   * \param base_pose_in_odom Base pose in the odometry frame.
+   */
+  void update_propagation(Sophus::SE2d base_pose_in_odom);
```
hidmic, line 221: @fbattocchia consider making this just another `update` overload.
```suggestion
  void update(const Sophus::SE2d& base_pose_in_odom);
```
It is updating after all, it just uses less data.

beluga_ros/src/amcl.cpp
hidmic: +1

beluga_amcl/src/amcl_node.cpp
hidmic: > I wonder if it can happen that we get a lidar message with a timestamp that is previous to a get_base_pose_in_odom() update.

That's a very good point, and yes, this can absolutely happen. It's an out of sequence measurement (OOSM) case.  A principled, general solution is probably out of scope for your purposes though (e.g. https://ieeexplore.ieee.org/document/8455401, but there is quite a bit of research on the topic).

> I tend to like more 1 or 2. 3 might add a lot more noise if the late arrivers are recurrent.

:thinking: What if we don't touch the filter itself? If we let it evolve at lidar rates, we can still propagate the latest distribution up to current time on timer callback. You could even do that deterministically on the latest estimate by applying an unscented transform with the motion nonlinear function (which we don't have, we just have a sampler, but it could be added) and spare the cost of sampled propagation.

hidmic: Ahh, I see now. I thought you wanted estimates at a rate higher than the lidar rate, but this is about processing odometry at a higher rate than lidar rate. 

If it were the former, what I meant is that you can always predict forward from the latest estimate (which is almost the same as relying on odometry given a fixed map to odom transform but for the explicit uncertainty modelling). 

But I sense it's the latter, so +100 to lazy updates. Maybe collect all odometry updates in a bigger control window.

beluga_ros/test/test_amcl.cpp
fbattocchia: Yes, I did the test to achieve coverage, but it doesn't make much sense. What I need to test is that it doesn't do anything because there are no particles. I changed the function's logic so I wouldn't have to do that check. (
Another option would be to return a bool and verify the return in the test.)

beluga_amcl/src/amcl_node.cpp
fbattocchia: I think it's a good idea to implement lazily updates.

beluga_ros/src/amcl.cpp
```diff
@@ -50,6 +50,26 @@ void Amcl::update_map(beluga_ros::OccupancyGrid map) {
...
 
+void Amcl::process_buffered_odometry_until(std::deque<OdometryMotion>& buffer, const tf2::TimePoint& until) {
+  while (!buffer.empty()) {
+    const auto& [odom_time, odom_pose] = buffer.front();
+    if (odom_time >= until)
+      break;
+    update(odom_pose);
+    buffer.pop_front();
+  }
+}
```
glpuga, line 61: it sounds to me like this belongs in beluga_amcl.cpp, justl ike the OdometryMotion datatype, since implementing the odometry message queue is that file's concern.

beluga_ros/src/amcl.cpp
```diff
@@ -50,6 +50,26 @@ void Amcl::update_map(beluga_ros::OccupancyGrid map) {
   std::visit([&](auto& sensor_model) { sensor_model.update_map(std::move(map)); }, sensor_model_);
 }
 
+void Amcl::process_buffered_odometry_until(std::deque<OdometryMotion>& buffer, const tf2::TimePoint& until) {
+  while (!buffer.empty()) {
+    const auto& [odom_time, odom_pose] = buffer.front();
+    if (odom_time >= until)
```
glpuga, line 56: I would still process it if it's ==

beluga_ros/include/beluga_ros/amcl.hpp
```diff
@@ -96,6 +96,9 @@ struct AmclParams {
   double spatial_resolution_theta = 10 * Sophus::Constants<double>::pi() / 180;
 };
 
+/// Type Buffer for queued odometry motions (timestamp, pose)
+using OdometryMotion = std::pair<tf2::TimePoint, Sophus::SE2d>;
```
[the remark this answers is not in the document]
glpuga: Nevermind, if we move this into a private type in  beluga_amcl, then a tuple is ok.

beluga_amcl/src/amcl_node.cpp
```diff
@@ -268,6 +287,7 @@ void AmclNode::do_cleanup(const rclcpp_lifecycle::State&) {
   particle_filter_.reset();
   likelihood_field_pub_.reset();  // attaching likelihood_field_pub_ lifespan to particle_filter_ lifespan
   enable_tf_broadcast_ = false;
+  odometry_motion_buffer_.clear();
```
glpuga, line 290: We should also clear this whenever the particle filter is initialized/created.

beluga_ros/src/amcl.cpp
hidmic: My only comment here, and it is as meta as https://github.com/Ekumen-OS/beluga/pull/492#discussion_r2285113313 was, is that processing a longer window of odometry measurements is something that could be pushed to the core motion model. This is like conditioning the proposal distribution on a control sequence that is longer than two elements.

Really, it is a non-issue here. Keep it simple. It can be explored in follow-up work.

beluga_amcl/src/amcl_node.cpp
```diff
@@ -183,6 +183,12 @@ AmclNode::AmclNode(const rclcpp::NodeOptions& options) : BaseAMCLNode{"amcl", ""
         "and ignore subsequent ones.";
     declare_parameter("first_map_only", false, descriptor);
   }
+
+  {
+    auto descriptor = rcl_interfaces::msg::ParameterDescriptor();
+    descriptor.description = "Set true to enable increased propagation, and set false to disable it.";
+    declare_parameter("propagation_rate", rclcpp::ParameterValue(false), descriptor);
```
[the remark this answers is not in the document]
fbattocchia: maybe use_odometry_propagation?

beluga_amcl/src/amcl_node.cpp
fbattocchia: I'm clearing the buffer in `do_activate()` where the odometry subscription is performed

beluga_amcl/src/amcl_node.cpp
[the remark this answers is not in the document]
glpuga: Better `odometry_driven_propagation`

beluga_amcl/src/amcl_node.cpp
glpuga: Yes, but that only happens when the node starts. The filter might be reinitialized multiple times afterwards, whenever the map is updated or when the user sends a pose intialization message. Look for calls to `initialize_from_estimate()` in amcl_node.cpp

beluga_amcl/include/beluga_amcl/amcl_node.hpp
```diff
@@ -143,6 +154,9 @@ class AmclNode : public BaseAMCLNode {
   /// Connection for laser scan updates filter and callback.
   ::message_filters::Connection laser_scan_connection_;
 
+  /// Odometry updates subscription.
+  rclcpp::Subscription<nav_msgs::msg::Odometry>::SharedPtr odom_sub_;
+
```
glpuga, line 159: nit, but for you to know:

In this node it does not really matter because this is a lifecycle node and thus subscribers get destroyed before the wrapping class does, but its a good habit to put anything that has a callback int the very last position in the class members declarations.

The reason is so that whenever the class gets destroyed, those are the first to be destroyed, and that includes rendering them unable to execute their callbacks. 

If they are not last, some datastructure their callbacks depend on might be destroyed first concurrently with something triggering the callback, and that will usually cause a crash. 

This is a frequent cause of flaky tests and nodes that randomly seem to crash when being stopped.

beluga_amcl/src/amcl_node.cpp
```diff
@@ -251,6 +257,18 @@ void AmclNode::do_activate(const rclcpp_lifecycle::State&) {
...
   RCLCPP_INFO(get_logger(), "Created request_nomotion_update service");
+
+  {
+    // Subscribe to odometry topic to buffer odometry motions
+    if (get_parameter("use_odometry_propagation").as_bool()) {
+      odom_sub_ = create_subscription<nav_msgs::msg::Odometry>(
+          get_parameter("odom_topic").as_string(), rclcpp::SensorDataQoS(),
+          std::bind(&AmclNode::odometry_callback, this, std::placeholders::_1), common_subscription_options_);
+
+      RCLCPP_INFO(get_logger(), "Subscribed to odom_topic: %s", odom_sub_->get_topic_name());
```
glpuga, line 268:
```suggestion
      RCLCPP_INFO(get_logger(), "Subscribed to odometry topic: %s", odom_sub_->get_topic_name());
```

beluga_amcl/src/amcl_node.cpp
```diff
@@ -466,13 +486,45 @@ void AmclNode::do_periodic_timer_callback() {
...
     RCLCPP_WARN_THROTTLE(
         get_logger(), *get_clock(), 2000, "Ignoring laser data because the particle filter has not been initialized");
     return;
   }
 
+  // If use_odometry_propagation is enabled, process odometry buffer up to lidar timestamp
+  if (get_parameter("use_odometry_propagation").as_bool()) {
+    const auto laser_scan_stamp = tf2_ros::fromMsg(laser_scan->header.stamp);
+    process_buffered_odometry_until(odometry_motion_buffer_, laser_scan_stamp);
+  }
```
glpuga, line 525: nit, but if odometry driven prop is off, the queue will be empty anyway.

glpuga (commented):
Great job! Final comments.

beluga_amcl/src/amcl_node.cpp
fbattocchia: @glpuga Removing the if statement would make the code more compact, but in my opinion, less explicit. Keeping it helps clarify that this action is only performed when use_odometry_propagation is enabled, which can be beneficial for someone reading or maintaining the code.

hidmic:
> @hidmic I'd like to have a devel or experimental branch in the repository to buffer changes. I'd rather no longer hit main with changes we have not measured to ensure they have an objective reason to exist and to ensure they don't represent a regression by some other measure.

:thinking: I'm on the fence. We don't provide stability guarantees on `main`. We have release branches and tags for that. How would the transition work? How is it different from ad-hoc feature branches with large patches? I really like the idea of the data driven process for vetting contributions but I want to weigh it against the cost of contribution. It's already difficult to land anything on `main`, and if we add another gate without a clear process to land a patch (which I know we don't really have, it's crafting dedicated experiments and rigs to convince ourselves) no one other than us will even dare to try.

PR #530, opened 2025-10-02
Add new overload for reweight action that tracks likelihood

### LINKED ISSUE #524: Add reweighting action that tracks measurement likelihood
### Feature description

`beluga::actions::reweight` computes measurement likelihood given a state hypothesis (aka particle) and promptly uses it to shape the posterior distribution. However, there are particle filters that perform additional steps based on raw likelihood (e.g. MH-AMCL at #52 is an obvious example of this). 

We should add functionality to keep likelihood figures around after a reweight.

### Implementation considerations

I'm naively inclined towards adding a view that computes likelihoods, and then use that in a `beluga::actions::reweight` overload to update weights and optionally store likelihood in the particle iff the particle has storage for it, but I'm open to other ideas.

JesusSilvaUtrera:
### Proposed changes

Closes #524. Changes include a new overload for the `reweight` action, tracking likelihood as well and storing it in the particle structure if there is a member for it.

#### Type of change

### Checklist

### Additional comments

### DIFF
--- a/beluga/include/beluga/actions/reweight.hpp
+++ b/beluga/include/beluga/actions/reweight.hpp
@@ -17,13 +17,14 @@
 
 #include <algorithm>
 #include <execution>
+#include <type_traits>
 
 #include <beluga/type_traits/particle_traits.hpp>
+#include <beluga/views/likelihoods.hpp>
 #include <beluga/views/particles.hpp>
 
 #include <range/v3/action/action.hpp>
-#include <range/v3/algorithm/max_element.hpp>
-#include <range/v3/view/common.hpp>
+#include <range/v3/view/zip.hpp>
 
 /**
  * \file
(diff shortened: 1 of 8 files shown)


JesusSilvaUtrera:
I am open to discussions about implementation, because @hidmic had the idea of using a view for the likelihood for example, but I wasn't so sure about that (that would mean to create the view before calling the action and passing it as argument, wouldn't it?).

Also, while researching about C++20 features for this, I found the 'concepts' feature, and I think that would be a good refactor for the templates restrictions.

hidmic:
> Also, while researching about C++20 features for this, I found the 'concepts' feature, and I think that would be a good refactor for the templates restrictions.

When and if we switch to C++20, yes.

beluga/include/beluga/actions/reweight.hpp
```diff
@@ -68,6 +68,50 @@ struct reweight_base_fn {
...
+   */
+  template <
+      class ExecutionPolicy,
+      class Range,
+      class Model,
+      class LikelihoodType,
+      class ParticleType,
+      std::enable_if_t<std::is_execution_policy_v<std::decay_t<ExecutionPolicy>>, int> = 0>
+  constexpr auto
+  operator()(ExecutionPolicy&& policy, Range& range, Model model, LikelihoodType ParticleType::*likelihood_member) const
```
hidmic, line 96: @JesusSilvaUtrera thanks for the PR! Consider the following alternative implementation.

Right above (line 67), you can see the original `reweight` transforming particle weights by multiplying the previous weight with the result of evaluating the model at each particle state. So what about if we introduce a `weight` view that just gets you the result of applying the model? Then, we could have a `reweight` overload that takes those weights instead of the model. Then we could those weights to compute other quantities, like the likelihood of an entire cluster :eyes: 

On a related note, I'd rather not add redundant information to the particle. The `weight` already contains `likelihood` information. But I could be convinced otherwise.

JesusSilvaUtrera:
@hidmic Regarding this, do you think this can be merged and other people can take on the rest of issues pending? I don't want to leave this hanging, I want to either close it so it can be continued by another person or merge it if you think adds value.

My priorities for this year are mainly focused on other subjects so I suspect I won't have time to allocate to this effort, that's why I wanted to double check. 

Thank you in advance!

beluga/include/beluga/actions/reweight.hpp
```diff
@@ -80,9 +102,12 @@ struct reweight_base_fn {
   }
 
   /// Overload that returns a view closure to compose with other views.
-  template <class ExecutionPolicy, class Model, std::enable_if_t<std::is_execution_policy_v<ExecutionPolicy>, int> = 0>
-  constexpr auto operator()(ExecutionPolicy policy, Model model) const {
-    return ranges::make_action_closure(ranges::bind_back(reweight_base_fn{}, std::move(model), std::move(policy)));
+  template <
+      class ExecutionPolicy,
+      class ThirdArg,
```
hidmic: @JesusSilvaUtrera hmm, mind to explain?

beluga/include/beluga/actions/reweight.hpp
```diff
@@ -91,16 +116,36 @@ struct reweight_fn : public reweight_base_fn {
...
+      class Model,
+      std::enable_if_t<ranges::range<Range>, int> = 0,
+      std::enable_if_t<!ranges::range<Model>, int> = 0>
   constexpr auto operator()(Range&& range, Model model) const -> Range& {
     return (*this)(std::execution::seq, std::forward<Range>(range), std::move(model));
   }
 
   /// Overload that returns a view closure to compose with other views.
-  template <class Model>
+  template <class Model, std::enable_if_t<!ranges::range<Model>, int> = 0>
```
hidmic: @JesusSilvaUtrera meta: let me see if I understand. To reweight we need either particle likelihoods or a model to compute those particle likelihoods, thus these SFINAE gated overloads. Now what if the `reweight_base_fn` implementation took either? If we only take model, and we perform an `if constexpr` check to decide if we need to apply the `beluga::views::likelihoods` view there, then we need fewer overloads, don't we? WDYT?

hidmic (commented):
Ahh, man, I completely dropped the ball here. 

Code's looking so much better, we should move forward!

PR #544, opened 2026-01-07
Add random probability term to landmark sensor model

PaulVerhoeckx:
### Proposed changes

This PR adds a configurable `random_prob` parameter to the landmark sensor model, enabling it to cope with false positive measurements. 
Without it, all particle weights will approach zero, making the distribution vulnerable to collapse.

#### Type of change

### Checklist

### Additional comments

Would be nice if this would be backported to the humble release

### DIFF
--- a/beluga/test/beluga/sensor/test_landmark_sensor_model.cpp
+++ b/beluga/test/beluga/sensor/test_landmark_sensor_model.cpp
@@ -167,7 +167,8 @@ TYPED_TEST(LandmarkSensorModelTests, NoSuchLandmark) {
   const auto pose = get_robot_pose_in_world<typename TypeParam::state_type>();
   // test case where there is not landmark in the map of the detected id
   auto map = LandmarkMap(default_map_boundaries, {{{0.0, 1.0, 0.0}, 99}});
-  auto sensor_model = TypeParam{get_default_model_params(), std::move(map)};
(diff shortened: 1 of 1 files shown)


hidmic (commented):
Thanks for the contribution @PaulVerhoeckx!

Patch looks reasonable. I'm curious about your use case. I presume the distribution no longer collapses on a bad measurement but have you got around to measure its effect on the estimate?

PaulVerhoeckx:
> Thanks for the contribution @PaulVerhoeckx!
> 
> Patch looks reasonable. I'm curious about your use case. I presume the distribution no longer collapses on a bad measurement but have you got around to measure its effect on the estimate?

Measuring the exact impact has been tricky. The change was motivated by a ROS bag where I saw the particles collapse after a bad measurement; in other similar situations the filter handled things just fine.

When I later replayed Beluga’s input data to compare behavior before and after the change, the collapse didn’t reproduce anymore, even without my patch. My guess is that this comes down to the estimator being non-deterministic (sampling/resampling effects), which makes this kind of failure hard to trigger reliably.

That said, I think the idea itself is sound and in line with existing robustness mechanisms in Beluga, like the `z_rand` term in the beam model, which serves a similar purpose.

glpuga (approved):
LGTM

Issue #265, opened 2023-11-05
Thick map walls cause a lot of uncertainty to the likelihood sensor model, damaging localization solution


glpuga:
### Bug description

Working with a synthetic map created from a floor plan, I notice that when the mapped walls around the robot are thick, the likelihood model tends to accumulate a lot of uncertainty in both position and orientation extremely fast.

For a very extreme case of this, see this case, where the robot immediately delocalized itself entering this passage with large areas marked as obstacles around it. It can be seen in the particle cloud that large volumes of the position belief are distributed lateral to the robot, even within the walls themselves.
 
![Screenshot from 2023-11-04 20-16-43](https://github.com/Ekumen-OS/beluga/assets/17802342/074704ba-d559-4ba6-aaa2-70834479515b)

This behavior is shared by both `beluga_amcl` and `nav2_amcl` and its due to the "see through walls" nature of the standard implementation of the likelihood model. Using the beam model in the same map works just fine, since ray-tracing is impervious to the thickness of the walls.

While it can be argued _it is a feature of the model and therefore maps should not be created like this_, arguments can also be set for addressing this:

1. it's not obvious to the user why a "perfect" map such as this would not work, unless they know about how the sensor model works.
2. this effect can also result from people editing capture-generated maps.
3. capture-generated maps can also have multi-cell width walls (see examples below). While this is unlikely to cause delocalization, it can add variance to the solution in the same order of magnitude of the base MCL localization error.
4. It can add to the overall unreliability of the system in particular places. For example, a long narrow passage with thick walls will have increased orientation variance compared to one with narrow walls, possibly causing the localization/planning loop to trip by not being able to plan a path out of the passage and leaving the robot stuck in it. 

A possible solution comes from the thinking that it's reasonable and still conservative to assume that any cell for which none of its 8 immediate neighbors is a free-space cells cannot be the result of a direct measurement during mapping, and as far as the sensor models are concerned: 

- For the beam model, the cell is unreachable.
- For the likelihood model, the cell is artificially reducing the distance to the closest obstacle metric for that cell, effectively widening the sensor measurement distribution causing the issues mentioned above.

Notice that a cell only surrounded by occupied and unknown cells (e.g. external wall cells in the magazino dataset example below) are also artificially skewing the likelihood model metric while not really providing information, since the lack of an adjacent free-cell indicates that this occupied cell was not measured directly and therefore does not really provide real information to either model.

**Platform (please complete the following information):**
 - OS: development container
[...]

PR #555, opened 2026-04-07
Change default of the strict boundaries setting to true

glpuga:
### Proposed changes

This feature provides a slight improvement for likelihood field sensor models, while having no impact for others, so it's likely to be an improvement for all. The improvement is map-quality-dependent, and basically comes from reducing the likelihood field uncertainty caused by having multiple consecutive cells marked as obstacles.

See for context:

- https://github.com/Ekumen-OS/beluga/issues/265
- https://github.com/Ekumen-OS/beluga/pull/541

To validate the impact of enabling this by default we run both beluga and nav2 against a number of datasets, both real world and simulated, and with and without unmapped obstacles.

The results can be seen here: 

[lambkin_run_beluga_with_strict_boundaries_on_report (draft).pdf](https://github.com/user-attachments/files/26541826/lambkin_run_beluga_with_strict_boundaries_on_report.draft.pdf)

For reference, a previous testbench campaign against mostly the same datasets can be found in:
https://github.com/Ekumen-OS/beluga/pull/476#issuecomment-2741609219

#### Type of change

### Checklist

### Additional comments


beluga_amcl/docs/ros2-reference.md
```diff
@@ -318,3 +318,5 @@ Also available as a standalone `amcl_node` executable.
 | `beam_skip_distance` | Maximum distance to an obstacle to consider that a beam coincides with the map. Beluga AMCL does not support beam skipping. | ✅ |  |  |
 | `beam_skip_threshold` | Minimum percentage of particles for which a particular beam must match the map to not be skipped. Beluga AMCL does not support beam skipping. | ✅ |  |  |
 | `beam_skip_error_threshold` | Maximum percentage of skipped beams. Too many skipped beams trigger a full update to recover in case of bad convergence. Beluga AMCL does not support beam skipping. | ✅ |  |  |
+| `model_unknown_space` | Whether to model unknown space or assume it free. Default is `false`. |  | ✅ |  |
+| `likelihood_from_strict_obstacle_edges` | Whether to regard inner obstacle cells as unknown space when calculating the likelihood field. Default is `true`. |  | ✅ |  |
```
hidmic: @glpuga @LaBruma nit²:
```suggestion
| `only_obstacle_boundaries` | Whether to treat inner obstacle cells as unknown space, particularly when calculating likelihood fields. Default is `true`. |  | ✅ |  |
```

hidmic (approved):

beluga_amcl/docs/ros2-reference.md
LaBruma: We can change (again) the name of the parameter. I think it makes sense to call it `only_obstacle_boundaries`. Beside this, I'm not being unable to merge, is that only due to the failure of codecov?

hidmic: I think all conversations need to be resolved to merge. Feel free to resolve this.

PR #563, opened 2026-05-20
Add localization quality estimation metric

papalmac:
### Proposed changes

Implements a quality/degradation metric based in a comparison between a expected covariance and the real covariance. It is based in the answer in https://github.com/Ekumen-OS/beluga/discussions/562.
A new publisher "localization_quality" provides a float between [0, 1] , the closer to 1 better quality.

#### Type of change

### Checklist

### Additional comments

It is my first PR. Need some help to test this and to add a test to check the feature.

### DIFF
--- a/beluga_ros/src/amcl.cpp
+++ b/beluga_ros/src/amcl.cpp
@@ -21,7 +21,11 @@
 #include <beluga/algorithm/cluster_based_estimation.hpp>
 #include <beluga/views/random_intersperse.hpp>
 #include <beluga/views/take_while_kld.hpp>
+#include <algorithm>
 #include <cmath>
+#include <limits>
+#include <random>
+#include <vector>
 
 namespace beluga_ros {
 
@@ -122,7 +126,47 @@ auto Amcl::update(
   }
 
   force_update_ = false;
-  return beluga::cluster_based_estimate(beluga::views::states(particles_), beluga::views::weights(particles_));
+  auto estimate =
+      beluga::cluster_based_estimate(beluga::views::states(particles_), beluga::views::weights(particles_));
+  last_quality_ = compute_quality(estimate.second);
+  return estimate;
+}
+
+double Amcl::compute_quality(const Sophus::Matrix3d& actual_covariance) {
+  const std::size_t N = params_.min_particles;
+
+  // a known seed is required to provide the exact same reference each time quality is computed.
+  std::mt19937 gen{42};
+  std::normal_distribution<double> dx{0.0, params_.expected_pose_x_stddev};
+  std::normal_distribution<double> dy{0.0, params_.expected_pose_y_stddev};
+  std::normal_distribution<double> dyaw{0.0, params_.expected_pose_yaw_stddev};
+
+  std::vector<Sophus::SE2d> ref_states;
+  ref_states.reserve(N);
+  for (std::size_t i = 0; i < N; ++i) {
+    ref_states.emplace_back(Sophus::SO2d{dyaw(gen)}, Sophus::Vector2d{dx(gen), dy(gen)});
+  }
+
+  std::visit(
+      [&](const auto& motion_model) {
+        auto sampling_fn = motion_model(control_action_window_);
+        for (auto& state : ref_states) {
+          state = sampling_fn(state, gen);
+        }
+      },
+      motion_model_);
+
+  const std::vector<double> uniform_weights(N, 1.0);
+  const auto [ref_mean, ref_covariance] = beluga::estimate(ref_states, uniform_weights);
+
+  double quality = 1.0;
+  for (int i = 0; i < 3; ++i) {
+    const double actual = actual_covariance.coeff(i, i);
+    if (actual > std::numeric_limits<double>::epsilon()) {
+      quality = std::min(quality, ref_covariance.coeff(i, i) / actual);
+    }
+  }
+  return std::clamp(quality, 0.0, 1.0);
 }
 
 }  // namespace beluga_ros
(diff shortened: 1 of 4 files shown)


beluga_ros/src/amcl.cpp
```diff
@@ -122,7 +126,46 @@ auto Amcl::update(
   }
 
   force_update_ = false;
-  return beluga::cluster_based_estimate(beluga::views::states(particles_), beluga::views::weights(particles_));
+  auto estimate = beluga::cluster_based_estimate(beluga::views::states(particles_), beluga::views::weights(particles_));
+  last_quality_ = compute_quality(estimate.second);
+  return estimate;
+}
+
+double Amcl::compute_quality(const Sophus::Matrix3d& actual_covariance) {
```
hidmic: @papalmac meta, maybe followup: I think it may be worth to move this into the core, taking the nominal distribution, the motion model, and the last available estimate.

beluga_ros/src/amcl.cpp
```diff
@@ -122,7 +126,46 @@ auto Amcl::update(
...
+  const std::size_t n = params_.min_particles;
+
+  // a known seed is required to provide the exact same reference each time quality is computed.
+  std::mt19937 gen{42};
+  std::normal_distribution<double> dx{0.0, params_.expected_pose_x_stddev};
+  std::normal_distribution<double> dy{0.0, params_.expected_pose_y_stddev};
+  std::normal_distribution<double> dyaw{0.0, params_.expected_pose_yaw_stddev};
+
+  std::vector<Sophus::SE2d> ref_states;
```
hidmic: @papalmac we have [`MultivariateNormalDistribution`](https://github.com/Ekumen-OS/beluga/blob/main/beluga/include/beluga/random/multivariate_normal_distribution.hpp) specifically for this purpose (ie. sampling an SE2 normal distribution).

beluga_ros/src/amcl.cpp
```diff
@@ -122,7 +126,46 @@ auto Amcl::update(
...
+      motion_model_);
+
+  const std::vector<double> uniform_weights(n, 1.0);
+  const auto [ref_mean, ref_covariance] = beluga::estimate(ref_states, uniform_weights);
+
+  double quality = 1.0;
+  for (int i = 0; i < 3; ++i) {
+    const double actual = actual_covariance.coeff(i, i);
+    if (actual > std::numeric_limits<double>::epsilon()) {
+      quality = std::min(quality, ref_covariance.coeff(i, i) / actual);
```
hidmic: @papalmac meta: this ratio bugs me a bit. A very small actual covariance would be indicative of distribution collapse, not a high quality estimate. I'd expect the covariance to stay in the vicinity of expectations. CC'ing @glpuga as the mastermind behind this heuristic.

hidmic (commented):
@papalmac mind to update the PR title to a more descriptive name?

beluga_ros/src/amcl.cpp
hidmic: This will also help unit testing this. That's step 1. Then we'll need to do some benchmarking and/or testing in the field. @agalbachicar would you be open to collaborate on that?

beluga_ros/src/amcl.cpp
papalmac: Comment addressed in commit 62117bc

beluga_ros/src/amcl.cpp
```diff
@@ -43,6 +47,17 @@ Amcl::Amcl(
...
+  Sophus::Matrix3d covariance = Sophus::Matrix3d::Zero();
+  covariance(0, 0) = params_.expected_pose_x_stddev * params_.expected_pose_x_stddev;
+  covariance(1, 1) = params_.expected_pose_y_stddev * params_.expected_pose_y_stddev;
+  covariance(2, 2) = params_.expected_pose_yaw_stddev * params_.expected_pose_yaw_stddev;
+  auto dist = beluga::MultivariateNormalDistribution<Sophus::SE2d>{covariance};
+  std::mt19937 gen{42};
+  ref_states_.reserve(params_.min_particles);
+  for (std::size_t i = 0; i < params_.min_particles; ++i) {
+    ref_states_.emplace_back(dist(gen));
+  }
```
glpuga: I'm not sure I would use min_particles here, especially since we are reusing systematically the same set, which means we will amplify any initial bias.

beluga_ros/src/amcl.cpp
```diff
@@ -122,7 +137,36 @@ auto Amcl::update(
...
+      motion_model_);
+
+  const std::vector<double> uniform_weights(ref_states.size(), 1.0);
+  const auto [ref_mean, ref_covariance] = beluga::estimate(ref_states, uniform_weights);
+
+  double quality = 1.0;
+  for (int i = 0; i < 3; ++i) {
+    const double actual = actual_covariance.coeff(i, i);
+    if (actual > std::numeric_limits<double>::epsilon()) {
+      quality = std::min(quality, ref_covariance.coeff(i, i) / actual);
```
glpuga: It's a bit distorsive than the quotient is made in terms of the squared values, since that will amplify any difference in the standard deviations.

glpuga: having a single index for all three dimensions clouds poor choices for a particular dimension. I think all three should be outputs.

glpuga: Also, calculating the min over the three with no averaging is very biasing, since you only need one of the values to take a tail end value of the distribution to make a huge difference in the results.

papalmac: Comment addressed in 9c8d158

PR #558, opened 2026-04-10
Add PoseSensorModel for absolute pose measurements

pepisg:
### Proposed changes

Adds a map-free, header-only sensor model that weights particles using a Mahalanobis-distance Gaussian likelihood computed in the Lie algebra of SE(2) and SE(3). Supports 2D and 3D state types via a single template.

This is a draft! @hidmic let me know if anything should be changed, this is my first PR to the repo

#### Type of change

- [ ] 🐛 Bugfix (change which fixes an issue)
- [x] 🚀 Feature (change which adds functionality)
- [ ] 📚 Documentation (change which fixes or extends documentation)

### Checklist

_Put an `x` in the boxes that apply. This is simply a reminder of what we will require before merging your code._

- [ ] Lint and unit tests (if any) pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] All commits have been signed for [DCO](https://developercertificate.org/)

### Additional comments

Tests were written by claude code. not sure if this is the right way of doing it.

PR #574, opened 2026-07-21
Add PoseSensorModel for absolute pose measurements (prev PR 558)

glpuga:
### Proposed changes

This PR takes from https://github.com/Ekumen-OS/beluga/pull/558 

From there:

> Adds a map-free, header-only sensor model that weights particles using a Mahalanobis-distance Gaussian likelihood computed in the Lie algebra of SE(2) and SE(3). Supports 2D and 3D state types via a single template.

#### Type of change

### Checklist

### Additional comments

### DIFF
--- a/beluga/include/beluga/sensor/pose_sensor_model.hpp
+++ b/beluga/include/beluga/sensor/pose_sensor_model.hpp
@@ -1,0 +1,141 @@
+// Copyright 2026 Ekumen, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef BELUGA_SENSOR_POSE_SENSOR_MODEL_HPP
+#define BELUGA_SENSOR_POSE_SENSOR_MODEL_HPP
+
+// external
+#include <Eigen/Cholesky>
+#include <Eigen/Core>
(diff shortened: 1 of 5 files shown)


beluga/include/beluga/sensor/pose_sensor_model.hpp
```diff
@@ -0,0 +1,128 @@
...
+  explicit PoseSensorModel(param_type params) : params_{std::move(params)} { assert(params_.min_weight > 0.0); }
+
+  /// Returns a state weighting function conditioned on an absolute pose measurement.
+  /**
+   * \param measurement Absolute pose measurement including covariance.
+   * \return a state weighting function satisfying \ref StateWeightingFunctionPage
+   */
+  [[nodiscard]] auto operator()(measurement_type&& measurement) const {
+    const auto llt = measurement.covariance.llt();
+    assert(llt.info() == Eigen::ComputationInfo::Success);
```
hidmic: @glpuga meta: should we add an assertion that doesn't go away if compiled for release?

beluga/include/beluga/sensor/pose_sensor_model.hpp
```diff
@@ -0,0 +1,128 @@
...
+   */
+  explicit PoseSensorModel(param_type params) : params_{std::move(params)} { assert(params_.min_weight > 0.0); }
+
+  /// Returns a state weighting function conditioned on an absolute pose measurement.
+  /**
+   * \param measurement Absolute pose measurement including covariance.
+   * \return a state weighting function satisfying \ref StateWeightingFunctionPage
+   */
+  [[nodiscard]] auto operator()(measurement_type&& measurement) const {
+    const auto llt = measurement.covariance.llt();
```
hidmic, line 111: Smart. I wonder if we are as consistent about using Cholesky decomposition for covariance matrix inversion.

hidmic (approved):
Great patch! Unfortunate the original author didn't push it through.

beluga/include/beluga/sensor/pose_sensor_model.hpp
glpuga: Replaced for hard check with graceful failure.

PR #577, opened 2026-08-03
Added: beam-skipping to likelihood-field-prob

glpuga:
### Proposed changes

Draft of the beam skipping feature, companion to the likelihood_field_prob sensor model .

#### Type of change

### Checklist

### Additional comments

### DIFF
--- a/beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
+++ b/beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
@@ -18,8 +18,11 @@
 #include <algorithm>
 #include <beluga/sensor/likelihood_field_model_base.hpp>
 #include <cmath>
+#include <cstddef>
 #include <random>
 #include <vector>
+
+#include <sophus/common.hpp>
 
 /**
  * \file
@@ -28,11 +31,37 @@
 
 namespace beluga {
 
-/// Parameters used to construct a LikelihoodFieldProbModelParam instance.
+/// Parameters used to construct a LikelihoodFieldProbModel instance.
 /**
  * See Probabilistic Robotics \cite thrun2005probabilistic Chapter 6.4, particularly Table 6.3.
+ *
+ * The first five fields mirror beluga::LikelihoodFieldModelBaseParam (preserving its
+ * field order for aggregate initialization). The remaining fields configure the optional
+ * beam skipping heuristic, which detects beams that disagree with the map across a large
+ * fraction of the particle set (e.g. caused by unmapped/dynamic obstacles) and excludes
+ * them from the weight computation. Beam skipping is disabled by default; see
+ * https://github.com/Ekumen-OS/beluga/issues/187 for context.
  */
-using LikelihoodFieldProbModelParam = LikelihoodFieldModelBaseParam;
+struct LikelihoodFieldProbModelParam {
+  /// Maximum distance to obstacle.
+  double max_obstacle_distance = 100.0;
+  /// Maximum range of a laser ray.
+  double max_laser_distance = 2.0;
+  /// Weight used to combine the probability of hitting an obstacle.
+  double z_hit = 0.5;
+  /// Weight used to combine the probability of random noise in perception.
+  double z_random = 0.5;
+  /// Standard deviation of a gaussian centered arounds obstacles.
+  double sigma_hit = 0.2;
+  /// Whether to enable the beam skipping heuristic.
+  bool do_beamskip = false;
+  /// Distance to the nearest obstacle below which a beam is considered to agree with the map.
+  double beam_skip_distance = 0.5;
+  /// Fraction of particles that must agree on a beam for it to be kept.
+  double beam_skip_threshold = 0.3;
+  /// If the fraction of skipped beams exceeds this value, skipping is disabled for the update.
+  double beam_skip_error_threshold = 0.9;
+};
 
 /// Likelihood field prob sensor model for range finders.
 /**
@@ -54,10 +83,97 @@ class LikelihoodFieldProbModel : public LikelihoodFieldModelBase<OccupancyGrid> 
 
   /// Constructs a LikelihoodFieldProbModel instance.
   /**
-   * @copydoc LikelihoodFieldModelBase::LikelihoodFieldModelBase
+   * \param params Parameters to configure this instance.
+   *  See beluga::LikelihoodFieldProbModelParam for details.
+   * \param grid Occupancy grid representing the static map that the sensor model
+   *  uses to compute a likelihood field for lidar hits and compute importance weights
+   *  for particle states.
    */
   explicit LikelihoodFieldProbModel(const param_type& params, const map_type& grid)
-      : LikelihoodFieldModelBase<OccupancyGrid>(params, grid) {}
+      : LikelihoodFieldModelBase<OccupancyGrid>(to_base_param(params), grid),
+        do_beamskip_{params.do_beamskip},
+        beam_skip_threshold_{params.beam_skip_threshold},
+        beam_skip_error_threshold_{params.beam_skip_error_threshold},
+        likelihood_threshold_{compute_likelihood_threshold(params)} {}
+
+  /// Precomputes the beam skipping mask from the current particle states.
+  /**
+   * Runs the first pass of the beam skipping heuristic: for every beam, it counts the
+   * fraction of particle states for which the beam endpoint lands close enough to a mapped
+   * obstacle (within `beam_skip_distance`, expressed here as a likelihood threshold). Beams
+   * for which that fraction falls below `beam_skip_threshold` are masked out and ignored by
+   * the subsequent weight computation in `operator()`. If skipping would discard more than
+   * `beam_skip_error_threshold` of the beams, the mask is reset so that all beams are used,
+   * preventing filter divergence. This is a no-op when beam skipping is disabled.
+   *
+   * Must be called once per update, after motion propagation and before reweighting.
+   *
+   * \tparam StateRange A range of particle states (Sophus::SE2d).
+   * \param points 2D lidar hit points in the reference frame of particle states.
+   * \param states Range with the (propagated) particle states for the current update.
+   */
+  template <class StateRange>
+  void prepare(const measurement_type& points, StateRange&& states) {
+    if (!do_beamskip_) {
+      return;
+    }
+
+    const std::size_t num_beams = points.size();
+    beam_mask_.assign(num_beams, true);
+    if (num_beams == 0) {
+      return;
+    }
+
+    const auto unknown_space_occupancy_prob = static_cast<float>(1. / this->params_.max_laser_distance);
+    std::vector<std::size_t> obs_count(num_beams, 0);
+    std::size_t num_states = 0;
+
+    for (const auto& state : states) {
+      ++num_states;
+      const auto transform = this->world_to_likelihood_field_transform_ * state;
+      const auto x_offset = transform.translation().x();
+      const auto y_offset = transform.translation().y();
+      const auto cos_theta = transform.so2().unit_complex().x();
+      const auto sin_theta = transform.so2().unit_complex().y();
+      for (std::size_t i = 0; i < num_beams; ++i) {
+        const auto& point = points[i];
+        const auto x = point.first * cos_theta - point.second * sin_theta + x_offset;
+        const auto y = point.first * sin_theta + point.second * cos_theta + y_offset;
+        const auto pz = this->likelihood_field_.data_near(x, y).value_or(unknown_space_occupancy_prob);
+        // The likelihood field is monotonically decreasing in the distance to the nearest
+        // obstacle, so "distance < beam_skip_distance" is equivalent to "pz > threshold".
+        if (pz > likelihood_threshold_) {
+          ++obs_count[i];
+        }
+      }
+    }
+
+    if (num_states == 0) {
+      return;
+    }
+
+    std::size_t skipped = 0;
+    for (std::size_t i = 0; i < num_beams; ++i) {
+      const double ratio = static_cast<double>(obs_count[i]) / static_cast<double>(num_states);
+      beam_mask_[i] = ratio > beam_skip_threshold_;
+      if (!beam_mask_[i]) {
+        ++skipped;
+      }
+    }
+
+    // Safety fallback: if too many beams would be skipped, integrate all of them instead.
+    const double skipped_ratio = static_cast<double>(skipped) / static_cast<double>(num_beams);
+    if (skipped_ratio > beam_skip_error_threshold_) {
+      std::fill(beam_mask_.begin(), beam_mask_.end(), true);
+    }
+  }
+
+  /// Returns the current beam skipping mask (one boolean per beam, true means the beam is used).
+  /**
+   * The mask is populated by `prepare()`. It is empty until the first call, which is
+   * equivalent to using every beam. Mainly useful for introspection and testing.
+   */
+  [[nodiscard]] const std::vector<bool>& beam_mask() const { return beam_mask_; }
 
   /// Returns a state weighting function conditioned on 2D lidar hits.
   /**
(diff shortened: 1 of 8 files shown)


beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
```diff
@@ -18,21 +18,50 @@
...
+  double sigma_hit = 0.2;
+  /// Whether to enable the beam skipping heuristic.
+  bool do_beamskip = false;
+  /// Distance to the nearest obstacle below which a beam is considered to agree with the map.
+  double beam_skip_distance = 0.5;
+  /// Fraction of particles that must agree on a beam for it to be kept.
+  double beam_skip_threshold = 0.3;
+  /// If the fraction of skipped beams exceeds this value, skipping is disabled for the update.
+  double beam_skip_error_threshold = 0.9;
+};
```
glpuga: I'm guessing we are duplicating fields in the original struct. You can instead inherit from LikelihoodFieldModelBaseParam to extend it.

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
```diff
@@ -54,10 +83,97 @@ class LikelihoodFieldProbModel : public LikelihoodFieldModelBase<OccupancyGrid>
...
+      for (std::size_t i = 0; i < num_beams; ++i) {
+        const auto& point = points[i];
+        const auto x = point.first * cos_theta - point.second * sin_theta + x_offset;
+        const auto y = point.first * sin_theta + point.second * cos_theta + y_offset;
+        const auto pz = this->likelihood_field_.data_near(x, y).value_or(unknown_space_occupancy_prob);
+        // The likelihood field is monotonically decreasing in the distance to the nearest
+        // obstacle, so "distance < beam_skip_distance" is equivalent to "pz > threshold".
+        if (pz > likelihood_threshold_) {
+          ++obs_count[i];
+        }
```
glpuga, line 140: It's a bit unfortunate that we need to use the probability as a proxy for distance, because it will not be correct. That is mainly because there exists a `laser_likelihood_max_dist` that sets a maximum distance to calculate the likelihood field map, and for distances larger than that we use a flat value.

That means if the if the beam_skip_distance > laser_likelihood_max_dist , we will count all rays that hit beyond  laser_likelihood_max_dist as being "in".

We don't have a distance map, which would solve this, but I think that's good because having one would bring trouble:
- Double the memory usage for large maps.
- laser_likelihood_max_dist exists because for most real-world maps it reduces the time it takes to calculate the distance map to very small fraction because most of the area of the map is void areas you don't care the likelihood in any case.

Both issues probably have other solutions that the ones currently in use, but that would mean a deviation from the current behavior of nav2.

We can discuss it, though.

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
```diff
@@ -18,21 +18,50 @@
...
+  /// Standard deviation of a gaussian centered arounds obstacles.
+  double sigma_hit = 0.2;
+  /// Whether to enable the beam skipping heuristic.
+  bool do_beamskip = false;
+  /// Distance to the nearest obstacle below which a beam is considered to agree with the map.
+  double beam_skip_distance = 0.5;
+  /// Fraction of particles that must agree on a beam for it to be kept.
+  double beam_skip_threshold = 0.3;
+  /// If the fraction of skipped beams exceeds this value, skipping is disabled for the update.
+  double beam_skip_error_threshold = 0.9;
```
glpuga, line 53: Are these the default values used in nav2? beam_skip_threshold_ at 0.9 sounds very conservative

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
```diff
@@ -54,10 +83,97 @@ class LikelihoodFieldProbModel : public LikelihoodFieldModelBase<OccupancyGrid>
...
+      ++num_states;
+      const auto transform = this->world_to_likelihood_field_transform_ * state;
+      const auto x_offset = transform.translation().x();
+      const auto y_offset = transform.translation().y();
+      const auto cos_theta = transform.so2().unit_complex().x();
+      const auto sin_theta = transform.so2().unit_complex().y();
+      for (std::size_t i = 0; i < num_beams; ++i) {
+        const auto& point = points[i];
+        const auto x = point.first * cos_theta - point.second * sin_theta + x_offset;
+        const auto y = point.first * sin_theta + point.second * cos_theta + y_offset;
```
glpuga, line 134: Looks like we can do this transformation using Sophus as a product of poses.

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
```diff
@@ -74,20 +190,59 @@ class LikelihoodFieldProbModel : public LikelihoodFieldModelBase<OccupancyGrid>
...
+        // Not using Eigen/Sophus because they make the routine x10 slower.
+        // See `benchmark_likelihood_field_model.cpp` for reference.
+        const auto& point = points[i];
+        const auto x = point.first * cos_theta - point.second * sin_theta + x_offset;
+        const auto y = point.first * sin_theta + point.second * cos_theta + y_offset;
+        const auto pz =
+            static_cast<double>(this->likelihood_field_.data_near(x, y).value_or(unknown_space_occupancy_prob));
+        log_weight += std::log(pz);
+      }
+      return std::exp(log_weight);
```
glpuga, line 208: You can do this without resorting back to for loops and at the same time reduce the amount of changes by zipping together the beam_mask and the points vectors, and returning 0.0 from the lambda if the beam is masked. You don't need to check do_beamskip_ because the mask will be true for all beams in that case anyways.,

glpuga: For an example see https://github.com/Ekumen-OS/beluga/blob/main/beluga/include/beluga/sensor/likelihood_field_model_base.hpp#L165-L175

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
```diff
@@ -74,20 +190,59 @@ class LikelihoodFieldProbModel : public LikelihoodFieldModelBase<OccupancyGrid>
...
     };
   }
+
+ private:
+  bool do_beamskip_;                  ///< Whether the beam skipping heuristic is enabled.
+  double beam_skip_threshold_;        ///< Fraction of particles that must agree for a beam to be kept.
+  double beam_skip_error_threshold_;  ///< Skipped-beam fraction above which skipping is disabled.
+  float likelihood_threshold_;        ///< Likelihood equivalent of `beam_skip_distance`.
+  std::vector<bool> beam_mask_;       ///< Per-beam mask computed by `prepare()` (true means used).
```
glpuga: JIC a word of advice, beware of bool vectors. https://www.geeksforgeeks.org/cpp/problem-with-std-vector-bool-in-cpp/

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
```diff
@@ -74,20 +190,59 @@ class LikelihoodFieldProbModel : public LikelihoodFieldModelBase<OccupancyGrid>
...
+
+  /// Builds the base model parameters from the full prob model parameters.
+  static LikelihoodFieldModelBaseParam to_base_param(const param_type& params) {
+    LikelihoodFieldModelBaseParam base;
+    base.max_obstacle_distance = params.max_obstacle_distance;
+    base.max_laser_distance = params.max_laser_distance;
+    base.z_hit = params.z_hit;
+    base.z_random = params.z_random;
+    base.sigma_hit = params.sigma_hit;
+    return base;
```
glpuga: I'm not in love with the manual nature of having to add a line here if we add a new field in LikelihoodFieldModelBaseParam. This is a bug waiting to happen.

If possible maybe we can make the LikelihoodfieldProb constructor receive _two_ different param structs, the conventional one and the beam skipping one. I'm not sure if that won't break the templates that instantiate the filter, though.

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
```diff
@@ -74,20 +190,59 @@ class LikelihoodFieldProbModel : public LikelihoodFieldModelBase<OccupancyGrid>
...
+   * "distance to obstacle < beam_skip_distance" agreement test can be evaluated directly on
+   * the precomputed likelihood field without keeping the distance map around.
+   */
+  static float compute_likelihood_threshold(const param_type& params) {
+    const double two_squared_sigma = 2.0 * params.sigma_hit * params.sigma_hit;
+    const double amplitude = params.z_hit / (params.sigma_hit * std::sqrt(2.0 * Sophus::Constants<double>::pi()));
+    const double offset = params.z_random / params.max_laser_distance;
+    const double squared_distance = params.beam_skip_distance * params.beam_skip_distance;
+    return static_cast<float>(amplitude * std::exp(-squared_distance / two_squared_sigma) + offset);
+  }
```
glpuga, line 229: this repeats the calculations in https://github.com/Ekumen-OS/beluga/blob/main/beluga/include/beluga/sensor/likelihood_field_model_base.hpp#L135-L146 .

Better extract the `to_likelihood` lambda in `to_likelihood` into a function in the base class, and reuse from there to calculate the treshold.

beluga/include/beluga/sensor/primitives.hpp
```diff
@@ -37,6 +41,26 @@ struct has_likelihood_field<T, std::void_t<decltype(std::declval<T>().likelihood
...
+// Specialization. Uses SFINAE to detect whether a sensor model exposes a
+// `prepare(measurement, states)` method, used to precompute a beam skipping mask over the
+// whole particle set before per-particle reweighting. The states range is probed with a
+// `std::vector<state_type>`; any range type is accepted by the actual templated method.
+template <class T>
+struct has_beam_skip<
+    T,
+    std::void_t<decltype(std::declval<T&>().prepare(
+        std::declval<const typename T::measurement_type&>(),
+        std::declval<const std::vector<typename T::state_type>&>()))>> : std::true_type {};
```
glpuga, line 58: Nice.

beluga_amcl/src/amcl_node.cpp
```diff
@@ -180,6 +180,48 @@ AmclNode::AmclNode(const rclcpp::NodeOptions& options) : BaseAMCLNode{"amcl", ""
...
+    descriptor.floating_point_range[0].from_value = 0;
+    descriptor.floating_point_range[0].to_value = 1;
+    descriptor.floating_point_range[0].step = 0;
+    declare_parameter("beam_skip_threshold", rclcpp::ParameterValue(0.3), descriptor);
+  }
+
+  {
+    auto descriptor = rcl_interfaces::msg::ParameterDescriptor();
+    descriptor.description =
+        "Fraction of skipped beams above which beam skipping is disabled for that update, preventing divergence.";
```
glpuga: I'm not sure I can tell from this description what the threshold does. Maybe write as "If more than this fraction of beam disagree with the map, assume localization error and disable beam skipping" or something like that.

beluga_amcl/src/amcl_node.cpp
```diff
@@ -180,6 +180,48 @@ AmclNode::AmclNode(const rclcpp::NodeOptions& options) : BaseAMCLNode{"amcl", ""
...
+    descriptor.floating_point_range.resize(1);
+    descriptor.floating_point_range[0].from_value = 0;
+    descriptor.floating_point_range[0].to_value = std::numeric_limits<double>::max();
+    descriptor.floating_point_range[0].step = 0;
+    declare_parameter("beam_skip_distance", rclcpp::ParameterValue(0.5), descriptor);
+  }
+
+  {
+    auto descriptor = rcl_interfaces::msg::ParameterDescriptor();
+    descriptor.description = "Fraction of particles that must agree on a beam for it to be kept, used for beam "
```
glpuga: Maybe "Agreement threshold above which beam-skipping considers a hit a static map hit" or something like that.

beluga_amcl/src/amcl_node.cpp
```diff
@@ -180,6 +180,48 @@ AmclNode::AmclNode(const rclcpp::NodeOptions& options) : BaseAMCLNode{"amcl", ""
...
+    descriptor.description =
+        "Whether to enable beam skipping, ignoring beams that disagree with the map across most "
+        "particles (e.g. caused by unmapped or dynamic obstacles). Only used by the "
+        "likelihood_field_prob model.";
+    declare_parameter("do_beamskip", false, descriptor);
+  }
+
+  {
+    auto descriptor = rcl_interfaces::msg::ParameterDescriptor();
+    descriptor.description = "Distance to the nearest obstacle below which a beam agrees with the map, used for beam "
```
glpuga:
```suggestion
    descriptor.description = "Distance threshold for beam-skipping to consider hit a likely true static map hit"
```

beluga_example/params/default.ros2.yaml
```diff
@@ -50,7 +50,7 @@ amcl:
     # Maximum angle difference from last resample for resampling to happen again.
     update_min_d: 0.25
     # Laser sensor model type.
-    laser_model_type: likelihood_field
+    laser_model_type: likelihood_field_prob
```
glpuga: This is not the default value, you probably don't want to merge this.

beluga_example/params/default.ros2.yaml
```diff
@@ -69,6 +69,16 @@ amcl:
...
+    # Whether to enable beam skipping (likelihood_field_prob only). When enabled, beams that
+    # disagree with the map across most particles (e.g. unmapped or dynamic obstacles) are
+    # ignored. Disabled by default; only beneficial in scenes with unmapped/dynamic obstacles.
+    do_beamskip: true
+    # Distance to the nearest obstacle below which a beam is considered to agree with the map.
+    beam_skip_distance: 0.5
+    # Fraction of particles that must agree on a beam for it to be kept.
+    beam_skip_threshold: 0.3
+    # Fraction of skipped beams above which beam skipping is disabled for that update.
+    beam_skip_error_threshold: 0.9
```
glpuga, line 81: Reconsider the comments with my comments on the parameters descriptors above.

glpuga (commented):
It's a great first pass @ralcoberro

glpuga:
@griswaldbrooks , this is the WIP beam-skipping feature that @ralcoberro has been working on that we mentioned last week.

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
glpuga: Nevermind. I think I got the threshold value backwards.

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
ralcoberro: I addressed this comment extending the base
`struct LikelihoodFieldProbModelParam : public LikelihoodFieldModelBaseParam {
  /// Whether to enable the beam skipping heuristic.
  bool do_beamskip = false;
  /// Distance to the nearest obstacle below which a beam is considered to agree with the map.
  double beam_skip_distance = 0.5;
  /// Fraction of particles that must agree on a beam for it to be kept.
  double beam_skip_threshold = 0.3;
  /// If the fraction of skipped beams exceeds this value, skipping is disabled for the update.
  double beam_skip_error_threshold = 0.9;
};`

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
ralcoberro: Thank you Gera, let's talk about this

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
ralcoberro: They are the same as NAV2 AMCL. I resolve this comment

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
ralcoberro: You are right. Changed it to std::vector<std::uint8_t>

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
ralcoberro: Yes, this code was eliminated because now I'm extending the LikelihoodFieldModelBaseParam struct

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
glpuga-eka: Using std::vector<bool> is ok as long as we don't do dangerous ops, such as capturing lambdas, auto assignments, etc.

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
ralcoberro: The manual transform is intentional: using Eigen/Sophus here makes the routine ~10x slower (this is the reweight hot path, run per beam per particle). It's the same pattern and rationale as operator() and the base LikelihoodFieldModel, backed by benchmark_likelihood_field_model.cpp. I added the explanatory comment to prepare() so it's documented there too.

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
ralcoberro: Good Idea. Done

beluga_amcl/src/amcl_node.cpp
ralcoberro: Done

beluga_example/params/default.ros2.yaml
ralcoberro: Right, restored to the previous value

beluga_example/params/default.ros2.yaml
ralcoberro: Done

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
glpuga-eka: let me know when you want to talk about this.

glpuga: tagging @LaBruma for awareness.

glpuga: @ralcoberro @LaBruma 

To unblock this, given that th beam skip distance will usually be much smaller than `laser_likelihood_distance`, and so are their default values, lets declare that `laser_likelihood_distance > beam_skip_distance` is a requirement of the configuration. Also add an assertion in the beamskipping code itself enforcing that constraint.

In https://github.com/Ekumen-OS/beluga/blob/main/beluga_amcl/src/amcl_node.cpp#L376-L395 add a check that if beam skipping is enabled, `laser_likelihood_max_dist` must be greater or equal to `beam_skip_distance`. If it is not, we throw to let the user know. 

Also, in https://github.com/Ekumen-OS/beluga/blob/feature/beam-skipping-likelihood-field-prob/beluga_amcl/src/amcl_node.cpp#L199C24-L199C42 and in the documentation entry for `beam_skip_distance` make that constraint explicit.

There are alternatives, but they will require a lot of refactoring to the `make_likelihood_field` function, and in practice will have the same computational cost as making `laser_likelihood_distance == beam_skip_distance`, because that's the only way not to have to create more than one copy of the distance map.

ralcoberro: Ok, @glpuga it works for me

ralcoberro: Done, implemented as agreed:
beluga_amcl/src/amcl_node.cpp, in get_sensor_model(): when do_beamskip is enabled and beam_skip_distance > laser_likelihood_max_dist we now throw std::invalid_argument. It is caught by the existing handler in map_callback, so the user gets Could not initialize particle filter (....). Equality is accepted.
LikelihoodFieldProbModel::compute_likelihood_threshold() now asserts the same constraint, so the requirement is enforced in the beam skipping code itself and not only at the ROS layer.
The constraint is documented in three places: the beam_skip_distance parameter descriptor, the LikelihoodFieldProbModelParam::beam_skip_distance doxygen (with the reason: the likelihood field saturates at max_obstacle_distance, so beyond it the field is flat and the agreement test stops discriminating), and beluga_amcl/docs/ros2-reference.md. While there I also added the four beam-skip parameters to the prose section and updated the Nav2 compatibility table, whose rows still claimed Beluga AMCL does not support beam skipping.
New tests in test_amcl_node.cpp: rejection when beam_skip_distance > laser_likelihood_max_dist, acceptance on equality, and no constraint when do_beamskip is false.

beluga/include/beluga/sensor/likelihood_field_prob_model.hpp
ralcoberro: Done. operator() is back to a declarative accumulation: it zips points with the beam mask and the projection returns 0.0 (the neutral element of the log-likelihood sum) for masked beams, so there is no do_beamskip_ check and no index bookkeeping in the hot path.
To make that work I had prepare() always size the mask to the measurement, leaving every beam enabled when skipping is disabled, which is the invariant your comment relies on. operator() also resolves the mask once per update before building the lambda: if prepare() was not called for this measurement (the model used standalone, as in the unit tests) it falls back to an all-enabled mask, so the plain likelihood-field-prob behavior is preserved. That resolution is once per update, not per particle. Covered by a new WithoutPrepareIntegratesEveryBeam test.

PR #579, opened 2026-08-18
Feature/use kd tree for landmark map search

PaulVerhoeckx:
### Proposed changes

Implement efficient nearest-neighbor landmark search using per-category KD-trees with nanoflann. Provides O(log n) spatial queries instead of O(n) linear search for position-based landmark matching.

Tested for a particle filter with:
- 1000 particles
- 3 landmark detections
- landmark map containing 50.000 landmarks

This reduced the filter update time from 1.4 seconds to 0.005 seconds. (12th Gen Intel® Core™ i7-1255U × 12)  

**Changes**
- Added category-based KD-tree indices for fast spatial nearest-neighbor search
- Implemented `find_nearest_landmark()` using KD-tree queries for efficient position-based matching
- Cached per-category indices with lazy construction during map initialization
- Added PositionCloud adapter and CategoryIndex structure for nanoflann integration
- Deprecated copy operations to prevent expensive kd-tree reconstruction

#### Type of change

### Checklist

### Additional comments

Note on `find_closest_bearing_landmark()` implementation:
This method currently uses O(n) linear search instead of the KD-tree indices because bearing-based matching requires angular distance metrics, which are fundamentally incompatible with the Euclidean L2 distance metric used by the position-based KD-trees.

### DIFF
--- a/beluga/include/beluga/sensor/data/landmark_map.hpp
+++ b/beluga/include/beluga/sensor/data/landmark_map.hpp
@@ -51,8 +51,7 @@ class LandmarkMap {
   /// @param boundaries Limits of the map.
   /// @param landmarks List of landmarks that can be expected to be detected.
   explicit LandmarkMap(const LandmarkMapBoundaries& boundaries, landmarks_set_position_data landmarks)
-      : landmarks_(std::move(landmarks)), map_boundaries_(std::move(boundaries))
-  {
+      : landmarks_(std::move(landmarks)), map_boundaries_(std::move(boundaries)) {
     build_category_indices();
   }
 
@@ -81,8 +80,7 @@ class LandmarkMap {
   /// @details Copying requires expensive reconstruction of the cached kd-tree indices.
   /// @deprecated Use move semantics (std::move) instead to avoid performance overhead.
   /// @param other Landmark map to copy from.
-  [[deprecated("LandmarkMap copying is expensive. Use std::move() instead.")]]
-  LandmarkMap(const LandmarkMap& other)
+  [[deprecated("LandmarkMap copying is expensive. Use std::move() instead.")]] LandmarkMap(const LandmarkMap& other)
       : landmarks_(other.landmarks_), map_boundaries_(other.map_boundaries_) {
     build_category_indices();
   }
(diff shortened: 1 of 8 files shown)


glpuga:
Thanks for the PR, @PaulVerhoeckx . 

Notice that you can run the test suite very similarly to how CI builds and runs it by executing `./src/beluga/tools/build-and-test.sh` in the [development container](https://github.com/Ekumen-OS/beluga/blob/main/DEVELOPING.md#environment). That way you don't need to wait for us to approve the CI run.

PaulVerhoeckx:
> Thanks for the PR, @PaulVerhoeckx .
> 
> Notice that you can run the test suite very similarly to how CI builds and runs it by executing `./src/beluga/tools/build-and-test.sh` in the [development container](https://github.com/Ekumen-OS/beluga/blob/main/DEVELOPING.md#environment). That way you don't need to wait for us to approve the CI run.

Thanks for the tip, I think it will pass now;)

beluga/include/beluga/sensor/data/landmark_map.hpp
```diff
@@ -161,8 +195,52 @@ class LandmarkMap {
...
+  std::unordered_map<LandmarkCategory, std::unique_ptr<CategoryIndex>> category_indices_;
+
+  /// @brief Builds per-category kd-tree indices for nearest-neighbor search.
+  void build_category_indices() {
+    for (const auto& l : landmarks_) {
+      auto& entry = category_indices_[l.category];
+      if (!entry) {
+        entry = std::make_unique<CategoryIndex>();
+      }
+      entry->cloud.pts.push_back(l.detection_position_in_robot);
```
glpuga, line 236: It may be worth it to do a first pass counting how many landmarks for each category are there, to reserve the size in the `pts` vector and that way avoid reallocations as it grows.

beluga/include/beluga/sensor/data/landmark_map.hpp
```diff
@@ -48,7 +56,9 @@ class LandmarkMap {
   /// @param boundaries Limits of the map.
   /// @param landmarks List of landmarks that can be expected to be detected.
   explicit LandmarkMap(const LandmarkMapBoundaries& boundaries, landmarks_set_position_data landmarks)
-      : landmarks_(std::move(landmarks)), map_boundaries_(std::move(boundaries)) {}
+      : landmarks_(std::move(landmarks)), map_boundaries_(std::move(boundaries)) {
+    build_category_indices();
```
glpuga, line 60: Feels wasteful to store the landmarks along with the nanoflann index, since both contain the same information in different forms, and we only use the index later. We only use the landmarks struct to recreate the index in case of copy/move operations.

We can probably remove `landmarks_` if we refactor the  `category_indices_`  storage. Maybe we would have to split `build_category_indices` in two, one step to fill the `pts` vectors, and another separate one to create the indices.

The first one is only done in the constructor that takes in the landmarks data. The map of categories and the points for each become trivially copyable/movable in all constructor/assignment operators, and we need only build the nanoflann index from that data.

Does it make sense?

glpuga (commented):
Nice job! we created this model for a proof of concept demo a long time ago and we did not put a lot performance into it, it's good to see it getting some dedication.

glpuga:
@PaulVerhoeckx ignore the rolling failure. If you rebase to the current main we no longer run CI against rolling as it is a moving target.

